Issue #489, opened 2014-09-29
Should libsass have an image-url function?


michaek:
See https://github.com/sass/libsass/blob/fca1f75a14fe5336c7b1a4b38c69bda7fece714f/functions.cpp#L1179

Because image-url is a Compass function, not a Sass function, it seems odd for it to be included in libsass. Building in any URL functions requires "Sass" to know about your paths - Compass has embraced that pattern, but I don't know if we should assume that Sass does.

xzyfer:
Unless there are any objections I'm going to make `image-url` produce a deprecation warning in 3.1.1. Then let's figure out what we are going to do about it for 3.2.

mgreter:
I'm all in for this! Maybe @hcatlin can pull some strings to get this finally done!? These features should be possible to implement via custom functions now IMO!? I can offer my help to give advice on how to do it via custom functions! But we'd need someone who actually uses this feature to work with us!

hcatlin:
Hey @chriseppstein, has the compass team gotten to a pure-Sass image-url() implementation yet?

chriseppstein:
No sass-based definition of image-url in compass since compass requires ruby, there's no point. Any given project can probably implement a trivial image-url() helper for that project.

hcatlin:
From @jonsuh and @roytomeij:

````
$image-url-path: “../images/” !default;
@function image-url($url) {
  @return url("#{$image-url-path}#{$url}");
}
````

hcatlin:
@chriseppstein Sorry, I thought there was an effort to port some functions from Ruby to Sass, and didn't know if this was one of them.
@xzyfer I think I've agreed since the top to remove it...

chriseppstein:
@hcatlin There is. This will be one of them :)

I wouldn't say this though:

> `image_path` has been removed from libsass till it becomes part of the Sass spec.

Because it's not something that Sass itself would ever add. Maybe this:

> `image_path` has been removed from libsass because it's not part of the Sass spec.

xzyfer:
Ok cool. I'm going to remove the guts of this and add a deprecation message. 3.2. Is pretty close atm so this is a good a time as any.

adyballa:
@hcatlin
btw
I have good use of image_path, because I dont know the prefix in sass. So this function will not work

callumacrae:
I'm confused. Was this deprecated or removed?

Updating to node-sass 3.1 broke all the images on our website. This is from the output CSS:

```css
.sun-rays {
[...]

PR #522, opened 2014-10-09
Remove "image_url" function and "image_path" option

nschonni:
# BREAKING CHANGE

This function doesn't currently exist in the Sass spec.
Fixes gh-489
____

The testing will fail unless SassC is also updated, or the `image_path` option is left in and noop'd


am11:
This was addressed by https://github.com/sass/libsass/pull/834.

Issue #450, opened 2014-07-24
Support import-once Functinality


Snugug:
Now I know this is adding functionality not in Sass yet, but Compass's [import-once](https://github.com/Compass/compass/tree/master/import-once) extension is fantastic but a Ruby based extension to Sass so can't be easily translated to libsass. Having a compile-time option to support this functionality (at least until Sass has it natively and Libsass supports it) would go a long way to being able to transition to Libsass from Ruby Sass

AndyOGo:
it absolutely matters for module based development to avoid duplicated code generation

xzyfer:
@AndyOGo import once is not part of the Sass language. There are community implementation in [sass](https://github.com/wilsonpage/sass-import-once), or [JavaScript](https://github.com/sass-eyeglass/eyeglass#import-once). Alternative it will land in the Sass v4 module system.

AndyOGo:
@xzyfer 
Yeah I know I read through all the `TL DR` stuff start in 2011.

I really see it as a key feature, though there seems to be so much of opposition - very sadly...
To clarify it's a bug in the SASS language, hence libsass suffers from the exact same design decision fault.

---

In case anyone is curious - To Whom It May Concern:
The Issue: https://github.com/sass/sass/issues/139

Available solutions depends on your tool chain:

* **ruby-sass**: https://github.com/Compass/compass/tree/master/import-once
* **node-sass**: https://www.npmjs.com/package/node-sass-import-once

PR #583, opened 2014-10-28
Compass-style import once behavior

sufianrhazi:
This adds [commpass-style import-once behavior](https://rubygems.org/gems/compass-import-once) to libsass, without the exclamation point suffix override.

Though this was a closed issue in #450, we depended on this functionality and using ruby sass was simply not performant enough for our needs.


xzyfer:
As [discussed on twitter](https://twitter.com/sufianrhazi/status/562425848096374785) this belongs in custom importer until modules lands in Sass next.

Issue #643, opened 2014-11-13
nth() doesn't handle maps


xzyfer:
> This can return the nth pair in a map as well. - http://sass-lang.com/documentation/Sass/Script/Functions.html#nth-instance_method

``` scss
$map: (foo: bar, bar: baz);

foo {
  a: nth($map, 2);
}
```

Expected output

``` css
foo {
  a: bar baz;
}
```

Actual output

``` css
stdin:4: index out of bounds for `nth($list, $n)`
```

esr360:
I still seem to be getting this issue, or at least an issue very similar to this, when using Libsass (I don't get the error when using Ruby Sass):

```
>> Error: index out of bounds for `nth($list, $n)`
>>         on line 21 of ../../Synergy/src/scss/mixins/_component.scss
>> >>     $namespace: nth(module-tree(&), length(module-tree(&)));
```

For reference here is the culpable line of code: https://github.com/esr360/Synergy/blob/4715c9ceec3ca1ad664ef0b7a093743b9929e3de/src/scss/mixins/_component.scss#L21

PR #647, opened 2014-11-14
Add support for nth with maps

xzyfer:
This PR add support for `nth` with maps.

> This can return the nth pair in a map as well. - http://sass-lang.com/documentation/Sass/Script/Functions.html#nth-instance_method

Fixes https://github.com/sass/libsass/issues/643. Specs added https://github.com/sass/sass-spec/pull/134.

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -1038,18 +1038,30 @@ namespace Sass {
     Signature nth_sig = "nth($list, $n)";
     BUILT_IN(nth)
     {
+      Map* m = dynamic_cast<Map*>(env["$list"]);
       List* l = dynamic_cast<List*>(env["$list"]);
       Number* n = ARG("$n", Number);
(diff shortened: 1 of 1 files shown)


mgreter:
FYI: I get a success for this test with https://github.com/sass/libsass/pull/1600

xzyfer:
@mgreter I'm not sure what you mean.

mgreter:
I got a positive result with my test runner with that branch!
Wondering why now or put different, why did it not pass before?

xzyfer:
I'm confused because the spec should have always passed. I'm not following
what your PR has to do with this old feature.

mgreter:
Me neither, that's why I posted here, since it seems this is still in todo?

xzyfer:
As far as I can tell this spec (https://github.com/sass/sass-spec/pull/134) as activated 11 months ago.

Issue #318, opened 2014-03-16
Support for in-place importing of CSS


jmagnusson:
This is currently possible with the Ruby version using https://github.com/chriseppstein/sass-css-importer. Would be great to have so we can import CSS from bower libs etc.

mgreter:
https://github.com/sass/libsass/pull/754 has been merged!

josephspens:
Do we know what version this will be available?

rach:
Hi, any idea why it's not currently working with the version of node-sass ( 3.1.2 using libsass 3.2.4 )
I do the import without the file extension.
Thanks.

vanilla-thunder:
the current sass version is 3.4.14 but this feature still seems not to work.
Did someone manage to get that working?

PR #754, opened 2014-12-22
Implement raw css imports

mgreter:
Also refactors the lookup code for better readability!

This is an attempt to fix one of our longer standing issues -> https://github.com/sass/libsass/issues/318
I tried to read all related issues on ruby sass to make this compatible!
This PR allows to import a `file.css` with the following syntax:

```
@import "file";
```

This will try to resolve the partial as normal, but will also look for `.css` files.
Previously this would produces an error if no file could be found.

The old behaviour is still active:
```
@import "file.css";
```

Will result in:
```
@import url(file.css);
```

So this should IMO be pretty much backward compatible!
If this is not acceptable by default I suggest to make this configurable!

@hcatlin what do you think?

### DIFF
--- a/file.cpp
+++ b/file.cpp
@@ -9,6 +9,7 @@
 #include <iostream>
 #include <fstream>
 #include <cctype>
+#include <vector>
 #include <algorithm>
 #include <sys/stat.h>
 #include "file.hpp"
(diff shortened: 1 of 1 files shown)


3den:
when will this change be published on the sass gem?

xzyfer:
I don't believe Libsass is bundled as a ruby gem. If it is it is done so by
someone else.

manuelro:
Was this feature added already? I still getting a `File to import not found or unreadable` while importing a css file from my bower dependencies.

Using:
`@import "angular-material/angular-material.min"`

The load path is right, since I can import `.scss` and `.sass` files.

smartapant:
@mgreter, i have the same problem that @manuelro has, is the feature available?

micahblu:
I'm also getting the same error with a confirmed css path

drewwells:
As the issue says, 

```
The old behaviour is still active:

@import "file.css";
Will result in:

@import url(file.css);
```

This doesn't inline, CSS resolves the path by issuing a get request for the CSS file.

micahblu:
@drewwells That is understood however as I stated is not the intent that the `@import "path/style"` will look for a partial or matching .css file and inline it into the output file?

* Note the missing .css

xzyfer:
@manuelro you'll a recent version of Libsass or node-sass@beta installed to use this feature.

chriseppstein:
> configurable to add additional extensions to the lookup

Please, make the default behavior for libsass match the default behavior for ruby sass. As you point out, it's trivial to add this capability to sass, but it creates an inconsistency in that the behavior of having an extension is different from when it is omitted. We don't like that kind of unpredictable behavior. If people opt-in, they will understand why the behavior is what it is. Because docs.

chriseppstein:
If you want to expose configuration that enables this, like sass does, I think that's ok.

ccraze-matt:
@chriseppstein Does rubygem sass expose a configuration option to allow the **inline** importing of css files?

baubbelayponge:
I can see this was merged but it's not really working on my side.  

I'm using **node-sass 4.5.3** which uses **libsass 3.5.0.beta.2** and  

`@import "~vue-wysiwyg/dist/vueWysiwyg.css";` works fine but  
`@import "~vue-wysiwyg/dist/vueWysiwyg";` does not, saying  

>Error: File to import not found or unreadable: ~vue-wysiwyg/dist/vueWysiwyg.
>Parent style sheet: E:/project/assets/scss/vendor/_vendor.scss
>on line 3 of assets/scss/vendor/_vendor.scss
>@import "~vue-wysiwyg/dist/vueWysiwyg";

Am I doing something wrong?

xzyfer:
Seeing ~ suggests your using an importer or webpack loader. I suspect
that's what is producing the error. I recommend strongly against rely on
this behaviour. It will be removed in future versions.

baubbelayponge:
Not sure what you mean by importer but I'm just using grunt-sass -> node-sass -> libsass to compile the file.  

I've tried with `@import "../node_modules/vue-wysiwyg/dist/vueWysiwyg";` and I have the same problem.

Issue #629, opened 2014-11-06
3.4 support blockers!


hcatlin:
Hey yo! Here is a mega ticket. I just went through all the specs over at sass-spec and made a new branch called "threefour" that is 100% s3.4 compatible in all tests. This is basically the primary hitlist for 3.2 development.

Here are the things I noticed breaking, and the tests that cover them.

* [x] global! variable support
  * [x] spec/basic/30_if_in_function
  * [x] spec/scss/each-in-function

* [x] fractions into decimals
  * [x] spec/libsass-todo-tests/libsass/div
  * [x] spec/libsass-todo-tests/libsass/delayed (*white-space diff)

* [x] when we run into utf-8 characters, we throw in the charset
  * [x] spec/libsass-todo-tests/charset

* [x] handling of at-rules / directives
  * [x] spec/libsass/047_test_unknown_directive_bubbling
  * [X] spec/basic/38_expressions_in_at_directives (* selector order diff)

* [x] `@at-root` support
  * [x] spec/libsass/at-root/135_test_simple_at_root
  * [x] spec/libsass/at-root/136_test_at_root_with_selector
  * [x] spec/libsass/at-root/137_test_at_root_in_mixin
  * [x] spec/libsass/at-root/139_test_at_root_in_bubbled_media
  * [x] spec/libsass/at-root/142_test_multi_level_at_root_with_parent_ref

* [x] hex math being a + #123 being #a123 instead of #a112233
  * [x] spec/basic/16_hex_arithmetic

* [x] weird issue with unquote(0xf00)
  * [x] spec/libsass-todo-tests/libsass/colors

* [x] selector ordering with @extends
  * [x] spec/extend-tests/180_test_basic_extend_loop (* selector order diff)

* [x] when you do 0.00000001 * -1 it should be 0.0 not 0
  * [x] negative_numbers

* [x] ignore attempts at escaping in strings. Aka "\t" should become "t" or "\n" to "n" instead of ""
  * [x] spec/basic/25_basic_string_interpolation

* [x] allow interpolation outside of quotes
  * [x] spec/libsass-todo-tests/libsass/interpolated-urls

alanhogan:
Is this still accurate? There is only one blocker from 3.4 compatibility?

xzyfer:
This is correct as of the current master with regards to major language features. If you want a more specific list of issues look at the github issues.

xzyfer:
[...]

PR #772, opened 2014-12-26
Fully evaluation nested binary operations

xzyfer:
This PR request is an evolution of https://github.com/sass/libsass/pull/770 which cover many more cases where division operations aren't correct evaluated.

Address parts of https://github.com/sass/libsass/issues/629. Specs added https://github.com/sass/sass-spec/pull/201.

I'm going to leave this until 3.1.0 is stable because there's a chance this introduces subtle regressions not covered by our spec suite.


mgreter:
LGTM! :+1:

Issue #694, opened 2014-12-09
Interpolation of quoted strings differs from Ruby SASS


wtfribley:
Given following (albeit wonky) SASS

```css
$str: '{' + '"foo": "bar"' + '}';

.interpolation-test {
  test: "#{$str}";
}
```

When compiled using Ruby:

```css
.interpolation-test {
  test: '{"foo": "bar"}';
}
```

The quoting behavior is different in libsass:

```css
.interpolation-test {
  test: "{"foo": "bar"}";
}
```

While this example may be odd, I think perhaps taking a look at the behavior of quotes when interpolating strings may not be a bad idea (looks like it's come up before in #323).

HugoGiraudel:
Possibly related: https://github.com/sass/libsass/issues/615.

xzyfer:
Not really #615. 

#615 is related to parsing interpolants, this is about preserving quotes styles.

mgreter:
The logic here seems pretty simple. If the string contains double quotes and no single quotes, it is wrapped in single quotes. This seems to be true for all output styles (I checked).

By the way, I used this test to verify the behaviour:

```css
$str01: "{" + '"' + "}";
$str02: '{' + '"' + '}';
$str03: '{' + '"' + "}";
$str04: "{" + '"' + '}';

$str05: "{" + "\"" + "}";
$str06: '{' + "\"" + '}';
$str07: '{' + "\"" + "}";
$str08: "{" + "\"" + '}';

$str11: "{" + "'" + "}";
$str12: '{' + "'" + '}';
$str13: '{' + "'" + "}";
$str14: "{" + "'" + '}';

$str15: "{" + '\'' + "}";
$str16: '{' + '\'' + '}';
$str17: '{' + '\'' + "}";
$str18: "{" + '\'' + '}';

$str21: "{" + "\"'" + "}";
$str22: '{' + "\"'" + '}';
$str23: '{' + "\"'" + "}";
$str24: "{" + "\"'" + '}';

$str25: "{" + '\'"' + "}";
$str26: '{' + '\'"' + '}';
$str27: '{' + '\'"' + "}";
$str28: "{" + '\'"' + '}';

.interpolation-test {
  test01: "#{$str01}";
  test02: "#{$str02}";
  test03: "#{$str03}";
  test04: "#{$str04}";
  test05: "#{$str05}";
  test06: "#{$str06}";
  test07: "#{$str07}";
  test08: "#{$str08}";

  test11: "#{$str11}";
[...]

PR #780, opened 2014-12-27
(DO NOT MERGE) 694 unquoting problems

smikes:
This is a reference PR, not finished yet, and not intended for merging.

I am trying to approach the problems in #694 (interpolation of quoted strings differs from Ruby) by writing some systematic unit tests to cover the `Sass::unquote()` and `Sass:quote()` function.  To write these unit tests, I am using the C++ unit test framework `catch` as a module pulled from `npm`, so I have added a `package.json` -- this file is purely temporary and will not be included in the final PR, but is included for the moment if anyone is interested in playing with catch and/or reproducing my results.

So far I have done the following:

1. factor out `Sass::quote` and `Sass::unquote` into a separate file for testing
2. add unit tests to cover the branches in `Sass::unquote`.

I see at least two problems so far.  One is an underread of allocated memory when a quoted string ~~contains~~ begins with an unescaped quote -- i.e., when calling `Sass::unquote("\"\"Hello\"")`.  When the argument is `""Hello"` then unquote treats it as a quoted string, but then calls `t.erase(t.length()-1)` when `t.length()` is zero.  This is what triggers the exception (`basic_string`) thrown in case # 694.

The second problem is slightly subtler: When unquoting a string, backslash is not treated as a special character.  This causes two problems: one is that backslashes are not unescaped correctly -- `Sass::unquote("'\\\\'")` yields two backslashes, but should only yield one.  The second problem is that unescaped quotes inside strings are not detected; `Sass::unquote` just assumes that the character preceding a quote in a quoted string **is** a backslash, resulting in erroneous unquoting: `Sass::unquote("'foo \'bar' baz'")` yields `foo 'ba' baz` because `unquote` assumes that the `r` before the `'` is a backslash and deletes it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ libsass/*
 *.a
 a.out
 libsass.js
+TAGS
 
 bin/*
 .deps/
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,7 @@ SOURCES = \
 	output_nested.cpp \
 	parser.cpp \
 	prelexer.cpp \
+	quote.cpp \
 	remove_placeholders.cpp \
 	sass.cpp \
 	sass_util.cpp \
--- a/Makefile.am
+++ b/Makefile.am
@@ -62,6 +62,7 @@ libsass_la_SOURCES = \
 	output_nested.cpp output_nested.hpp \
 	parser.cpp parser.hpp \
 	prelexer.cpp prelexer.hpp \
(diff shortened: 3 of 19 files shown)


mgreter:
Unfortunately most of these cases are also covered by #910, so  I'm going to close this. Sorry @smikes, but if you see anything I have missed in that commit, feel free to open more PRs! Thank you!

Issue #771, opened 2014-12-26
source maps ignoring commented out lines


philippkuehn:
Hey, i have a problem to get the source maps working.
 
```scss
@import "base/reset";
    
body {
  color: blue;
}
```

gives me for the body line 3 which is correct.

```scss
// @import "base/reset";
    
body {
  color: blue;
}
```

gives me for the body line 1 which is false.

It seems that all commented out lines are ignored in calculating the line numbers for the source map.

mgreter:
This issue should (hopefully) be addressed/fixed by https://github.com/sass/libsass/pull/792.
Please try with latest master and open a new issue if the problem still exists.
Thank you!

Issue #734, opened 2014-12-16
Unify common code for different output styles


mgreter:
Currently we only support nested and compressed output. But these currently do not share a common parent class. IMO we should create such a base class for all output styles, which will implement all common parts.

https://github.com/sass/libsass/pull/717: Main reasons (beside DRY):

- Abstract the output buffer and provide a simple API
- Source-maps should be created automatically
- The UTF-8 charset should be added automatically
- Makes it easier to implement other output styles

I already made some experiments when implementing the charset feature.
It should be usable as a starting point, if someone else wants to start with it:

```C
#file: output.hpp
#define SASS_OUTPUT

#ifndef SASS_OPERATION
#include "operation.hpp"
#endif

#ifndef SASS_CONTEXT
#include "context.hpp"
#endif

#include <string>

namespace Sass {
  using namespace std;
  struct Context;

  template<typename T>
  class Output : public Operation_CRTP<void, T> {
    // import all the class-specific methods and override as desired
    using Operation_CRTP<void, T>::operator();

  protected:
    Context* ctx;
    string buffer;
    string rendered_imports;
    bool seen_utf8;

    virtual void fallback_impl(AST_Node* n) = 0;

  public:

    Output(Context* ctx = 0)
    : ctx(ctx),
      buffer(""),
      rendered_imports(""),
      seen_utf8(false)
    { }
    virtual ~Output() { };

    string get_buffer()
    {
      return rendered_imports + buffer;
    }

    // append some text or token to the buffer
    // it is the main function for source-mapping
    void append_buffer(const string& data)
    {
      // search for unicode char
      for(const char& chr : data) {
        // abort clause
        if (seen_utf8) break;
        // skip all normal ascii chars
        if (chr >= 0 && chr < 128) continue;
        // prepend charset to buffer
        buffer = "@charset \"UTF-8\";\n" + buffer;
        // singleton
        seen_utf8 = true;
      }
      // add to buffer
      buffer += data;
      // account for data in source-maps
      ctx->source_map.update_column(data);
    }

  };

}
```

I was not able to implement this outside the header, which is probably because my c++ template foo is not very good. Maybe someone else knows if this is the only/right way?

mgreter:
This issue should (hopefully) be addressed/fixed by https://github.com/sass/libsass/pull/792.
I have created a common Output class from which nested/compressed inherit!

Issue #621, opened 2014-11-05
Source Maps break with Autoprefixer


bdkjones:
### Context: 

According to the Source Map 3 spec, the `file` parameter of the source map is: "An optional name of the _generated code_ that this source map is associated with."

### The Issue: 

Libsass 3.0.2 is generating the WRONG value for the `file` parameter and this breaks other tools that try to use the source maps Libsass creates (example: Autoprefixer). See this screenshot:

![screen shot 2014-11-04 at 22 43 24](https://cloud.githubusercontent.com/assets/926566/4913748/bee94bf0-64b7-11e4-8000-9c0c96aa7bf3.png)

The `file` parameter SHOULD be simply `app.css`, because that's the output file and that output file is right next to the source map file on disk.

However, Libsass is incorrectly prepending `../scss/` to the beginning of this `file` parameter. This is the path to the folder where the original Sass file resides (as shown in the Finder). It should NOT be used here, because the Source Map spec states that the `file` parameter is the OUTPUT file, not the input file.

### What This Breaks:

This incorrect value does not seem to affect browsers; the source maps are still loaded correctly. HOWEVER, other tools like Autoprefixer that need to read a source map and then modify it, ARE broken by this bug. The source map no longer works after Autoprefixer modifies it. (See post below for details.)

mgreter:
This issue might (hopefully) be addressed/fixed by https://github.com/sass/libsass/pull/792.
Please try with latest master and open a new issue if the problem still exists.
Thank you!

Issue #605, opened 2014-11-01
[RFC] Source maps - General


am11:
(Redirected from: https://github.com/sass/libsass/pull/601#issuecomment-61199911, credit to @mgreter)

This is a general source-maps issue. It covers all the source maps progress. Going by LESS CLI, it provides these options:

```bash
 --source-map[=FILENAME]  Outputs a v3 sourcemap to the filename (or output filename.map)
 --source-map-rootpath=X  adds this path onto the sourcemap filename and less file paths
 --source-map-basepath=X  Sets sourcemap base path, defaults to current working directory.
 --source-map-less-inline puts the less files into the map instead of referencing them
 --source-map-map-inline  puts the map (and any less files) into the output css file
 --source-map-url=URL     the complete url and filename put in the less file
```

> *the description of last line is confusing, it is basically for `sourceMappingURL` in output CSS file. see https://github.com/less/less.js/issues/2264*

Here we will investigate the progress being made for open issues: https://github.com/sass/libsass/issues/498, https://github.com/sass/libsass/issues/444, https://github.com/sass/libsass/issues/324 and https://github.com/sass/libsass/issues/212.

Following are the requirements:

- [ ] Report correct columns and lines, matching the "index base" for of other transcompilers: ruby-sass, less, coffeescirpt and since so forth. 
- [ ] Capture mapping for each token. In the code: `body, nav { background: linear-gradient(top, left, #111) center no-repeat #111 }`, there are 14 tokens: `body`, `,`, `nav`, `{`, `background-image`, `:`, `linear-gradient`, `top`, `left`, `#111`, `center`, `no-repeat`, `#111` and `}`.
- [x] Option for map JSON to be encoded in base64 and set to sourceMappingURL's value in CSS (as opposed to exist as a separate file). An equivalent of LESS' `source-map-map-inline`.
- [x] Option to add source-code in `sourceContents` array in map JSON. Answer to LESS' `source-map-less-inline`. This comes handy when the source files are not accessible remotely (HTTP?).
- [ ] Ability to set custom URL for sourceMappingURL. if this option is set, it will use the string provided by user as is. Answer to LESS' `source-map-url`.
- [ ] Add tests; which decode B64 and B64VLQ first and then match the mappings. Also test paths in sourceMappingURL comment in CSS, sources array and file elements in map's JSON.

mgreter:
@am11 can we close this and make more clear targets from this?
[...]

PR #792, opened 2014-12-31
Add improved source-mapping implementation

mgreter:
This PR is the successor of #603 and #786. IMO this is already better than the current implementation, so I dare anyone to try it out. IMO we could include this "as is" in the next version to get this moving in the right direction, since the current implementation does not really work anyway IMHO. The whole feature is very complex, so I cannot give any guarantee that it will work for all mappings. But IMHO the current state of this PR is satisfactory and we could move on from that!

### DIFF
--- a/sass_interface.cpp
+++ b/sass_interface.cpp
@@ -142,7 +142,7 @@ extern "C" {
     }
     catch (Sass_Error& e) {
       stringstream msg_stream;
-      msg_stream << e.path << ":" << e.position.line << ": " << e.message << endl;
+      msg_stream << e.pstate.path << ":" << e.pstate.line << ": " << e.message << endl;
       c_ctx->error_message = strdup(msg_stream.str().c_str());
       c_ctx->error_status = 1;
       c_ctx->output_string = 0;
@@ -229,7 +229,7 @@ extern "C" {
     }
     catch (Sass_Error& e) {
       stringstream msg_stream;
-      msg_stream << e.path << ":" << e.position.line << ": " << e.message << endl;
+      msg_stream << e.path << ":" << e.pstate.line << ": " << e.message << endl;
       c_ctx->error_message = strdup(msg_stream.str().c_str());
       c_ctx->error_status = 1;
       c_ctx->output_string = 0;
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,7 @@ SOURCES = \
 	output_compressed.cpp \
 	output_nested.cpp \
 	parser.cpp \
+	position.cpp \
 	prelexer.cpp \
 	remove_placeholders.cpp \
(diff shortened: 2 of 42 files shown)


mgreter:
I have added a simple "test runner" to perl-libsass while developing this feature to keep at least some track of what I was doing. @am11 you probably know how hard it is to debug this kind of stuff. I currently (not published yet) store this directory structure as the "sourcemap tests":

- config
- input.s[ca]ss
- output.css
- output.css.map

The config file is a map to context options (C-API options in a colon separated map), while the rest should be self explanatory. The output.css.map file is parsed and the stuff like included paths should be compared. I currently only implemented to compare the mappings, which does a search for expected mappings only, while ignoring additonal ones! This proved to be pretty effective during developement, since I was able to update the expected file incrementally (after inspecting and verifying the specific cases/mappings manually!).

Basically this topic is endless since one could discuss about how a mapping should be done is pretty open (how do you map a calculation? Include every operator, number and so on?). This PR is already at a point where I would say is at least 50% better than before. Some code is certainly not optimal (reparse/remap part), but I don't really know how to improve it much more!  And it gets tedious to maintain such a big commit! This particular feature (remap) would not have worked bevore anyway; so I guess I'm willing to take the risk that it will now only work a few percent better. This might be 100% or 0.05%!

I guess not many people really want to try it out and report back specific issues here! So I only can encourage the "first scream" approach! Just get it merged into 3.2 and let's see :banana:

sass_interface.cpp
```diff
@@ -140,7 +140,7 @@ extern "C" {
     }
     catch (Sass_Error& e) {
       stringstream msg_stream;
-      msg_stream << e.path << ":" << e.position.line << ": " << e.message << endl;
+      // msg_stream << e.pstate.path << ":" << e.pstate.line << ": " << e.message << endl;
```
mgreter: Todo: Revert!

mgreter:
This should (hopefully) fix:

- https://github.com/sass/libsass/issues/324
- https://github.com/sass/libsass/issues/621
- https://github.com/sass/libsass/issues/605
- https://github.com/sass/libsass/issues/771
- https://github.com/sass/libsass/issues/469
- https://github.com/sass/libsass/issues/734

am11:
Today is turning out to be a big achievement day! :+1: :star: 

Congratulations! :)

hcatlin:
WEEEEEEE

xzyfer:
Nice work!

mgreter:
Thx all! Just to be perfectly clear, this does not magically fix all of our source-map problems. This is merely a step in the right direction to provide a base to make this happen. I only verified a few common complicated cases. Most of the rest should behave as before plus a few additional perks.

PR #795, opened 2014-12-31
Fix makefile portability 

mgreter:
Improve compatibility with mingw

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,16 +193,10 @@ lib/libsass.dll: lib $(COBJECTS) $(OBJECTS) $(RCOBJECTS)
 
 install: install-$(BUILD)
 
-$(DESTDIR):
-	$(MKDIR) $(DESTDIR)
-
-$(DESTDIR)$(PREFIX): $(DESTDIR)
-	$(MKDIR) $(DESTDIR)$(PREFIX)
-
-install-static: $(DESTDIR)$(PREFIX) lib/libsass.a
+install-static: lib/libsass.a
 	install -pm0755 $< $(DESTDIR)$(PREFIX)/$<
 
-install-shared: $(DESTDIR)$(PREFIX) lib/libsass.so
+install-shared: lib/libsass.so
 	install -pm0755 $< $(DESTDIR)$(PREFIX)/$<
 
 $(SASSC_BIN): $(BUILD)
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ CC       ?= gcc
 CXX      ?= g++
 RM       ?= rm -f
 CP       ?= cp -a
-MKDIR    ?= mkdir -p
+MKDIR    ?= mkdir
 WINDRES  ?= windres
 CFLAGS   ?= -Wall -O2
 CXXFLAGS ?= -Wall -O2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -167,16 +167,16 @@ debug-shared: shared
 static: lib/libsass.a
 shared: $(LIBRARIES)
 
-lib/libsass.a: $(COBJECTS) $(OBJECTS)
+lib:
 	$(MKDIR) lib
```
mgreter: Unfortunately mkdir -p does not work on windows. But maybe you can confirm if it is legal/ok to have directories as targets in makefiles. Problem was that mkdir on windows reports unsuccessful exit status when it tries to create the same directory again ... still not sure what the best portable way is.

am11: (not sure if its applicable)
`mkdir C:\libsass 2> NUL`, redirect stderr to NUL (something analogous to, but not exactly, devnull?).

So even if it throws, it will keep continue.

More at: http://stackoverflow.com/q/4165387.

xzyfer: You might be able to use the `@` prefix which will suppress any errors caused by `@mkdir ..` and continue on as normal. Not sure if that works on windows.

mgreter: The problem is that mkdir returns a failed exit state if it tries to re-create an already existing directory (IMO that is 100% messed up by Windows/MS). And there doesn't seem to be an easy portable way to test for existing directories! But it seems that we can also use directories as targets within makefiles, which would make the check for existing directories obsolete! If anyone has a better solution for this, please enlighten me!

xzyfer: > The problem is that mkdir returns a failed exit state if it tries to re-create an already existing directory

Do we care if the directory already exists? I assume if the directory exists we can just continue on as normal correct?

> And there doesn't seem to be an easy portable way to test for existing directories!

Assume the above is correct I think you can do the following `@$(MKDIR) lib || true`. This will just will suppress the error that the direct already (and any other errors unfortunately).

mgreter: Nope, we do not care if the directory already exists, but the make process aborted if it tried to "re-create" an already existing directory. And there is no direct option to surpress it (the -p option lead to create directories named "-p" on windows). For now I resolved to create targets for the directory in the makefile. AFAIK these targets will only be executed if the path/file does not exist. I know that this is ok for files, but I'm not sure if this is also true for directories by definition! At least this PR seems to work on linux, mingw (windows) and mac os x!

Just tried your suggestion with `@$(MKDIR) lib || true` and it seems to be supported by windows! Seems like another possible solution to this problem. But still not sure what the best and most portable solution is!

QuLogic: Usually the problem with creating directories in makefiles is that it's
racy. But since you have a separate rule for it, that might be okay (I
forgot to mention that initially; sorry if I made it sound too big). I
_think_ there should not be a problem with having makefile targets depend
on the directory (as long as there's only one), but I normally just put
`.keep` in them to ensure they exist.

PS, the `@` prefix suppresses output; the `-` prefix ignores errors.

mgreter:
The basic problem was that mkdir on windows reported an error when the directory already exists, and there seems to be no other solution beside not calling the command if the directory already exists (this really only matters on windows/mingw!). Next problem is that this check is not really possible in a portable way (maybe the `-` prefix as mentioned could come handy here, but I'm not convinced that this is supported by mingw32-make or dmake!?). Naming the directory as a target seems to fix it with all implementations. "Seems", since I never read a statement that this would be guaranteed! Just to make it clear, what I mean is this:

```
directory: 
  mkdir directory
all: directory
  make
```

So the `directory` would only be created once if it did not already exist!

QuLogic:
As long as there's only _one_ rule that creates the directory, I think there should be no problems due to race conditions. However, there is still the minor issue that the targets now depend on the timestamp of the directory and not merely its existence. But apparently you can use an [order-only prequisite](http://stackoverflow.com/a/6170280) which should convey that intent.

Issue #783, opened 2014-12-27
px/em unit calculations outputting incorrect units.


browniefed:
From sass/node-sass#585

A few compass translations to raw scss have a hide-text mixin
```css
    $approximate-em-value: 12px / 1em;
    $wider-than-any-screen: -9999em;
    text-indent: $wider-than-any-screen * $approximate-em-value;  
```
In node-sass 1.2.2 (unsure of libsass version at that time) the calculation would return just the px unit like `-119988px` however in 2.0 beta the result is `-119988em*px/em`

sass/sass-spec#212

xzyfer:
This is fixed and will be available in the first patch release of 3.1.

PR #797, opened 2015-01-01
Properly hydrate numbers after operations in variable assignments

xzyfer:
This PR fixes an issue where a number variable's units would be mangled after a binary operation i.e. `*`, `/`.

I've also updated the number validity check and error message to be more inline with Ruby sass.

Fixes https://github.com/sass/libsass/issues/783. Specs added https://github.com/sass/sass-spec/pull/216.

### DIFF
--- a/eval.cpp
+++ b/eval.cpp
@@ -591,11 +591,8 @@ namespace Sass {
     // behave according to as ruby sass (add leading zero)
     if (value->concrete_type() == Expression::NUMBER) {
       Number* n = static_cast<Number*>(value);
-      value = new (ctx.mem) Number(n->path(),
-                                   n->position(),
-                                   n->value(),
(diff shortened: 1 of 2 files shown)


am11:
:+1: 

@xzyfer, (just curious) what about those commented lines in the test? Are they part of the spec?

xzyfer:
Those lines cause errors which we cannot assert yet. I've left them in with
the appropriate error message for when we can.

am11:
Awesome! Apparently, number theory is (also( becoming a strong suit of libsass. :sunglasses: 

On a separate note, would be nice if you guys can setup gitter (like node-sass channel there: https://gitter.im/sass/node-sass), so peeps can throw  in questions, which you guys can answer later. IRC is another way to go. :)

Issue #808, opened 2015-01-05
Not all headers have include guards


mgreter:
IMO it is common standard in C++ to add an include guard to each header to prevent multiple inclusions. Libsass seems to have adopted an IMHO strange use of this feature, as the headers do define a guard, but the checks seem to be done in the actual code files and not in the header itself.

The expected format would be:

```C++
#ifndef SASS_MODULE
#define SASS_MODULE
// header code ...
#endif
```

//CC @xzyfer If you agree I would add this to all headers in one big commit!? Or do you see any reason not to it this way (beside minor compile time increases)?

akhleung:
Yeah, I had been meaning to make this change -- I didn't know what the convention was when I started the project!

xzyfer:
Yes please! I've been doing it as I touch files

mgreter:
I am about ready to merge my work on source maps. Want to wait till this one also touches a lot of files. @xzyfer Do you want me to wait till you merge the stuff for @at-root? Just want to prevent too many conflicts when rebasing!

xzyfer:
That would be great. I'll be merging at-root in an hour or two.

mgreter:
:+1:

xzyfer:
Actually.. If you're not changing indentation of files I don't expect any
conflicts with this. Just do it and I'll deal with it

mgreter:
I'm also going to remove to guards in all cpp files! That could give some conflicts!

xzyfer:
Should be fine. They're easy yo resolving and I have some rebasing to do
any way.

mgreter:
OK, I guess I will merge source-maps now; so I can get right to it!

xzyfer:
Ah ok. I thought you were talking about shipping the guards. I don't think sourcemaps will conflict too much since at-root is mostly new code, but it may break sourcemaps :/

xzyfer:
Hmm your selector stuff might be an issue for at-root. Do you want to hold off on sourcemaps until I merge at-root? I'm going to be at a bit of a loss at how to get my stuff up to date with you change.

It'd be worth doing a wiki page about how to work with sourcemaps (in code) and typical gotchyas.

mgreter:
The changes are pretty simple IMO, you mostly just need to replace `path, position` with `pstate` for `AST_Node` creation. Some parts might be trickier, but IMO there is no chance to get the source-map part merged in a 100% state (it's still better than current).

xzyfer:
:+1:

PR #809, opened 2015-01-05
Refactor header include guards

mgreter:
Fixes https://github.com/sass/libsass/issues/808


QuLogic:
You might want to suffix the names with `_H` or similar (just because something like `SASS_CONTEXT` seems a bit generic.)

mgreter:
Good point, will update this PR now!

xzyfer:
This is a thing of beauty.

Issue #801, opened 2015-01-04
v3.1 doesn't compile under Windows and MSVC


patlecat:
Just downloaded the 3.1.0 tagged version, unzipped it and tried to compile *libsass.sln* under Windows 8.1 64bit with MSVC2013 and got those errors/warnings straight away:

```
1>------ Build started: Project: libsass, Configuration: Debug x64 ------
1>  sassc.c
1>c1 : fatal error C1083: Cannot open source file: '..\sassc\sassc.c': No such file or directory
1>  cencode.c
1>..\cencode.c(43): warning C4244: 'return' : conversion from '__int64' to 'int', possible loss of data
1>..\cencode.c(54): warning C4244: 'return' : conversion from '__int64' to 'int', possible loss of data
1>..\cencode.c(65): warning C4244: 'return' : conversion from '__int64' to 'int', possible loss of data
1>..\cencode.c(77): warning C4244: 'return' : conversion from '__int64' to 'int', possible loss of data
1>..\cencode.c(100): warning C4244: 'return' : conversion from '__int64' to 'int', possible loss of data
1>  getopt.c
1>..\posix\getopt.c(348): warning C4996: 'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
1>          C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\stdlib.h(462) : see declaration of 'getenv'
1>  Generating Code...
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

Then I removed the non-existant sassc.c file from the project and got this error:

```
1>MSVCRTD.lib(crtexe.obj) : error LNK2019: unresolved external symbol main referenced in function __tmainCRTStartup
1>C:\Users\xxx\Sass\libsass-3.1.0\win\bin\Debug\sassc.exe : fatal error LNK1120: 1 unresolved externals
```

QuLogic:
I don't use Windows, but it appears the solution has the `sassc` executable in the build. You also need sassc sources: https://github.com/sass/libsass/wiki/Building-on-Windows#building-via-visual-studio-community-2013

am11:
@QuLogic is correct. You can either use the git clone way (for which please refer to @QuLogic's link) or download [sassc](https://github.com/sass/sassc/) and extract the folder in libsass directory and compile.

[...]

PR #292, opened 2014-02-20
[DNM] CMake build files and directory structure

nschonni:
# DO NOT MERGE
The commits really need to be cleaned up before this can be merged.

- Replace Make/autotools with [CMake](http://cmake.org) for cross platform building.
- Loose layout based on https://code.google.com/p/cpp-library-project-template/
- Each namespace moved to folders below `src`
- Adds version information 
- Testing is invoked through CTest for now, but this only gives the full output for the specs when using the verbose flag
- Existing test harness under the test folder all pass currently because CTest expects a non-zero exit for failures
- Base setup for Doxygen generation, but not fully plumbed

TODO
 - [x] Rebase down commits
 - [x] Remove Boost/Unit++ commented out code
 - [ ] Look at replacing GLOB_RECURS for tests
 - [x] Update all license top license lines with actual content
 - [x] Exclude generated version info file
 - [ ] Emscripten binding using https://github.com/kripken/emscripten/blob/master/cmake/Platform/Emscripten.cmake

PR #810, opened 2015-01-06
Build: Separates libsass and sassc solution files

am11:
This addresses at least one concern discussed here https://github.com/sass/libsass/issues/801.

To generate static library, open `win/libsass.sln`, right-click libsass in Solution Explorer and open Project Properties page > Configuration Properties > General and on right hand pane, change Configuration Type to Static Library (.lib).

@mgreter, now we have two sets of files; libsass generates a `dll` and sassc generates `.exe`. IMO, sassc set (`.vcxproj`, `.sln` and `.filters`)  should live in https://github.com/sass/sassc. What do you think?

//cc @patlecat, @KaeNote

On a related note, if you guys want them to ship compiler separately (independent of entire IDE) like I do, then please vote for this: http://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/6742251-make-c-c-compiler-cl-exe-independent-of-ide.

---- 

**Update:**

Separates libsass and sassc solution files. …  
* libsass solution lives in libsass, which generate
  a dll.
* sassc solution lives in sassc repository, which
  generates an exe.
* Changes output directory for sassc project to
  sassc/bin, instead of sassc/win/bin. Debug build
  is not created in a separate Debug folder.
* Updates Appveyor accordingly.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -4,7 +4,7 @@ Libsass
 by Aaron Leung ([@akhleung]) and Hampton Catlin ([@hcatlin])
 
 [![Linux CI](https://travis-ci.org/sass/libsass.png?branch=master)](https://travis-ci.org/sass/libsass)
-[![Windows CI](https://ci.appveyor.com/api/projects/status/s88o0u5qra2ng4vy/branch/master?svg=true)](https://ci.appveyor.com/project/sass/libsass/branch/master)
+[![Windows CI](https://ci.appveyor.com/api/projects/status/github/sass/libsass?svg=true)](https://ci.appveyor.com/project/sass/libsass/branch/master)
 [![Bountysource](https://www.bountysource.com/badge/tracker?tracker_id=283068)](https://www.bountysource.com/trackers/283068-libsass?utm_source=283068&utm_medium=shield&utm_campaign=TRACKER_BADGE)
 [![Coverage Status](https://img.shields.io/coveralls/sass/libsass.svg)](https://coveralls.io/r/sass/libsass?branch=feature%2Ftest-travis-ci-3)
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ cache:
   - C:\Ruby21-x64\bin
 
 install:
-  - git clone https://github.com/mgreter/sassc.git --branch feature/build-dll-win-mingw
+  - git clone https://github.com/sass/sassc.git
   - git clone https://github.com/sass/sass-spec.git
   - set PATH=C:\Ruby%ruby_version%\bin;%PATH%
   - set SASS_LIBSASS_PATH=..
--- a/sass_context.cpp
+++ b/sass_context.cpp
@@ -154,6 +154,7 @@ extern "C" {
     void* root;
   };
 
+  static void copy_options(struct Sass_Options* to, struct Sass_Options* from) { *to = *from; }
 
   #define IMPLEMENT_SASS_OPTION_ACCESSOR(type, option) \
     type ADDCALL sass_option_get_##option (struct Sass_Options* options) { return options->option; } \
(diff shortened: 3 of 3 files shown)


mgreter:
First thanks for taking the initiative here!

IMO the build file for sassc should depend somehow on the libsass build. I don't know very much about MSVC, so I don't know how this is normally done. But IMHO best would be if we could mimic the BUILD variable somehow (BUILD=static|shared). I fully agree that the MSVC build for sassc should go into sassc repo itself. But you need to make sure that AppVeyor CI still compiles (which is a bit of a pain to do, since you have to test it with a local repo that checks out the sassc dependency also from your local repo).

am11:
@mgreter, its fixed. To make this work (and pass CI), https://github.com/sass/sassc/pull/91 needs to be merged first. Check the update in first comment above.

Please also review the commit messages. //cc @xzyfer yay! isn't that we always wanted!  :D

mgreter:
Actually I don't think that we want to reference each libsass source file in sassc MSVC build!
Isn't there a way to at least depend on the parent build file or to link to the library otherwise?

am11:
@mgreter, this is case, libsass.sln is not used by CI build at all. This makes libsass package independent. Anyone can download zip, open win/libsass.sln press ctrl+f5 and boom: they got the dll in win/bin/.

However, the one in sassc is bit different. It requires sassc folder to be present (cloned) at libsass root. Then it generates exe file in libsass/sassc/bin/. So separation of concerns is there. 

IMO, this is more "separation of concern" oriented approach.

But if we want to consume dll file generated by libsass.sln or the build references across the project, then it would be a tricky scenario (may require some code change)! I am not expert on MSBuild system by any means. Lets see if @darrenkopp can shed some light on that?

@darrenkopp, what we have with this commit is:

* clone libsass, cd libsass
* clone sassc (inside libsass)
* both libsass and sassc has folder called win with sloluton and project files.
* the one in libsass/win is libsass.sln (and vcproj with same name), which generates dll file.
* the one in (libsass/)sassc/win is sassc.sln, which generates exe file.

What we want is to without making code changes, have a dependency of one vcproj on another, in such a way that the libsass references appear in only libsass.vcproj and sassc.proj depend on those references. Is it possible?

mgreter:
I guess you are facing the same underlying problems as the linux build. The linux builds uses a `SASS_LIBSASS_PATH` env variable to tell the `sassc` build where to look for `libsass`. This is only different with autotools and shared (system) library. I guess there is no clean approach in regard to system libraries on windows, so you should be save to make some assumptions on where the `libsass` and `sassc` code is check out to. If this would be configurable somehow, even better :smile:

am11:
@mgreter, this worked: :zap: 

```
  <ClCompile Include="..\..\*.cpp" />
  <ClCompile Include="..\..\*.c" />

and

  <ClCompile Include="..\..\*.hpp" />
  <ClCompile Include="..\..\*.h" />
```

for sassc/win/sassc.sln and changed `..\*.` for libsass/win/libsass.sln. :)

I think we can also define environment variable in powershell and then use it in MSBuild! 

Doesn't cmake unify Make and NMake? We can generate the nmake of this file and then do scripting in cmake. (For the record, I have never used cmake, just nmake once or twice) :)

mgreter:
I guess the wildcard solution is better than nothing :smile:  CMake should be able to create build files for native Makefiles (linux/mingw), autotools and MSVC. FYI: There is a old PR (https://github.com/sass/libsass/pull/292) that we keep around that could be used as a base to get CMake working.

am11:
@mgreter, according to [this](http://www.cmake.org/Wiki/CMake_Generator_Specific_Information), Make and NMake will work with CMake, so we can add NMake later. :)

For now, to make this work on appveyor, we first need to merge https://github.com/sass/sassc/pull/91, then this PR (to take changes from appveyor.yml) and then rerun the build on appveyor to confirm if it is working (which I am pretty sure will work). :wink:

mgreter:
I have merged your `sassc` PR. IMHO the wildcard solution is not optimal but still a step forward. Would be very nice of you if you also could update and verify the building steps at the wiki, once everything is merged. :stuck_out_tongue_winking_eye:

am11:
Sure thing, will update the wiki accordingly. I promise!! :smile:

darrenkopp:
Yeah, there isn't a better way to have it dynamically update outside of using wild cards. However, you could still use wildcards but constrain them to tighter directories, but then you get to a similar situation as the first one where adding files in new directories could cause them to not be picked up.

am11:
Thanks @darrenkopp! For that matter, I propose that we restructure the repo like [libuv](https://github.com/libuv/libuv), or even better:
  
```
root
  \
  other-directories
  \
  src
      \
       header {contains `.hpp` and `.h` files}
      \
       code {contains `.cpp` and `.c` files}
```

And header and code directories should always include those files which are required for the compilation.

One benefit would be to keep the root cleaner and hence on GitHub, readme would be visible (without major scrolling down). :)

am11:
@mgreter, is there anything that i need to update for this? I think I would have to update the wiki after this is merged right? :)

Issue #815, opened 2015-01-07
str-slice returns incorrect value when last $end-at is omitted


xzyfer:
```scss
foo {
    foo: str-slice("bar", 1, 2);
    bar: str-slice("bar", 3);
}
```

expected

```css
foo {
  foo: "ba";
  bar: "r"; }
```

actual

```css
foo {
  foo: "ba";
  bar: ""; }
```

This is a regression in 3.1.0. Specs added https://github.com/sass/sass-spec/pull/225.

PR #816, opened 2015-01-07
str-slice() $end-at should default to -1 if omitted

xzyfer:
This PR fixes a regression in `str-slice()` when `$end-at` is omitted.

Fixes https://github.com/sass/libsass/issues/815. Specs added https://github.com/sass/sass-spec/pull/225.

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -904,7 +904,10 @@ namespace Sass {
 
         // `str-slice` should always return an empty string when $end-at == 0
         // `normalize_index` normalizes 1 -> 0 so we need to check the original value
-        if(start == end && m->value() > 0) {
+        if(m->value() == 0) {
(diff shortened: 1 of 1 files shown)


functions.cpp
```diff
@@ -904,7 +904,10 @@ namespace Sass {
 
         // `str-slice` should always return an empty string when $end-at == 0
         // `normalize_index` normalizes 1 -> 0 so we need to check the original value
-        if(start == end && m->value() > 0) {
+        if(m->value() == 0) {
+          if(!quotemark) return new (ctx.mem) Null(pstate);
+          newstr = "";
+        } else if(start == end && m->value() != 0) {
```
xzyfer: ` && m->value() != 0` is redundant. I'll clean this up next time.

PR #834, opened 2015-01-12
Remove the non-standard image-url function

xzyfer:
This PR removes the non-standard `image-url` function as per discussion in https://github.com/sass/libsass/issues/489.

Calling `image-url` will now result in an error message informing the user this function no longer exists. 
I felt this was more developer friendly than the standard fallback of transparently outputting the function signature to css.

An alternative approach would be to produce a warning in the console and then do the normal fallback behaviour?

Thoughts @mgreter @hcatlin ?

Fixes https://github.com/sass/libsass/issues/489. Corresponding sassc change sass/sassc#92.


drewwells:
:100: yay!

Issue #585, opened 2014-10-28
Should libsass have a compact function?


xzyfer:
I've noticed libsass defines a `compact($values...)` function. Ruby sass doesn't appear to now or ever had this function. Since we're aiming for 3.4 parity this should probably go.

HugoGiraudel:
`compact` has never been part of Ruby Sass. It is an old function from Compass.

davidangel:
I'm confused by this, isn't 'compact' documented here: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style

Why wouldn't all of these options get included?

xzyfer:
@davidangel that's the compact output style. This is in regards to a native `compact` function that Sass currently does not have http://sass-lang.com/documentation/Sass/Script/Functions.html

davidangel:
@xzyfer so is there support for the compact output style currently? I've been trying to use it with gulp-sass/nose-sass without much luck...

xzyfer:
No there is not. We only support nested and compressed. Feel free to open an issue to support that output style.

PR #835, opened 2015-01-12
Remove the non-standard compact function

xzyfer:
This PR removes the non-standard `compact` function.

This PR exhibits the same behaviour as https://github.com/sass/libsass/pull/834 and will be adjusted to match the final decision made there.

Fixes https://github.com/sass/libsass/issues/585.

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -1224,25 +1224,8 @@ namespace Sass {
     Signature compact_sig = "compact($values...)";
     BUILT_IN(compact)
     {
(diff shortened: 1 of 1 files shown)


xzyfer:
Any objections to shipping this?

hcatlin:
Hmm... makes me think maybe we should as Sass if they will add this feature.

But, remove it for now, as it's not covered in Sass.

akhleung:
I believe variadic argument lists were supposed to make this function unnecessary.

Issue #615, opened 2014-11-03
Interpolation inside an interpolation isn't handled correctly


xzyfer:
The following results in an `invalid property name` error.

```scss
$foo: "bar";
%#{"foo--#{$foo}"} {
    foo: bar;
}

a { 
  @extend %foo--bar; 
}
```

I suspect this is partly responsible for https://github.com/sass/libsass/issues/439 as well.

Specs added https://github.com/sass/sass-spec/pull/123.

jmhe:
Created a pull request for this:

https://github.com/sass/libsass/pull/854

mgreter:
This should be solved by #910!

PR #847, opened 2015-01-19
Improve interpolate parsing and unquote output

mgreter:
This is an extracted patch from my current work on the output part.
- It makes the parsing for interpolations more robust
- Adds correct unescaping in unquote (unicode chars)

This should fix the following todo tests:
(It does at least with my WIP branch ...)
- libsass-todo-tests/25_basic_string_interpolation
- libsass-todo-tests/42_css_imports
- libsass-todo-tests/unquote

And there's more where this is coming from :wink:
~~@xzyfer travis seems to break on the [newest added test] [1] !?~~ 

[1]: https://github.com/sass/sass-spec/commit/ddbd785cbb351b69635a94930dbc0c870e0add2a

### DIFF
--- a/prelexer.cpp
+++ b/prelexer.cpp
@@ -84,54 +84,6 @@ namespace Sass {
       >(src);
     }
 
-    // Match double- and single-quoted strings.
-    const char* double_quoted_string(const char* src) {
-      src = exactly<'"'>(src);
-      if (!src) return 0;
-      const char* p;
-      while (1) {
-        if (!*src) return 0;
-        if((p = escape(src))) {
-          src = p;
-          continue;
-        }
-        else if((p = exactly<'"'>(src))) {
-          return p;
-        }
-        else {
-          ++src;
-        }
-      }
-      return 0;
-    }
-    const char* single_quoted_string(const char* src) {
-      src = exactly<'\''>(src);
-      if (!src) return 0;
-      const char* p;
-      while (1) {
-        if (!*src) return 0;
-        if((p = escape(src))) {
-          src = p;
-          continue;
-        }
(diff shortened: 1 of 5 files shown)


prelexer.cpp
```diff
@@ -686,6 +685,7 @@ namespace Sass {
     const char* static_component(const char* src) {
       return alternatives< identifier,
                            static_string,
+                           percentage,
```
xzyfer: Nice catch

prelexer.hpp
```diff
@@ -83,6 +83,58 @@ namespace Sass {
       }
     }
 
+    // Match a sequence of characters delimited by the supplied chars.
```
xzyfer: It's worth mentioning what makes this `smart`

prelexer.cpp
```diff
@@ -130,7 +130,7 @@ namespace Sass {
 
 
     const char* interpolant(const char* src) {
-      return delimited_by<hash_lbrace, rbrace, false>(src);
```
xzyfer: IMO almost all uses of `delimited_by` are erroneous. We should probably aim to wholesale replace this with something smarter. 

https://github.com/sass/libsass/issues/615 is good example where `delimited_by` fails spectacularly.

xzyfer:
:+1:

mgreter:
Another improvement I was able to pull out of my WIP branch. Fixes another todo issue:
- libsass-todo-tests/selector_interpolation_in_string/input.scss

mgreter:
I'm going to merge this now. I already have some other PRs pending, which will adress/improve the naming of the prelexer functions (which @xzyfer correctly pointed out with `smart_del`)

PR #848, opened 2015-01-19
README: Updates compiler versions

am11:
* GCC 4.5 -> 4.6.
* Adds instructions for Windows.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -15,7 +15,9 @@ then go to https://github.com/sass/sassc or
 https://github.com/sass/ruby-libsass or
 [find your local implementer](https://github.com/sass/libsass/wiki/Implementations).
 
-LibSass requires GCC 4.5+ or Clang/LLVM. If your OS is older, this version may not compile.
+LibSass requires GCC 4.6+ or Clang/LLVM. If your OS is older, this version may not compile.
+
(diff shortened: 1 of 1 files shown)


xzyfer:
Thanks!

am11:
:tada:

patlecat:
Can I ask why libsass wants to be compatible with such an old compiler? What is the point?

xzyfer:
It enables us to reach the largest audience. In our experience people tend
not to update their compilers. More often than not we're dealing with the
OS default. Notably older CentOS users have very old compilers, lower than
even we support.

patlecat:
Ah that's why you demand the use of Visual Studio 2013 Upd4 (the latest) just to support a larger audience? LOL No seriously, why do you do it? Why not just say GCC 4.8.2 is the minimum?

am11:
AFAICT libsass doesn't compile below VS2013 Update 2. The C99, C++11 features were not available in older iterations of VS (see [this vcblog entry](http://blogs.msdn.com/b/vcblog/archive/2013/12/02/c-11-14-core-language-features-in-vs-2013-and-the-nov-2013-ctp.aspx)).

Besides GCC v4.6 *is* the minimum version which can compile libsass. Why should we misguide the users? It does not make sense.

Issue #859, opened 2015-01-27
Segmentation fault with specific `@media` and `@at-root` combinations


anlutro:
I'm using sassc (compiled from master today) and ran across a segmentation fault with a specific combination of `@media`, selectors and `@at-root`.

This is the minimal code I've been able to reproduce the segmentation fault in.

```sass
@media screen
	.two
		@at-root .one
			background: blue
			.three
				color: red
```

```
$ sassc ./test.sass 
Segmentation fault
```

Removing any bit of this example makes it compile just fine - it's something about the combination of rules (`background: blue`) and selectors (`.three`) within the `@at-root` declaration that's messing up.

xzyfer:
I've confirmed this is an issue on the latest master

```scss
@media screen {
  .two {
    @at-root .one {
      background: blue;
      .three {
        color: red;
      }
    }
  }
}
```

HugoGiraudel:
@xzyfer Does this mean I missed a case with my `@at-root` test suite?

xzyfer:
It would appear so.

xzyfer:
I've got a fix https://github.com/sass/libsass/pull/861.

I don't have time to create the spec now, but when I do this is ready to go.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/248

xzyfer:
This is fixed and will be 3.2.

anlutro:
Awesome. I love using sassc, keep up the great work :)

PR #861, opened 2015-01-29
Fix a segfault in at-root

xzyfer:
This PR fixes a segfault in `@at-root`.

Fixes https://github.com/sass/libsass/issues/859. Spec added https://github.com/sass/sass-spec/pull/248.


mgreter:
LGTM :+1:

Issue #864, opened 2015-01-30
desaturate() on grey colours saturates instead of doing nothing


anlutro:
```scss
div { color: desaturate(#999, 50%); }
```

I'd expect the result to be `#999999` but it is actually `#66cccc`.

Calling `desaturate()` with perfect white or black works as expected.

PR #865, opened 2015-01-30
Fix desaturation of greyscale colours

anlutro:
See issue #864

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -408,15 +408,19 @@ namespace Sass {
       HSL hsl_color = rgb_to_hsl(rgb_color->r(),
                                  rgb_color->g(),
                                  rgb_color->b());
-      //Check saturation is not negative before saturate it
-      double hslcolorS = hsl_color.s;
+
+      double hslcolorS = hsl_color.s + amount->value();
+
+      // Saturation cannot be below 0 or above 100
       if (hslcolorS < 0) {
         hslcolorS = 0;
       }
-
+      if (hslcolorS > 100) {
+        hslcolorS = 100;
+      }
 
       return hsla_impl(hsl_color.h,
-                       hslcolorS + amount->value(),
+                       hslcolorS,
                        hsl_color.l,
                        rgb_color->a(),
                        ctx,
@@ -431,14 +435,19 @@ namespace Sass {
       HSL hsl_color = rgb_to_hsl(rgb_color->r(),
                                  rgb_color->g(),
                                  rgb_color->b());
-      //Check saturation is not over 100 before desaturate it
-      double hslcolorS = hsl_color.s;
+
+      double hslcolorS = hsl_color.s - amount->value();
+
+      // Saturation cannot be below 0 or above 100
+      if (hslcolorS <= 0) {
+        hslcolorS = 0;
+      }
       if (hslcolorS > 100) {
         hslcolorS = 100;
       }
 
       return hsla_impl(hsl_color.h,
-                       hslcolorS - amount->value(),
+                       hslcolorS,
                        hsl_color.l,
                        rgb_color->a(),
                        ctx,


QuLogic:
It looks like the related functions also don't clamp results, e.g., `saturate` does not limit the result to 100. Maybe you could add those fixes too.

mgreter:
@anlutro Could you come up with a test case which includes all edge cases? If so please add another PR to https://github.com/sass/sass-spec!

anlutro:
@QuLogic I did that originally but thought it might be misleading because it is originally the saturation of the colour passed to the function.

As for clamping `saturate()`, it actually seems to be working fine - `saturate(red, 50%)` still returns red.

`saturate()` on grey colours behaves weirdly in my opinion, but it behaves weirdly in ruby sass so I suppose it'd be bad form to try and change it in libsass.

I'll see if I can make a PR to the spec.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/249

xzyfer:
@anlutro I've merged your specs which now causes this PR build to fail. Could you please investigate?

https://travis-ci.org/sass/libsass/jobs/48892884

anlutro:
Looks like I was wrong about `saturate()` not having the same issue. I can fix that as well I suppose.

functions.cpp
```diff
@@ -431,14 +435,23 @@ namespace Sass {
...
                                  rgb_color->b());
-      //Check saturation is not over 100 before desaturate it
+
       double hslcolorS = hsl_color.s;
+
+      // If saturation is 0, it can't be desaturated
+      if (hslcolorS <= 0) {
+        hslcolorS = 0;
+      } else {
+         hslcolorS += amount->value();
```
QuLogic: Hmm, this is `+=` now for desaturation?

functions.cpp
```diff
@@ -408,15 +408,19 @@ namespace Sass {
       HSL hsl_color = rgb_to_hsl(rgb_color->r(),
                                  rgb_color->g(),
                                  rgb_color->b());
-      //Check saturation is not negative before saturate it
-      double hslcolorS = hsl_color.s;
+
+      double hslcolorS = hsl_color.s + amount->value();
```
QuLogic: This is subtly different, but maybe that's intended. Originally for negative input `-X%` saturated by `Y%`: `-X%` -> `0%` -> `Y%`, but now `-X%` -> `(-X+Y)%` (which is likely not the same as `Y%`).

anlutro: Can input be negative? I assume `rgb_to_hsl` will never return a colour with negative saturation, and if you pass a negative number to the `saturate()` function, an error will be thrown.

QuLogic: I've no idea, but someone must have written the check for some reason! :P

functions.cpp
anlutro: Yikes, stupid mistake. Glad you spotted it.

functions.cpp
xzyfer: > $amount (Sass::Script::Value::Number) — The amount to increase the saturation by, between 0% and 100%
> [throws] if $amount is out of bounds, or either parameter is the wrong type

http://sass-lang.com/documentation/Sass/Script/Functions.html#saturate-instance_method

anlutro:
OK, there is one more problem I have to fix - `desaturate(saturate(#999, 1%), 10%);` should return `#999999` but returns `#90a2a2`.

functions.cpp
anlutro: @QuLogic you were right, though the comment and the way it was done before was a bit misleading. Basically if you desaturate a colour with 1% saturation by 10% it would wrap around and the bug would appear again.

anlutro:
Fixed and squashed commits, everything should be OK now.

xzyfer:
Great work @anlutro

PR #866, opened 2015-01-30
Pass "keep comments" flag to improve error message line numbers

anlutro:
Currently, if I have a Sass file with invalid syntax and comments, the line numbers are off because the sass2scss process doesn't preserve comments. This is fixed by changing the options bit int passed to the sass2scss function.

Hopefully I got the bit flag right, I don't work with this sort of thing on a regular basis!

Before:
```
$ cat -n ../test.sass 
     1	// some comment
     2	;
$ ./sassc/bin/sassc ../test.sass 
/home/andreas/dev/test.sass:1: invalid top-level expression
```

After:
```
$ ./sassc/bin/sassc ../test.sass 
/home/andreas/dev/test.sass:2: invalid top-level expression
```

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -298,7 +298,7 @@ namespace Sass {
     if (!source_c_str) return 0;
     queue.clear();
     if(is_indented_syntax_src) {
-      char * contents = sass2scss(source_c_str, SASS2SCSS_PRETTIFY_1);
+      char * contents = sass2scss(source_c_str, SASS2SCSS_PRETTIFY_1 | SASS2SCSS_KEEP_COMMENT);
       add_source(input_path, input_path, contents);
       return parse_file();
     }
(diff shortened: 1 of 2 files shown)


mgreter:
It actually should be a `|` instead of a `+`
```
SASS2SCSS_PRETTIFY_1 | SASS2SCSS_KEEP_COMMENT
```
Other than that I guess we would merge it in!

anlutro:
Fixed, thanks.

anlutro:
Anything blocking this?

xzyfer:
Thanks for you work!

This is @mgreter's area and he is currently busy with https://github.com/sass/libsass/pull/899

We'll make sure this makes it into the next release.

anlutro:
Ah, I see. I've been compiling all the time so releases aren't a huge deal
to me personally - I'll just keep rebasing against master until then, I was
just curious.

am11:
:+1:

PR #882, opened 2015-02-10
Add RPM spec file

GawainLynch:
Wasn't sure of the project's desired location, but will happily rebase with a different location after review.
* ~~Tested on RHEL/CentOS 6 & 7~~
* Tested on RHEL/CentOS 7
* Tested on Fedora 21

### DIFF
--- a/libsass.spec
+++ b/libsass.spec
@@ -1,0 +1,66 @@
+Name:           libsass
+Version:        3.1.0
+Release:        1%{?dist}
+Summary:        A C/C++ implementation of a Sass compiler 
+
+License:        MIT
+URL:            http://libsass.org
+Source0:        %{name}-%{version}.tar.gz
+
+BuildRequires:  gcc-c++ >= 4.7
+BuildRequires:  autoconf
+BuildRequires:  automake
+BuildRequires:  libtool
+
+
+%description
+LibSass is a C/C++ port of the Sass engine. The point is to be simple, fast, and easy to integrate.
+
+%package        devel
+Summary:        Development files for %{name}
+Requires:       %{name}%{?_isa} = %{version}-%{release}
+
+
+%description    devel
+The %{name}-devel package contains libraries and header files for
+developing applications that use %{name}.
+
+
+%prep
+%setup -q
+autoreconf --force --install
+
+
+%build
+%configure --disable-static \
+           --disable-tests \
+           --enable-shared
+
+make %{?_smp_mflags}
+
+
+%install
+%make_install
+find $RPM_BUILD_ROOT -name '*.la' -exec rm -f {} ';'
+
+
+%post -p /sbin/ldconfig
+
+%postun -p /sbin/ldconfig
+
+
+%files
+%doc Readme.md LICENSE
+%{_libdir}/*.so.*
+
+%files devel
+%doc
+%{_includedir}/*
+%{_libdir}/*.so
+%{_libdir}/pkgconfig/*.pc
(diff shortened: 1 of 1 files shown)


jrobeson:
It would be cool to have this officially in fedora/EPEL.   In the near term we could have builds on https://copr.fedoraproject.org/ though.

am11:
I think it does not compile with GCC 4.6 (or even 4.4.7, which was released after 4.6) right away, unless you apply [this kind of a patch](https://github.com/am11/libsass/commit/5138b02d8de139d3b26f7d780fe42ec728419c06) (originally provided by @mgreter) for libsass v3.1.0. I testing building node-sass with many versions of GCC (4.4.7, the manually build 4.5.2 and 4.5.4, 4.6 and 4.9) but since our requirement for node-sass was to support CentOS 5 (basically the specific glibc version that it was released with) and above, I stopped and applied the aforementioned patch to compile with 4.6. But I think version 4.7.x might be the bare minimum, as `CentOS 6.{don't-remember}` was able to build libsass as is, with its default dev-toolkit installation.

:+1: for making official RPM package of libsass! Any ETA on this?

@jrobeson, can we build on Copr today? It seems like it is active and community is engaged building packages using copr.

jrobeson:
you just need a FAS account and the copr package and you can get started pretty quickly.

mgreter:
@xzyfer I'm fine with merging this, although I didn't test it!?

xzyfer:
@GawainLynch I believe gcc 4.7 is the minimum required. Can confirm you had gcc 4.6 on the environments you tested this on?

Otherwise LGTM.

GawainLynch:
Derp! The CentOS 6 build machine I used a patch similar to the one that @am11 linked! I just retried on a vanilla install (gcc-* 4.4.7), and it indeed won't build. 

As for CentOS 7, that comes with GCC 4.8. I *think* I put GCC 4.6 as the minimum from something in the docs somewhere, and don't have a 4.6 install handy to test.

Should I just bump this to 4.7?

xzyfer:
I thought the 4.6 was the minimum but according @am11 it's 4.7. Update that
field and it's good to go.

GawainLynch:
Updated and rebased on `master`.

libsass.spec
```diff
@@ -0,0 +1,66 @@
+Name:           libsass
+Version:        3.1.0
```
xzyfer: Does this need to be 3.2.0? That's when this will land.

libsass.spec
```diff
@@ -0,0 +1,66 @@
...
+
+%files devel
+%doc
+%{_includedir}/*
+%{_libdir}/*.so
+%{_libdir}/pkgconfig/*.pc
+
+
+%changelog
+* Tue Feb 10 2015 Gawain Lynch <gawain.lynch@gmail.com> - 3.1.0-1
```
xzyfer: `3.1.0-1` what does this version number represent?

GawainLynch: With RPM packages the `-n` is the packaging number that is incremented for each build of *that specific* version.

xzyfer: Should this be 3.2.0 since that's the stable release this will land in?

xzyfer:
@mgreter can we automate these version numbers like the make file?

jrobeson:
I've never tried the CPack generated rpm files, but maybe those are worth a try? Then it could be part of the build process (once the CMake build system PR is finished)

PR #899, opened 2015-02-21
Preparation for interpolate/output refactor 

mgreter:
This is in preparation for #839

@xzyfer Maybe you could give it a review? It's already a pretty big PR and the rest of the refactoring will be much more new code. This PR should contain most refactoring I needed in the code base. Some code does not do much yet (like String_Quoted type), but it should make it easier this way to review.

I tried to make usefull commits, but I can merge them to one commit if wanted!?


ast.hpp
```diff
@@ -1326,10 +1331,10 @@ namespace Sass {
   // "flat" strings.
   ////////////////////////////////////////////////////////////////////////
   class String : public Expression {
-    ADD_PROPERTY(bool, needs_unquoting);
+    ADD_PROPERTY(bool, sass_fix_1291);
```
xzyfer: I don't know about this variable name

cssize.cpp
```diff
@@ -27,6 +27,7 @@ namespace Sass {
     new_env.link(*env);
     env = &new_env;
     Block* bb = new (ctx.mem) Block(b->pstate(), b->length(), b->is_root());
+    // bb->tabs(b->tabs());
```
xzyfer: Did you mean to comment out these?

eval.cpp
```diff
@@ -1069,15 +1073,15 @@ namespace Sass {
...
+    To_String to_string(&ctx);
     Expression::Concrete_Type ltype = lhs->concrete_type();
     Expression::Concrete_Type rtype = rhs->concrete_type();
 
     string lstr(lhs->perform(&to_string));
     string rstr(rhs->perform(&to_string));
 
-    bool l_str_quoted = ((Sass::String*)lhs) && ((Sass::String*)lhs)->needs_unquoting();
-    bool r_str_quoted = ((Sass::String*)rhs) && ((Sass::String*)rhs)->needs_unquoting();
+    bool l_str_quoted = ((Sass::String*)lhs) && ((Sass::String*)lhs)->sass_fix_1291();
```
xzyfer: Again, this method name tells me nothing about what it's doing.

expand.cpp
```diff
@@ -144,7 +146,7 @@ namespace Sass {
 
   Statement* Expand::operator()(Media_Block* m)
   {
-    To_String to_string;
+    To_String to_string; // (&ctx); -> src-map test fails!??
```
xzyfer: Should this not have `(&ctx)`?

expand.cpp
```diff
@@ -198,10 +200,11 @@ namespace Sass {
...
 
-    return new (ctx.mem) Declaration(d->pstate(),
-                                     new_p,
-                                     value,
-                                     d->is_important());
+    Declaration* decl = new (ctx.mem) Declaration(d->pstate(),
+                                                  new_p,
+                                                  value,
+                                                  d->is_important());
+    return decl;
```
xzyfer: Is there any advantage to this change? (personal curiosity)

functions.cpp
```diff
@@ -1495,6 +1485,7 @@ namespace Sass {
     BUILT_IN(image_url)
     {
       error("`image_url` has been removed from libsass because it's not part of the Sass spec", pstate);
+      return 0; // suppress warning, error will exit anyway
```
xzyfer: Thanks I've been meaning to fix this.

parser.cpp
```diff
@@ -1185,9 +1194,9 @@ namespace Sass {
     { return new (ctx.mem) Null(pstate); }
 
     if (lex< identifier >()) {
-      String_Constant* str = new (ctx.mem) String_Constant(pstate, lexed);
+      String_Constant* str = new (ctx.mem) String_Quoted(pstate, lexed);
```
xzyfer: Are you replacing all `String_Constant` with `String_Quoted`?

The semantics here are confusing since map keys don't have to be quoted.

xzyfer:
Please squash these commits before merging :)

xzyfer:
Can you provide a list of specs that now pass due to this change? You may need to rebase master.

ast.hpp
mgreter: The `needs_unquoting` flag will not be needed with the full refactoring. I my WIP this flag is needed to have the same behavior as ruby sass (https://github.com/sass/sass/issues/1291) which will be deprecated. Which means the `sass_fix_1291` can probably be removed once sass has it deprecated. But you're right, currently it doesn't make sense as `needs_unquoting` is still in the code. Please see comment below about `String_Constant` vs. `String_Quoted`. I would favor to ignore it for now.

cssize.cpp
mgreter: Yes, it probably produced wrong output in my WIP branch and unit tests don't seem to fail. I will need to check nested/compressed rendering anyway in the end. The goal is to get 100% equal output as ruby sass (which my WIP branch currently does).

expand.cpp
mgreter: It produces an error with my source-map unit tests, therefore I just marked it with a comment. Will need to check source maps after output rendering is stable.

expand.cpp
mgreter: In my WIP branch I set a flag in between. So this is just to get better diffs.

parser.cpp
mgreter: In the end `String_Quoted` will contain an unquoted value (it means the parsed string might has been a quoted string), while `String_Constant` is taken as is.

mgreter:
Thanks for the review! I'm making progress with the fallow up PR:
https://github.com/mgreter/libsass/commits/refactor/output-preparation

expand.cpp
xzyfer: :+1:

parser.cpp
xzyfer: > it means the parsed string might has been a quoted string

The semantics here are confusing to me. `String_Constant` is an unquoted string where as `String_Quoted` is either a quoted or unquoted string?

Feels like there should be a higher abstraction that given a string turns it into either a `String_Constant` or an `String_Quoted`?
Or maybe `String_Quoted` is a bad name if the string isn't necessarily quoted?

My concern is not knowing which type to use in any one circumstance.

mgreter: `String_Constant` will not be unquoted, never. This way we can preserve escape sequences (which is not the case if you unquote/quote it). To be honest it's also quite confusing to me and I don't really know if I have confuses me too. But be aware, that this PR has no real implementation for `String_Quoted`. This will be added later. Maybe this helps to clarify it a bit more: [ast.hpp] [1] and [inspect.cpp] [2]. A `String_Quoted` will be added with quotes to the output if `quote_mark` is set (has no effect on `String_Constant`).

[2]: https://github.com/mgreter/libsass/blob/59d811967e3530c2c6125a8628f1fa3cc8793968/ast.hpp#L1393
[1]:  https://github.com/mgreter/libsass/blob/59d811967e3530c2c6125a8628f1fa3cc8793968/inspect.cpp#L570

mgreter:
I finished the merge of my WIP branch. Although it currently only supports nested output and it probably is off by quite a bit. But everything is prepared to implement different output styles inside the Emitter class. Next I hope I can refactor my last part of my working WIP branch, becuase that was actually the most dirty one. But I know it is doable via Emitter class!
https://github.com/mgreter/libsass/commits/refactor/output-preparation

xzyfer:
I think there's room to reduce the number of string types in the near
future. It's getting a bit confusing. Do what you need to do and we'll come
back and evaluate the situation when you're done.

:+1:

mgreter:
Just checked wich tests pass with the output-preparation branch:
- libsass-todo-issues/issue_615
- // libsass-todo-issues/issue_639
- libsass-todo-issues/issue_694
- libsass-todo-issues/issue_870
- libsass-todo-issues/issue_884
- libsass-todo-tests/libsass/image-url
- libsass-todo-tests/libsass/mixin
- libsass-todo-tests/scss-tests/172_test_parsing_decimals_followed_by_comments_doesnt_take_forever

I guess also pass on current master! Maybe you have time to move them?

mgreter:
I have squashed the whole refactoring into two commits (the first one beeing exactly this PR) at:
https://github.com/mgreter/libsass/commits/refactor/output-preparation
Now the work needs to be done in Emitter class. Maybe you want to review it already? To have the full work finally merged I first need a correct Emitter implementation first!
I will make sure to add a third commit once it is ready!

xzyfer:
Some of these pass on master, however a few appear to be interpolation related and pass because of this PR :+1:

xzyfer:
Enabled the specs that pass on master https://github.com/sass/sass-spec/pull/263.

mgreter:
Got a first implementation for nested output rendering at:
https://github.com/mgreter/libsass/commit/3a983e1716703be7d385deef88b4cb19a638201a

That branch passes the fallowing additional tests:
- libsass-todo-issues/issue_615
- libsass-todo-issues/issue_639
- libsass-todo-issues/issue_694
- libsass-todo-issues/issue_870
- libsass-todo-tests/scss-tests/172_test_parsing_..._take_forever

It also passes all but 31 spec test exactly to ruby sass output (I only normalize multiple linefeed to one linefeed). Next is to refactor the whitespace/linefeed preserving for selectors and lists.

mgreter:
Mixed in the line-feed preserving from my WIP branch at: https://github.com/mgreter/libsass/commit/1210dfa5e3cbea93068e405f95b39feccd7346e4

With this it additionaly seems to pass one more spec test:
- libsass-todo-tests/186_test_newlines_removed_from_selectors_when_compressed

mgreter:
I have pretty much a finished implementation in another WIP branch:
https://github.com/mgreter/libsass/commits/test/emitter-implementation
Still need to clean up quite a few edges, but it already implements all 4 output modes!
I had to ignore white-space for some tests for now:
- 10 tests in compact mode
- 14 tests in compressed mode
- 9 tests in expanded and nested mode
All other tests only normalize multiple linefeeds to one linefeed.

xzyfer:
@mgreter feel free to just do what you need to do, and merge at will. I'd prefer you just merge you changes incrementally (as long as specs are passing).

mgreter:
@xzyfer :+1: Btw: you can think if `String_Quoted` as a replacement for `needs_unquote`, since I will now unquote strings when parsing. This seems to be the right thing to do and is needed for interpolation. If ok for you, I'll squash this whole PR into one commit and merge it and post other PRs accordingly!

xzyfer:
>  If ok for you, I'll squash this whole PR into one commit and merge it and post other PRs accordingly!

:+1: I'm keen to get this work over the line. Long lasting branches make me uneasy :)

mgreter:
Closing in favor of #910

Issue #892, opened 2015-02-18
Sassc 3.1 compiled with LibSass 3.0.3?


camflan:
I have used homebrew to install libsass and sassc before, but when updating to 3.1 (homebrew successfully installed both libsass 3.1 and sassc 3.1 to Cellar), sassc -v is reporting LibSass 3.0.3

```
$ sassc -v
sassc: 3.1.0
libsass: 3.0.3
sass2scss: 1.0.3
```

Because of this, I'm missing out on the awesome update to Libsass

xzyfer:
Hi @camflan pardon my naivety around brew. I think sassc binds to whichever version of Libsass you have installed. I could be that your brew was out of date when pulling Libsass.

Please uninstall both Libsass and sassc, update your brew, and reinstall. After updating brew you should get these versions from `brew info`.

```
➜  sassc git:(fbdc12b) ✗ brew info libsass
libsass: stable 3.1.0 (bottled), HEAD
https://github.com/sass/libsass
Not installed
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/libsass.rb
==> Dependencies
Build: autoconf ✔, automake ✘, libtool ✘
➜  sassc git:(fbdc12b) ✗ brew info sassc
sassc: stable 3.1.0 (bottled), HEAD
https://github.com/sass/sassc
Not installed
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sassc.rb
==> Dependencies
Build: autoconf ✔, automake ✘, libtool ✘, pkg-config ✔
Required: libsass ✘
```

***

If you problem persists please let us know.

camflan:
No problem. I deleted all traces of both sassc and libsass, upgraded homebrew and reinstalled. Still getting it bound to 3.0.3

I'll keep fiddling with it, unless you have another idea. thanks!

![screenshot 2015-02-22 09 44 30](https://cloud.githubusercontent.com/assets/27292/6319142/7f03b87a-ba77-11e4-82b0-aeb3bf4376a2.png)

mgreter:
Maybe it is linking against an old libsass system library?

camflan:
That's what I've been looking for. I've checked all of the /usr/ dirs, everything is symlinked to /usr/local/Cellar/libsass/3.1.0

Any other places it could be?

mgreter:
What does `ldd sassc` show?

camflan:
no ldd on mac, but here's what I got from otool

[...]

PR #906, opened 2015-02-23
Fix autotools versioning

mgreter:
This addresses (the already closed) issue https://github.com/sass/libsass/issues/892

- Changes autotools build to use "version" from env variable or git
- Adds third option to add a VERSION file (both build flavors)
- Fixes a bug on mingw with autotools (was not tested until now)
- Removes now obsolete code paths in configure.ac

@drewwells maybe you can give it a try with homebrew?


configure.ac
```diff
@@ -2,7 +2,10 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.61])
-AC_INIT([libsass], [3.0.3], [support@moovweb.com])
+AC_INIT([libsass], m4_esyscmd_s([ echo "$LIBSASS_VERSION" | \
+                                  git describe --abbrev=4 --dirty --always --tags 2>/dev/null | \
+                                  cat VERSION 2>/dev/null | \
+                                  echo "[na]" ]), [support@moovweb.com])
```
QuLogic: Does this actually work? I wouldn't expect `git` to copy its stdin to stdout even if it failed.

QuLogic: Also, I suggest putting this in a variable before the `AC_INIT` call, just to make it clearer.

drewwells:
LGTM homebrew now returns `Version: na` when I build it. 

homebrew will need updating to pass the version in prior to building. At least people won't think they have outdated versions now!

mgreter:
That's exactly the point. And `$VERSION` is defined by the second argument to `AC_INIT`! I tested it on linux and windows, so it seems to work. Not sure if it is even possible to change conditionals before setting AC_INIT.

drewwells:
Ah okay, thought it was looking in the environment for VERSION.

QuLogic:
`echo 1 | git describe --abbrev=4 --dirty --always --tags 2>/dev/null | cat VERSION 2>/dev/null | echo "[na]"` does not print `1`.

You can place any _shell script_ before `AC_INIT`; you (probably) can't use _autoconf_ macros.

configure.ac
mgreter: How can I use a variable inside AC_INIT (for version)?

QuLogic: Looks like you need to [use m4_define](https://hg.pidgin.im/pidgin/main/file/4432aff3b90d/configure.ac#l53).

drewwells:
can you add `VERSION` file to the PR?

mgreter:
@drewwells it's not the idea to carry this file around in the git repo. It should only be added for libsass releases (like the homebrew one). @QuLogic you're right, and `echo "" || ...` doesn't seem to do it either. Problem is that shell code tends to break system compatibility quite easy (mingw, bash, ...).

drewwells:
Okay, I'm going to write the brew formula to expect a VERSION file since it no longers has versioning. I don't see `VERSION` in 3.1.0, when does this file get added?

mgreter:
@drewwells how does the brew formula get the libsass source? I guess they are either copied in or a specific commit is referenced somewhere? The idea is that as soon as you update the sources inside your release (or the reference), you have to update the VERSION file. We cannot really increment the file for each commit and I do not want to have a VERSION file around not indicating if it is the actual tagged version or if more commits have been added! And this can IMHO easily be automated!

configure.ac
mgreter: I guess a shell script is easier and more clear at this point:
```
if [ "$LIBSASS_VERSION" = "" ]; then
  LIBSASS_VERSION=`git describe --abbrev=4 --dirty --always --tags 2>/dev/null`
fi
if [ "$LIBSASS_VERSION" = "" ]; then
  LIBSASS_VERSION=`cat VERSION 2>/dev/null`
fi
echo $LIBSASS_VERSION
```

QuLogic: Yea, I think you're right.

mgreter:
@QuLogic I added another commit that should hopefully fix the issues you correctly pointed out. I just tested it on mingw32 and travis seems to be happy too. I hope it is all correct now and it IMO looks already a lot cleaner.

mgreter:
@QuLogic incorporated you're latest remarks :+1: I also squashed all commits down to one!

QuLogic:
> IMO looks already a lot cleaner.

Yep, looking good.

drewwells:
@mgreter homebrew gets it via the url: https://github.com/sass/libsass/archive/3.1.0.tar.gz

When installing from head, the version is set to head so that is not the issue. On official stable releases, the version number needs to be provided somehow. Typically homebrew projects do that via a file with version info in it. This makes release maintenance simple and doesn't require messing with build instructions just to update a release.

mgreter:
@drewwells so why can't you just set LIBSASS_VERSION to the version encoded in that url. I mean you already have to update the string in one location. Hope you see my point ...

drewwells:
Updating line 21 is not required and there won't be any indication of a bug until someone installs sassc and runs `sassc -version`. I searched around for examples of inferring version from the formula url in the project and found no examples.

Leaving the formula as is will enter `[na]` in the version which imo is better than having it tagged as an outdated version. Homebrew itself keeps track of the version, I think that is good for now. See this output from brew info on the 3.1.0 bottle (that reports 3.0.3 from sassc).

```
brew info libsass
libsass: stable 3.1.0 (bottled), HEAD
https://github.com/sass/libsass
/usr/local/Cellar/libsass/3.1.0 (12 files, 4.1M) *
  Built from source
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/libsass.rb
==> Dependencies
```

When I find some time , I'll continue digging to see if I can expose the version programmatically from the formula definition. There has to be a way!

QuLogic:
On that ebuild, couldn't you have `DEPEND="gcc>=4.5.0"` without the `pkg_pretend`?

mgreter:
I guess this way it will also work with compatible compilers (like clang). Also `DEPEND` will probably just check if such a gcc package is installed, and not if it is the currently active compiler version.

Issue #894, opened 2015-02-20
`SASS_STYLE_COMPRESSED` fails to eliminate empty rules with comments


asottile:
Input:

```scss
a {/**/}
b {content: 'something so I have a non-empty expected output'}
```

Expected output (ruby sass + pyscss):

```
b{content:'something so I have a non-empty expected output'}
```

Output from libsass (using HEAD of https://github.com/dahlia/libsass-python (libsass 3.1.0))
```
a{}b{content:'something so I have a non-empty expected output'}
```

xzyfer:
Thanks for the report @asottile . I can confirm this is still present on master.

Could you please open  sass-spec PR? Here's a good example to get you started https://github.com/sass/sass-spec/pull/260

asottile:
@xzyfer I didn't see an obvious way to add a regression test for compressed mode, could you give me a pointer or two?

xzyfer:
@asottile sorry I was under the mistaken impression `a` would be removed in non-compressed modes as well. You're correct we don't support compressed for specs. 

There is currently work happening to better deal with output modes so hopefully this will be fixed as part of that. It's hard to say when that'll land though since these minor stylistic issues tend to be a low priority. 

***

As an aside in my opinion compressed mode is best avoided (even in ruby sass) in favour proper css minifiers like http://cssshrink.com/. The compressed mode output is rather naive and leaves a lot on the table.

asottile:
Yeah that's fine, I'm not too concerned about this (we've identified as not a blocker from moving from pyscss to libsass), just figured I'd be a good sumaritan and report :)

xzyfer:
> just figured I'd be a good sumaritan and report

Thanks <3 !

mgreter:
This should be solved by #910!

Issue #876, opened 2015-02-04
Interpolations are parsed incorrectly


jmhe:
I started working on Issue #615 recently and even submitted a pull request (https://github.com/sass/libsass/pull/854) for it. 

During the process, I realized that the root problem was something different; something definitely not addressed by my pull request.

I believe the root problem is that we are not parsing interpolations correctly.

Throughout the code, there are quite a few very similar blocks like the following:

    p = find_first_in_interval< exactly<hash_lbrace> >(i, end_of_selector);
        if (p) {
        // accumulate the preceding segment if there is one
        if (i < p) (*schema) << new (ctx.mem) String_Constant(pstate, Token(i, p, Position(0, 0)));
        // find the end of the interpolant and parse it
        const char* j = find_first_in_interval< exactly<rbrace> >(p, end_of_selector);

If I read the code correctly, the block is looking for the beginning of an interpolation, then it’s trying to seek to the end of it (seeking the closing right-brace).

But there could be many kinds of expressions between the two braces, including nested interpolations (my pull request was trying to fix that), quoted strings (with or without escaped characters), etc., etc.

The right way, I think, is that we should recursively parse whatever in-between the braces - like going back to the main `Parser::parse()` loop.

The question is: in our parser, do we have some mechanism that we can push any expressions onto a stack of some sort, then parse them recursively? 

Please advise. Thanks!

xzyfer:
I believe @mgreter has start heavily refactoring interpolation parsing and quote normalisation. Pieces of this work have already shipped in https://github.com/sass/libsass/pull/847.

jmhe:
Awesome. I will cancel my pull request and wait for the right fix.

jmhe:
@mgreter are you thinking of parsing interpolations recursively? without recursion, we might end up chasing lots of edge cases and having boilerplate code...

mgreter:
Since interpolants can be nested, this is kind of the key point. I already [merged some improvements] [1] in the prelexer and currently working on extracting more stuff from my WIP branch, which includes a lot more of these improvements, but there are also a lot of places where those need to be applied correctly.

[1]: https://github.com/sass/libsass/blob/master/prelexer.hpp#L93

mgreter:
This should be solved by #910!

Issue #870, opened 2015-02-02
Fix string-interpolation of lists, and normalization of string-quoting


lunelson:
This issue seems to have some crossover with #639 and #694. It pertains both to the normalization of string-quoting, and to the string-interpolation of lists. The test takes two lists of quoted strings, a comma-separated one and a space-separated one, each containing both double-quoted and single-quoted items, and passes them to a few interpolations, involving containing quotes and/or other characters. Corresponding spec to be PR'd forthwith.

```scss
$quoted-strings-csv: "alpha", "beta", 'gamma', 'delta';
$quoted-strings-ssv: "alpha" "beta" 'gamma' 'delta';

.csv {
  output: $quoted-strings-csv;
  output: #{$quoted-strings-csv};
  output: "[#{$quoted-strings-csv}]";
  output: "#{$quoted-strings-csv}";
  output: "["#{$quoted-strings-csv}"]";
  output: '#{$quoted-strings-csv}';
  output: "['#{$quoted-strings-csv}']";
}

.ssv {
  output: $quoted-strings-ssv;
  output: #{$quoted-strings-ssv};
  output: "[#{$quoted-strings-ssv}]";
  output: "#{$quoted-strings-ssv}";
  output: "["#{$quoted-strings-ssv}"]";
  output: '#{$quoted-strings-ssv}';
  output: "['#{$quoted-strings-ssv}']";
}
```
ruby-sass 3.4.11 output, according to SassMeister
```css
.csv {
  output: "alpha", "beta", "gamma", "delta";
  output: alpha, beta, gamma, delta;
  output: "[alpha, beta, gamma, delta]";
  output: "alpha, beta, gamma, delta";
  output: "["alpha, beta, gamma, delta"]";
  output: "alpha, beta, gamma, delta";
  output: "['alpha, beta, gamma, delta']"; }

.ssv {
  output: "alpha" "beta" "gamma" "delta";
  output: alpha beta gamma delta;
  output: "[alpha beta gamma delta]";
  output: "alpha beta gamma delta";
  output: "["alpha beta gamma delta"]";
  output: "alpha beta gamma delta";
  output: "['alpha beta gamma delta']"; }
```
output from sassc: 3.1.0-2-g2033 / libsass: 3.1.0-76-g71448:
```css
.csv {
  output: "alpha", "beta", 'gamma', 'delta';
  output: "alpha", "beta", 'gamma', 'delta';
  output: "["alpha", "beta", 'gamma', 'delta']";
[...]

Issue #838, opened 2015-01-14
Support 'compact' output style.


davidangel:
I think support for the 'compact' output is important to have. Especially since gulp-sass, and therefore nose-sass depends on libsass. The 'compact' output style is useful when trying to keep source maps correct when piping through other plugins.

Thanks.

am11:
@davidangel, while this is a great idea, but you may want to consider using the alternative approach meanwhile.

Since node.js is blessed with [postcss](https://github.com/postcss/postcss) a foundationally package, you can use packages like `csswring` to re-wire the source-maps:

```javascript
var postcss = require('postcss');
var csswring = require('csswring');
var nodeSass = require('node-sass');

var result = noseSass.renderSync({file: 'test.scss');

postcss().use(
  csswring.postcss
).process(result.css);
```

Or if you want to consume autoprefixer (yet another postcss based package which lets you configure various vendors' prefixes and autoprefix the output), it won't even require you to use csswring. See https://github.com/sass/node-sass/issues/555#issuecomment-64913888 for example and [this](https://github.com/madskristensen/WebEssentials2013/blob/master/EditorExtensions/Resources/server/services/srv-scss.js) for full-scale node-sass example, which runs two postcss processors (Autoprefixer and RTLCss) on the node-sass output, while keeping the node-sass intact.

Also, note that the most optimised and correct minification or CSS or JS is a very complicated and delicate matter. The most efficient CSS minifier IMO and in my experience is CSSO: https://github.com/css/csso (not sure whether they are postcss compatible).

Therefore, for superset CSS languages, Scss, Less, Slylus etc., IMO it does not make much sense to at least "perfect" this aspect, for sake of separation of concerns. I think it would make sense to have a unified minification and source-mapping experience from all these post-processors, rather than the varied and yet fragmented one. (but most probably it is not applicable here, since libsass follows ruby-sass for feature selection)

xzyfer:
> The most efficient CSS minifier IMO and in my experience is

There's no right answer here, but this is certainly worth checking out. YMMV. https://github.com/GoalSmashers/css-minification-benchmark

am11:
@xzyfer, interesting! I haven't tried `more-css` (!!?). Thanks for the link. :)

ericnewton76:
:+1:

The options present in ruby-sass would be greatly appreciated.  :wink:

mgreter:
This should be solved by #910!

ericnewton76:
[...]

PR #910, opened 2015-03-01
Refactor output emitter and implement missing output styles

mgreter:
This superseeds https://github.com/sass/libsass/pull/899

Related to https://github.com/sass/sass-spec/pull/270 and https://github.com/sass/sassc/pull/97

- Heavily refactored output code
- Should make source maps more reliable
- Output styles are mostly done in emitter.cpp
- Refactored Parser (added new String type)
- Interpolation should now work as with ruby sass
- Commented out obsolete code (missing spec tests?)

Fixes
- #552 libsass doesn't preserve newlines
- #615 Interpolation inside an interpolation isn't handled correctly
- #639 Interpolation mangles up lists
- #694 Interpolation of quoted strings differs from Ruby SASS
- #870 Fix string-interpolation of lists, and normalization of string-quoting
- #876 Interpolations are parsed incorrectly
- #838 Support 'compact' output style
- #894 `SASS_STYLE_COMPRESSED` fails to eliminate empty rules with comments
- libsass-todo-tests/186 and libsass-todo-tests/scss-tests/172
- partially libsass-todo-tests/css_nth_selectors


mgreter:
@xzyfer IMHO this is ready to be merged! Anything speaking against it from your side? What do you think about the changes in the spec-test? Really hope everything is still in order :sweat_smile:

xzyfer:
The only think that jumped out at me is https://github.com/sass/sass-spec/pull/270#issuecomment-76639465.  Otherwise :+1:

am11:
:+1:

am11:
AppVeyor builds are failing for VC: https://ci.appveyor.com/project/sass/libsass/build/1.0.200.

Adding `output_nested` and `output_compressed` in `win/libsass.vcxproj` and applying the following patch would fix it:

```diff
diff --git a/inspect.cpp b/inspect.cpp
index d58dd3c..f3998b1 100644
--- a/inspect.cpp
+++ b/inspect.cpp
@@ -6,6 +6,7 @@
 #include <string>
 #include <iostream>
 #include <iomanip>
+#include <stdint.h>
```

mgreter:
@am11 I've added the additional header you mentioned in your comment above, but output_nested and output_compressed have been removed actually (or I did not understand your comment)!

am11:
@mgreter, yes actually the error was about files not found (because they were referenced in vcxproj but were removed from repo). :smiley: 

It is now complaining that we need `#include<stdint.h>` in util.cpp.
https://ci.appveyor.com/project/sass/libsass/build/1.0.207/job/bbhkel9xigunkppk#L950

Would it make sense to add headers in `.hpp`s, so when the header is used by multiple `.cpp`, it shares forward declarations for those lang features?

am11:
All Green! https://ci.appveyor.com/project/sass/libsass/build/1.0.208. 
Yay! :tada:

mgreter:
Thanks for the help with the appveyor ci build! :rocket:

am11:
@mgreter, there are around [266 warnings](https://ci.appveyor.com/project/sass/libsass/build/1.0.208/job/segk9rjviip2dk7j#L1005) in VC job on AppVeyor. Some warnings are ignorable (POSIX style vs. STD-C++ style like `_stdup`).
But others may be useful. Like there are 100+ warnings regarding this error `C4099`: https://msdn.microsoft.com/en-us/library/695x5bes.aspx.

> 'Sass::Context' : type name first seen using 'struct' now seen using 'class'

(and vice versa)

For instance, in `bind.hpp` you have `struct Context`, while elsewhere it is used as class. Would it be a good idea to visit those warnings, it might improve some code patterns. :)

mgreter:
Maybe someone that actually uses MSVC could create a patch to solve these warnings? Beside a struct is principally just a class with all variables defined as public ... but wouldn't hurt to get rid of these. We also would have a copy_c_str function which could act as a replacement for strdup.

mgreter:
Since nobody has any objections against it, I'm going to hit "merge" here in a few minutes!

am11:
Sure. I will try to take a stab at those soon. :smiley:

xzyfer:
Amazing work @mgreter !!

PR #913, opened 2015-03-03
Code: Fixes VC warnings

am11:
//cc @mgreter

### DIFF
--- a/ast.hpp
+++ b/ast.hpp
@@ -352,7 +352,7 @@ namespace Sass {
     ADD_PROPERTY(bool, group_end);
   public:
     Bubble(ParserState pstate, Statement* n, Statement* g = 0, size_t t = 0)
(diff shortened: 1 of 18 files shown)


xzyfer:
Gosh VS is so picky ;)

xzyfer:
Master fails to compile with clang now. This appears to fix this.

```
fatal error: too many errors emitted, stopping now [-ferror-limit=]
```

ast.hpp
```diff
@@ -1523,7 +1523,7 @@ namespace Sass {
 
       if (with)
       {
-        if (!l || !l->length()) return str.compare("rule");
+        if (!l || !l->length() != 0) return str.compare("rule") != 0;
```
xzyfer: Shoudl be `if (!l || l->length() != 0)`

am11: Oops. TravisCI was passing, I didn't ran tests locally. :p
Thanks for fixing it! :+1:

PR #918, opened 2015-03-04
CI: Improves AppVeyor build performance.

am11:
* Adds gems directory to cache.
* Prevents gem install when available in cache.
* Adds extracted mingw to cache, to save
  extraction step.

//cc @xzyfer, @mgreter, to save few minutes per job. (-8


xzyfer:
\o/

Issue #878, opened 2015-02-05
Proposal: Add a plugin system to load 3rd party custom functions


mgreter:
IMO it would be usefull if we could load custom functions via a plugin system. The idea is that we can pre-load 3rd party custom functions (or maybe also importers) when libsass starts-up. This should make it possible to create plugins that are useable by all libsass implementors/bindings.

I don't really know yet how this should be implemented. I think they will need to be compiled to shared objects, which we then need to load dynamically into libsass. I know that other C++ libraries do it that way (i.e. GraphicsMagick). I think once this is sorted out, the rest should not be too hard to implement.

I will propably not come around anytime soon to take a closer look at this, so I just wanted to open this proposal for further discussion. Do you think there are any benefits in such a feature? Maybe someone has already some hands on experience with "plugin systems" in C/C++?

hcatlin:
:+1:

xzyfer:
:+1:  this might also interested @chriseppstein for https://github.com/sass-eyeglass/eyeglass

xzyfer:
Would this be a replacement/extension for custom functions? 

Seems like there is a lot of over lap between the two concepts, is there a clear differentiation?

mgreter:
No replacement! You can i.e. load custom function via plugins (one or more).
https://github.com/mgreter/libsass/blob/feature/plugin-functions/plugin.cpp
There are more stuff that could be usefull for plugins to do, but it's a start!

xzyfer:
Cool. This seems perfect for eyeglass. I can't wait to use it ;)

mgreter:
You should be able to use it since a few minutes if you apply #919 :wink:

xzyfer:
:+1:  I'll wait to 3.2.0. My first order of business is creating [map-merge-recursive](https://github.com/sass/sass/issues/1021) function! I'm excited!

mgreter:
Started to document it: https://github.com/mgreter/libsass/wiki/Creating-Plugins

PR #919, opened 2015-03-05
Implement third party plugin loader

mgreter:
This is a first working draft to address #878
Also added an [example plugin] [1] to this branch!

libsass-plugins-2

Build currently fails because of sassc: https://github.com/sass/sassc/pull/98

[1]: https://github.com/mgreter/libsass/blob/feature/plugin-functions/plugin.cpp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,20 +24,20 @@ else
 endif
 
 ifeq "$(LIBSASS_VERSION)" ""
-  ifneq "$(wildcard ./.git/ )" ""
-    LIBSASS_VERSION ?= $(shell git describe --abbrev=4 --dirty --always --tags)
-  endif
+	ifneq "$(wildcard ./.git/ )" ""
+		LIBSASS_VERSION ?= $(shell git describe --abbrev=4 --dirty --always --tags)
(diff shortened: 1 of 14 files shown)


mgreter:
IMHO this is ready to be merged since it compiles fine in all CI build!
I would like to mark this feature as experimental and include it in 3.2!

xzyfer:
I think you should move the example plugin out of root i.e. `/examples/plugin/basic.cpp`

Otherwise :+1:

mgreter:
Actually I intended to remove it before I remove the WIP tag ...

xzyfer:
That'll work also

mgreter:
I've removed the sample files. Will add the code to the wiki once this is merged!

xzyfer:
Awesome work @mgreter

lunelson:
This is a super exciting feature if it does what I think. Would this be a means to add advanced math functions, for example?

Example plugin link above is 404.

mgreter:
Yep, you could create a dll with various custom (math) functions to be loaded by libsass.

PR #926, opened 2015-03-07
Fix parsing and output of unknown at-rules (Fixes #855)

### LINKED ISSUE #855: Parsing of directives creating confusing error messages
I just spent way too much time hunting down an error that turned out to be me just typing `@extends` instead of `@extend`.

The error message you get when you do this is "nested selectors may not be extended", which had me on a wild goose chase.

On further inspection it turns out that `@extend %foo` and `@extend%foo` both work fine and yield the same result, so what `@extends %foo` was really parsing down to was `@extend s %foo`.

Is this part of the SASS spec? I think enforcing the whitespace here would be the sane thing to do.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7951279-parsing-of-directives-creating-confusing-error-messages?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Also contains the commits from PR #915, will rebase once that is merged!


xzyfer:
:+1:

PR #935, opened 2015-03-11
Move rpm spec file into contrib folder

mgreter:
Removed hard-coded version string because it's nearly impossible to maintain it while under source control. IMO this should only be a [reference for downstream] [1] to create their packaging for releases.
[1]: http://stackoverflow.com/questions/19082451/rpm-spec-files-in-version-control-system#comment28268846_19083011

### DIFF
--- a/contrib/libsass.spec
+++ b/contrib/libsass.spec
@@ -1,7 +1,7 @@
 Name:           libsass
-Version:        3.1.0
+Version:        %{version}
 Release:        1%{?dist}
-Summary:        A C/C++ implementation of a Sass compiler 
+Summary:        A C/C++ implementation of a Sass compiler
 
 License:        MIT
 URL:            http://libsass.org


contrib/libsass.spec
```diff
@@ -1,7 +1,7 @@
 Name:           libsass
-Version:        3.1.0
+Version:        %{version}
```
QuLogic: `%{version}` is defined by what's in `Version:`, so I'm not sure how this works out. Maybe you can define it when building, like `rpmbuild --define 'version 1.2.3'`, but I'm not sure how it will work when the two things are clashing like this.

mgreter: Actually it is meant to be replaced manually (the file should just serve as a template). I just thought it is worth a try that way, but I have no idea if this actually works or has any benefits.

PR #939, opened 2015-03-12
Bugfix for relative path creation on win

mgreter:
Do not try to create relative links if both files are
on different drives (no way to reference them relative).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Miscellaneous stuff
 
+VERSION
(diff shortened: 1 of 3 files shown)


QuLogic:
Was that second commit intended for this branch?

mgreter:
It's an example how to implement a plugin. Just had it laying around so I thought better to include it there than to have it wasted (I used it to verify the plugin implementation).

Issue #947, opened 2015-03-15
Interpolate variables in keyframe selector name


mechanicalduck:
Variables in keyframe selectors are not interpolated (it works in @keyframes, though).

Input:
````
$cur-frame: 10%;

#{$cur-frame} {
  color: red;
}
````

Ruby sass output (expected):
````
10% {
  color: red;
}
````

Libsass output (not interpolated):
````
#{$cur-frame} {
  color: red;
}
````

sassc: 3.1.0-13-g4e9a
libsass: 3.2.0-beta.1-3-gdea2
sass2scss: 1.0.3

With best regards

**Comment by xzyfer** [comment]
Thanks for the report @mechanicalduck.

I'm unable to reproduce this in 3.2.0-beta.1

```
➜  libsass git:(master) ✗ sassc -v
sassc: 3.1.0-13-g4e9a
libsass: 3.2.0-beta.1-3-gdea2
sass2scss: 1.0.3
➜  libsass git:(master) ✗ sassc test.scss
10% {
  color: red; }
```

Could you try you try a fresh compilation (`make clean && make`) and see if this issue still persists?

If so please provide the environment you're running on.

**Comment by mechanicalduck** [comment]
The interpolation won't happen when the keyframe selector is wrapped in @keyframes selector (what is necessary anyway).

````
@keyframes test {
  $cur-frame: 10%;
  #{$cur-frame} {
    color: red;
  }
}
````

**Comment by xzyfer** [comment]
You're correct, apologies. I was using the sample code supplied.

Could you please create a spec for this issue? You can use the following as a good template https://github.com/sass/sass-spec/tree/master/spec/libsass-todo-issues/issue_823

**Comment by mechanicalduck** [comment]
I am glad this is added to 3.2 milestone.

**Comment by mechanicalduck** [comment]
Added Pull Requests for the specs:
https://github.com/sass/sass-spec/pull/279

**Comment by mechanicalduck** [comment]
The needs-test tag could be removed now I guess - 
except the case of course that more extensive tests are needed.

With best regards.

**Comment by xzyfer** [comment]
For reference this is the `@keyframes` grammar

```
keyframes_rule: KEYFRAMES_SYM S+ IDENT S* '{' S* keyframes_blocks '}' S*;

keyframes_blocks: [ keyframe_selector '{' S* declaration? [ ';' S* declaration? ]* '}' S* ]* ;

[...]

PR #953, opened 2015-03-16
Implement selector contextualization for keyframe rules

mgreter:
Fixes https://github.com/sass/libsass/issues/947

### DIFF
--- a/ast.hpp
+++ b/ast.hpp
@@ -431,10 +431,10 @@ namespace Sass {
   // Keyframe-rules -- the child blocks of "@keyframes" nodes.
   ///////////////////////////////////////////////////////////////////////
   class Keyframe_Rule : public Has_Block {
-    ADD_PROPERTY(String*, rules);
+    ADD_PROPERTY(Selector*, selector);
   public:
     Keyframe_Rule(ParserState pstate, Block* b)
-    : Has_Block(pstate, b), rules_(0)
+    : Has_Block(pstate, b), selector_(0)
     { statement_type(KEYFRAMERULE); }
     ATTACH_OPERATIONS();
   };
--- a/cssize.cpp
+++ b/cssize.cpp
@@ -87,7 +87,7 @@ namespace Sass {
 
     Keyframe_Rule* rr = new (ctx.mem) Keyframe_Rule(r->pstate(),
                                                     r->block()->perform(this)->block());
-    if (r->rules()) rr->rules(r->rules());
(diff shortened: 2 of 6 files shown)


am11:
:+1: :tada:

xzyfer:
I don't think this is correct. `@keyframes` do not have selectors, they have a value/label that is a string. 

We should just resolve the interpolate value.

am11:
@xzyfer, nice catch. But the *keyframes-selector* can be `from` and `to` as well? Would it make sense to keep a whitelist to qualify for the *keyframes-selector* (in case CSS brings more terms to the mix in future)? `["{numeric_value}%", "from", "to"]`
Ref: http://www.w3schools.com/cssref/css3_pr_animation-keyframes.asp

mechanicalduck:
Specs: https://github.com/sass/sass-spec/pull/279

mgreter:
@xzyfer seems odd, since IMO the only place where this value is set is in `Expand(Ruleset*)`:
```
Statement* Expand::operator()(Ruleset* r) {
...
  Keyframe_Rule* k = new (ctx.mem) Keyframe_Rule ...
  k->selector(r->selector()->perform(contextualize->with(0, env, backtrace)))
```

So the initial value seems to be a ruleset selector? But maybe you're right. It just looked like an oversight to me, and I thought that this would actually always be a selector ...

xzyfer:
@mgreter it's because the parse is rather dumb about keyframes. It's on my todo list to improve.

@am11 I'd rather not for now because there's a lot of combinators to account for and it may change in the future.

mgreter:
@xzyfer What do you think about this PR in the meantime? Just wonder if it's better than nothing, regarding that it doesn't seem to break any existing spec tests? Or should I close it then?

xzyfer:
@mgreter I say leave for now. I'll take a pass doing naive interpolation expanding. If it works as expected I'll use that otherwise I'll ship this.

xzyfer:
I'll take a look at this over the weekend. I think it's rather straight forward.

mgreter:
@xzyfer do you think you can come up with a better solution in a reasonable timeframe for the 3.2.0 release? IMO I'd rather include this PR as is than not including a fix for the reported issue. Would really like to include this fix in the next beta! Or do you see any obvious regressions from this?

mgreter:
Merged it beside the concerns given by @xzyfer. IMO this seems to be an improvement and we can add a more precise implementation later. Just want to have this included in the next release!

PR #955, opened 2015-03-17
Refactor parser for css block comments

### LINKED ISSUE #941: CSS comments inside selectors yields invalid top-level expression
The following SCSS (taken from normalize.css) causes an "invalid top-level expression" error.

Removing the `/* 1 */` comment makes it work.

```css
button,
html input[type="button"], /* 1 */
input[type="reset"],
input[type="submit"] {
  -webkit-appearance: button; /* 2 */
  cursor: pointer; /* 3 */
}
```

I can write a sass-spec PR a bit later.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9519683-css-comments-inside-selectors-yields-invalid-top-level-expression?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
The latest output emitter refactoring removed the silent parsing of css block comments. This PR tries to ease the pain due to this change. Previously libsass was eating away some block comments that ruby sass preserved. We now don't do that anymore, but some parsing will choke if unexpected block comments occur. I created two new lexer functions (`lex_css` and `peek_css`), to mimic the old behavior (ie. skipping over block comments).

Fixes #941 and should improve handling of other similar cases, but is far from complete. Actually we should have spec-tests that do test where block comments are allowed! So if anybody wants to volunteer to create (quite) a few tests that cover all parser cases for comment; please step forward :wink: 

This refactoring also brings a few code optimizations, most notably the variadic template functions (c++11) for `alternatives` and `sequence`. I hope this doesn't raise the bar for the compilers too much; the benefits are IMO worth it (unlimited or/and prelexer clauses)!

Also added some minor optimizations to the parser where I saw potential. Mostly merged multiple `peek` calls into one `alternative` call. This way the parser only has to skip over the white-space once!

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -297,7 +297,7 @@ namespace Sass {
         0, 0
       );
       import_stack.push_back(import);
-      Parser p(Parser::from_c_str(queue[i].source, *this, ParserState(queue[i].abs_path, i)));
+      Parser p(Parser::from_c_str(queue[i].source, *this, ParserState(queue[i].abs_path, queue[i].source, i)));
       Block* ast = p.parse();
       sass_delete_import(import_stack.back());
       import_stack.pop_back();
--- a/error_handling.cpp
+++ b/error_handling.cpp
@@ -13,8 +13,6 @@ namespace Sass {
 
   void error(string msg, ParserState pstate, Backtrace* bt)
   {
-    if (!pstate.path.empty() && Prelexer::quoted_string(pstate.path.c_str()))
-      pstate.path = pstate.path.substr(1, pstate.path.size() - 1);
 
     Backtrace top(bt, pstate, "");
     msg += top.to_string();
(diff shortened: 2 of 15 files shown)


am11:
Success! VC also compiled it. C++11's variadic templates are supported in VS2013, but C++14's one are forthcoming. Here is the latest standard compliance table available (VS2013 vs. VS2015-preview): http://blogs.msdn.com/b/vcblog/archive/2014/11/17/c-11-14-17-features-in-vs-2015-preview.aspx.

mgreter:
@xzyfer IMO this is a pretty important fix, so we should probably include in 3.2? We might see more similar errors in the future, since we now try to preserve all block comments as ruby sass does, which is pretty much impossible, if any lexer or peek just skips over them (I guess we only need to preserve comments directly in `Blocks`, but can ignore all other cases (like in declarations or values).

Issue #942, opened 2015-03-13
Multiline strings with backslashes are silently truncated


anlutro:
I have helpers like this in some of my code. These used to work fine, but are now silently truncated (no errors) so that anything after the slash and newline isn't included as part of the variable.

```scss
$media-medium: "screen and (min-width: #{$break-medium-min}) \
and (max-width: #{$break-medium-max})";
```

I can provide a sass-spec PR a bit later.

anlutro:
Oddly enough, the above now works when I remove the `\` - maybe this is intentional? I actually prefer it without the `\`, but I still feel that it should be allowed - if it's not too much work to implement.

mgreter:
Libsass now much more correctly parses escaped chars. Maybe ruby sass has a special rule for lines with trailing backslashes, which I probably didn't encounter when I did the refactoring. Anyway, seems easy enough to support. But in term of CSS the backslash there does not make much sense IMO!

anlutro:
https://github.com/sass/sass/issues/1237 may be of interest regarding the backslash. It was my understanding that backslashes to denote multiline strings was valid CSS and should therefore be valid Sass/SCSS.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/278

am11:
I can reproduce this issue with current master. 
In other words, sass/sass-spec#278 is failing.

mgreter:
Unfortunately I, Travis-CI and Appveyor cannot reproduce the issue.
@am11 I'm going to close this since we have this CI tested!
Please open a new issue with your output if problem persists!

am11:
@mgreter, this only happens when git is configured to auto-correct endings. Lets say you type/paste the following in notepad:

```scss
$media-medium: "screen and (min-width: #{$break-medium-min}) \
and (max-width: #{$break-medium-max})";
```

notepad will surely add `\r\n`. :sunglasses:

mgreter:
Can you open a new issue for that with expected and actual output! Thanks!

am11:
@mgreter done https://github.com/sass/libsass/issues/1067. :)

xzyfer:
There used to be a couple specs dealing with windows line ending but I fear that may have broken since we've auto generated the specs a couple times recently.

PR #956, opened 2015-03-17
Implement css string reading function

mgreter:
Handle trailing backslash + newline token (Fixes https://github.com/sass/libsass/issues/942)

### DIFF
--- a/util.cpp
+++ b/util.cpp
@@ -79,6 +79,30 @@ namespace Sass {
     return out;
   }
 
+  // read css string (handle multiline DELIM)
+  string read_css_string(const string& str)
+  {
+    string out("");
+    bool esc = false;
+    for (auto i : str) {
+      if (i == '\\') {
+        esc = ! esc;
+      } else if (esc && i == '\r') {
+        out.resize (out.size () - 1);
+        continue;
+      } else if (esc && i == '\n') {
+        out.resize (out.size () - 1);
+        esc = false;
+        continue;
+      } else {
+        esc = false;
+      }
+      out.push_back(i);
+    }
+    if (esc) out += '\\';
+    return out;
+  }
+
   // evacuate unescaped quoted
   // leave everything else untouched
   string evacuate_quotes(const string& str)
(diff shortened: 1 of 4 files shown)


am11:
You would probably need to bump to C++11 to consume its `<string>`'s `pop_back()` implementation. (-8

mgreter:
There are alternatives that are compatible, just dumb I need to rebase that many commits :disappointed:

am11:
yay! :+1: :tada: 
PS: Wow so many commits and PRs are revolving around! Keep them coming! :sunglasses: :100:

util.cpp
```diff
@@ -79,6 +79,29 @@ namespace Sass {
...
 
+  // read css string (handle multiline DELIM)
+  string read_css_string(const string& str)
+  {
+    string out("");
+    bool esc = false;
+    for (auto i : str) {
+      if (i == '\\') {
+        esc = ! esc;
+      } else if (esc && i == '\r') {
```
xzyfer: Do we need to worry about platform difference here? Are `\r\n` line ending going to cause this to back track too many times?

mgreter: Actually it should only remove escaped newlines (a backslash before an actual linefeed). The implementation should be pretty correct and should support windows linefeeds to (but not old mac style). There is IMO only a very small "issue", that it will also actually remove `/\\\r*\n/`.

From https://github.com/sass/sass/issues/1237#issuecomment-42056762

> FYI http://www.w3.org/TR/CSS21/syndata.html#characters
>
> First, inside a string, a backslash followed by a newline is ignored (i.e., the string is deemed not to contain either the backslash or the newline). Outside a string, a backslash followed by a newline stands for itself (i.e., a DELIM followed by a newline).
>

xzyfer:
:+1:

util.cpp
am11: This is failing sass/sass-spec#278 on Windows because of `\r\n`. If I convert the endings using `dos2unix test/fixtures/spec/spec/libsass-closed-issues/issue_942/input.scss` it passes. But I suppose we don't users to use the `dos2unix` utility before compilation. :)

xzyfer: @am11 can you please open a new issue for this?

PR #957, opened 2015-03-17
Add parsing error for empty declaration value

### LINKED ISSUE #945: It does not error when a property-value is missing
Test http://sassmeister.com/gist/3e38a9c489a9627ffa9d:

```scss
.a {
  b:c;
  d:
}
//  see^ even the markdown notifies that something is wrong! (8 
```

`libsass 3.1.0` and `libsass 3.2.0-beta.1` (tested locally) output:

```css
.a {
  b: c; }
```

`ruby-sass 3.4.12` errors:

```bash
Invalid CSS after "  d:": expected expression (e.g. 1px, bold), was "}"
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9557278-it-does-not-error-when-a-property-value-is-missing?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Fixes #945 - Adds a function to output errors in the same way as ruby sass does, which IMO is the final goal. IMO it doesn't matter too much if error messages are not consistent in libsass for the time beeing, so we can gradually change all errors to give the same output (IMO this is easier once we actually can test errors with spec tests).

### DIFF
--- a/debugger.hpp
+++ b/debugger.hpp
@@ -298,7 +298,7 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
     cerr << ind << "Map " << expression << " [Hashed]" << endl;
   } else if (dynamic_cast<List*>(node)) {
     List* expression = dynamic_cast<List*>(node);
-    cerr << ind << "List " << expression <<
+    cerr << ind << "List " << expression << " (" << expression->length() << ") " <<
       (expression->separator() == Sass::List::Separator::COMMA ? "Comma " : "Space ") <<
       " [delayed: " << expression->is_delayed() << "] " <<
       " [interpolant: " << expression->is_interpolant() << "] " <<
--- a/parser.cpp
+++ b/parser.cpp
@@ -2187,4 +2187,39 @@ namespace Sass {
     throw Sass_Error(Sass_Error::syntax, ParserState(path, pos.line ? pos : before_token, Offset(0, 0)), msg);
   }
 
+  // print a css parsing error with actual context information from parsed source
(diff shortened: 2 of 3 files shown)


am11:
We would probably need to prevent sassc from prompting error on `stderr`, but redirect to `stdout` (`2>&1`)?

mgreter:
Holy moly, AppVeyor returned before Travis-CI :astonished: @am11 IMO it would be more usefull to split the stderr into its own file (`2>output.stderr`). Basically we expect the same error for all four output styles, so if the test runner sees such a stderr file, it only has to do one test (at least that's how I probably would implement it today).

am11:
Now Sass org is using AppVeyor FOSS Pro account (since last night), so we can experiment with it with much faster pace. :wink: :sunglasses: :sunny: 
There was no way to wait for hours to get the job done!

mgreter:
@xzyfer This could either be 3.2 or 3.3 IMO, but since it fixes one bug, we may want to include it in 3.2 anyway. I could also remove the new error output and only move that to 3.3 (I added some more text to the first post in this PR). IMO we should at least address the error itself!

xzyfer:
I think this is fine to ship in 3.2. it'll give us a chance to fix any gotchya in 3.3. :ship:

PR #958, opened 2015-03-17
Add arrow indicator to error message (sass_context)

### LINKED ISSUE #938: Provide a formatted error output
Moved from https://github.com/sass/node-sass/issues/687

Would be useful for consumer to just be able to log out an already nicely formatted error string. Currently I have to [manually create](https://github.com/sindresorhus/grunt-sass/blob/5b0c8a231574f043e16c3d2c2e7c6b84bc1796c3/tasks/sass.js#L36) the error output in my grunt plugin.

I'm thinking something like this (but also with filename):

image

*(from [`gulp-jscs`](https://github.com/jscs-dev/gulp-jscs))*

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9470428-provide-a-formatted-error-output?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Fixes #938 (also includes other PRs, so wait till these are merged #955!)

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -297,7 +297,7 @@ namespace Sass {
         0, 0
       );
       import_stack.push_back(import);
-      Parser p(Parser::from_c_str(queue[i].source, *this, ParserState(queue[i].abs_path, i)));
+      Parser p(Parser::from_c_str(queue[i].source, *this, ParserState(queue[i].abs_path, queue[i].source, i)));
       Block* ast = p.parse();
       sass_delete_import(import_stack.back());
       import_stack.pop_back();
--- a/error_handling.cpp
+++ b/error_handling.cpp
@@ -13,8 +13,6 @@ namespace Sass {
 
   void error(string msg, ParserState pstate, Backtrace* bt)
   {
-    if (!pstate.path.empty() && Prelexer::quoted_string(pstate.path.c_str()))
-      pstate.path = pstate.path.substr(1, pstate.path.size() - 1);
 
     Backtrace top(bt, pstate, "");
     msg += top.to_string();
--- a/error_handling.hpp
+++ b/error_handling.hpp
@@ -14,7 +14,6 @@ namespace Sass {
     enum Type { read, write, syntax, evaluation };
 
     Type type;
-    string path;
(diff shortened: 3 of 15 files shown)


xzyfer:
Can you document the changes to the lexer? `peek` vs `sneak` vs `peek_css` vs `lex_css`.

If it's just about handling the "consume all whitespace/comments" behaviour then I would opt to keep `peek` and `lex`, and instead remove this frustrating behaviour.

mgreter:
@xzyfer from #955 

> We might see more similar errors in the future, since we now try to preserve all block comments as ruby sass does, which is pretty much impossible, if any lexer or peek just skips over them (I guess we only need to preserve comments directly in Blocks, but can ignore all other cases (like in declarations or values).

So `lex_css` is basically the same as `lex` was before my major refactoring on output and interpolation. New `lex` function is the same as it was after my refactoring. This introduces regressions like we are seeing in #938. So `lex_css` basically does a `lex < block_comment >` before it tries to parse the actual token. This also has the benefit, that source-maps will get more accurate. We should never use something like `lex < optional_block_comment, token >`, because the source-map position will point to the position of the `block_comment`. IMO we had a few such cases. Other than that I just tried to clean up those function, but they should behave the same!

xzyfer:
Yeah I agree this is important. Could you add comments to the function signatures describing when to one over the other? 

It's something I know I'm likely to forget, and it'll certainly trip up new contributors.

mgreter:
@xzyfer Due to your request I tried to be even more specific, but I already added comments for all new relevant functions. Maybe I didn't understand your concerns, but it was pretty well annotated already!

xzyfer:
I think it's more clear now :+1:  nice work!

mgreter:
So you're ok to include in 3.2?

xzyfer:
Yep :ship:

PR #961, opened 2015-03-17
[WIP] Improve static value parsing

### LINKED ISSUE #593: Loss of spaces in interpolated selectors
The following:
```scss
h1:nth-of-type(#{2 + 'n + 1'}) {
    color: red;
}
```
is transformed by libsass to:
```css
h1:nth-of-type(2n+1) {
  color: red; }
```
As you can see, spaces around the `+` are lost. Ruby Sass (3.4.5) preserves them.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5641727-loss-of-spaces-in-interpolated-selectors?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
This should fix #593, but needs more testing. I guess this topic would be great to have more generated tests to really cover all differences between ruby sass and libsass. And static value parsing is IMO not very well covered yet, since there are so many little details.

### DIFF
--- a/prelexer.cpp
+++ b/prelexer.cpp
@@ -430,6 +430,9 @@ namespace Sass {
     const char* number(const char* src) {
       return sequence< optional<sign>, unsigned_number>(src);
     }
+    const char* sass_number(const char* src) {
+      return sequence< optional<sign>, unsigned_number, optional < identifier > >(src);
+    }
     const char* coefficient(const char* src) {
       return alternatives< sequence< optional<sign>, digits >,
                            sign >(src);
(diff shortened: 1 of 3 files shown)


prelexer.cpp
```diff
@@ -423,6 +423,9 @@ namespace Sass {
     const char* number(const char* src) {
       return sequence< optional<sign>, unsigned_number>(src);
     }
+    const char* sass_number(const char* src) {
+      return sequence< optional<sign>, unsigned_number, optional < identifier > >(src);
```
xzyfer: What's the use case for `optional < identifier >` here? Are all identifiers really valid in the context?

xzyfer: Also could this reduced to? 

```
return sequence< number, optional < identifier > >(src);
```

mgreter: My tests seems to indicate that ruby sass allows pretty funny units for numbers, so yes, it should IMO be quite correct. But I think it might be usefull to explicitly define a prelexer for number units. Basically its the same as number + units. This fixes following case:

```
1px    /     2px
```

xzyfer: Interesting. It'd be worth looking at the Ruby parse to see what their rule is but for now :+1:

mgreter: Yes, that would be great. Let me know if you find something usefull in their source!

xzyfer: I wont get a chance to look at this anytime soon. If this fixes an issue :+1: I'll open another issue to take a look at static value parsing.

xzyfer: https://github.com/sass/libsass/issues/974

PR #964, opened 2015-03-18
SourceMap: Fixes relative file path issue

am11:
This fixes the issue described here: sass/node-sass#764.

> When `output_path` is set as absolute path, the path should be resolved with respect to map file path for emitter.

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -111,7 +111,7 @@ namespace Sass {
       }
     }
 
-    emitter.set_filename(output_path);
+    emitter.set_filename(resolve_relative_path(output_path, source_map_file, cwd));
 
   }
 


am11:
With this change, node-sass passes all 749 tests. (and without it, nine source-map related tests were failing)

@xzyfer, would it be possible to release `libsass-v3.0.0-beta.2`, so we can release node-sass `v3.0.0-alpha.1` with all the awesome libsass new feature? (-8

xzyfer:
@am11 https://github.com/sass/libsass/releases/tag/3.2.0-beta.2

am11:
@xzyfer, thank you! :+1: :tada:

am11:
Now, I would just need to attempt the code to compile against gcc 4.4.7 on *o0ooold* CentOS! 8-)

PR #968, opened 2015-03-18
[WIP] Created an Xcode project to build libsass as a static-libary

onedayitwillmake:
This is a simple pull request to be able to build / work on the project from Xcode on OSX taking advantage of the debugger to step through the code.

It is built as a static-library, which is placed in $PROJECT/lib (as in the Makefile), with a simple attached project that contains a simple `main` function to allow stepping through the code when debugging.


xzyfer:
No objections from me /cc @mgreter

mgreter:
No idea what this is actually about, since I'm not really a fan of fruits that others already took a bite off :smile: Since it doesn't touch any other files, I'm ok with it, just wonder if it would be better suited under the new `contrib` folder, since we probably cannot support this "officially"!

mgreter:
By the way, by adding the plugin API it probably got even harder to compile libsass as a pure static library! I had to remove the `-static` flag to get it to compile on CI, since the `libdl` dependency was throwing errors when compiled static (since it is expected to load other libraries on runtime dynamically for plugins). It might be possible, but my gcc/make foo was not good enough to keep the real static compilation in the Makefiles!

onedayitwillmake:
Ah I see, this is mostly meant to help with debugging while on OSX. 

@mgreter Yes, I am not officially representing my employer in this pull request. I can move it into the contrib directory if it makes more sense there.

mgreter:
@onedayitwillmake Maybe I could also "force" you to create a small [wiki page] [1] to describe what you expect people to do with these files :wink: 

https://github.com/sass/libsass/wiki/External_Contributions

xzyfer:
I've tried getting Xcode to play nice with Libsass to no success. I see this as no different the Visual Studio files we have. If this works I'll be over the moon! :+1: on the wiki page also

onedayitwillmake:
Sure thing, creating the wiki page now. Will be a little thin, but I can fill it up per your suggestions after

mgreter:
Currently every line of documention is probably a 1% increase in documentation :+1:

onedayitwillmake:
Here's a quick write up, let me know if there are any extra blanks to fill 
https://github.com/sass/libsass/wiki/External_Contributions

mgreter:
@onedayitwillmake Thanks a :100: ! As I said, I would rather see this under the new contrib directory, so maybe you could ammend your commit to put these filese there (//CC @xzyfer)!? Not sure, but could this be tested by a CI service?

xzyfer:
I believe Travis no has beta OSX support we may be able to use.

mgreter:
I saw the update on the UI which indicates linux builds, but I didn't find anything about opening mac support (maybe we would be eligable anyway, if we applied for mac os builds).

xzyfer:
It exists but you have to directly apply and they _may_ let you use it. The node-sass team tried but their OSX builds don't support node (IIRC). It might be easier for us since we're C++ but even so I'm not sure we can test Xcode directly.

TBH if this works I'll probably switch to using Xcode so I'll know if it breaks.

mgreter:
@xzyfer so you're willing to "maintain" these build files as I do the mingw stuff? If so, I'm also ok to include it in the main directory (we actually should move the source/header files to different directories, as the makefiles are ok there). And maybe someone could just ask travis-ci about that (and no I didn't just volunteer, I'm away the next 5 days). I'm just asking, but let @onedayitwillmake have a chance to make it right, so we can merge it :stuck_out_tongue_winking_eye:

am11:
:+1: 
This is also one step towards our ultimate CMake goal /cc @nschonni. If this works, we will have all the criteria we need to create universal CMake configuration: https://github.com/sass/libsass/pull/292.

onedayitwillmake:
I just pushed another commit, which removed an extra scheme I had created in the workspace.

@mgreter I'm not familiar with Travis-CI, but in order to build from command line it's possible to run:
```
$ xcodebuild -workspace xcode/xcode.xcworkspace/ -scheme static-library
```

I can be in charge of maintaining the Xcode build files. 
It is not much work, I simply added only the CPP files listed in the Makefile to the static-library project target, so I have no problem keeping it in line. 

(to be sure, I have no problem placing the /xcode/ directory in 'contrib' either, up to you guys)

mgreter:
@onedayitwillmake you can create a space on travis-ci for your own libsass fork, where you can play with the config file (`.travis.yml`). I use that workflow when I work on feature branches!

onedayitwillmake:
@mgreter Sounds good, I'll make sure it's building in Travis-CI in my next commit.

onedayitwillmake:
@mgreter I notice the `.travis.yml` file is calling `./script/ci-build-libsass` which builds `libsass` into the 'lib' directory. 

What would be the best way for me to get the xcode build working, without stepping on the above script's work since they are building the same libsass.a file?

mgreter:
@onedayitwillmake Each build should actually just build one library, so not sure why you seem to compile two libraries at once. IMO you will need to add a new compiler type, like we currently have for gcc and clang, and you may have to use `env` vars to do it. It could help if you post the url to some travis build, since I'm not familiar with Xcode at all.

onedayitwillmake:
@mgreter I'm not familiar with Travis, I was thinking when it looks at `travis.yml` it looks for information on how to build the library, but the project already contains that information and builds using GCC/Clang.

I'm not sure where I would add steps to build using xcode, without in effect bumping out the current information, or building the library a second time in order to keep both instructions. Again this could be because i'm not familiar with Travis CI

mgreter:
travis.xml just sets some env variables and calls others scripts (as you already found out), so you need to put the xcode stuff somehow into those scripts. It may need some refactoring to do this properly!

onedayitwillmake:
Sorry, this might sound like a broken record but I'm not sure what the task is.

`travis.yml` currently builds using GCC/Clang command line, but we would like to have it /instead/ build using xcodebuild command line tool /if/ it's running on an OSX environment in travis? 

Otherwise, building via command line GCC/Clang will then be erased once the xcodebuild is done. Correct?

mgreter:
As I said before, it would help if you could show us your work in progress ...

Travis currently starts 4 processes, each starting with a clean state. Currently we have no mac support on travis, so once this would be added (is it even evailable for all on travis?), we will probably have about 4 more processes on travis. In the build scripts you actually find:

```
if [ "x$AUTOTOOLS" == "xyes" ]; then
```

So you need to test for your variable or arch, or whatever fits your needs here. Maybe it will be more efficient to add another build script for use specific use case and add a switch to travis.xml which scripts gets called. But since I have no idea what you actually need to do either, I'm just guessing!

onedayitwillmake:
Ah yes ok this clears everything up, it will just run it in parallel! Got it 👍

onedayitwillmake:
So Travis gives the variable, `TRAVIS_OS_NAME` which I check for `osx`, to trigger the build. 

However the build even with:  
```
os:
  - linux
  - osx
```

It will still trigger a linux only build.
Apparently mult-os projects are not exactly supported currently, seems to be due to server load (http://docs.travis-ci.com/user/multi-os/](http://docs.travis-ci.com/user/multi-os/)

It also appears that by default, all builds run on ` Ubuntu 12.04 LTS Server Edition 64 bit` **unless** the language is set to `objective-c` 

http://docs.travis-ci.com/user/ci-environment/#CI-environment-OS
> Travis CI virtual machines are based on Ubuntu 12.04 LTS Server Edition 64 bit, with the exception of Objective-C builds, which runs on Mac OS X Mavericks.

I tried a few the obvious thing of setting two languages in the YAML just to kick start it, but it didn't fall for that 😢 (https://travis-ci.org/onedayitwillmake/libsass/builds/55532426)

Just to see if it was a possibility, I set the language to 'objective-c' which did indeed trigger the OSX host, which launched xcodebuild and everything went according to plan 😁(https://travis-ci.org/onedayitwillmake/libsass/builds/55552324 - although libsass.so version currently fails because im not sure how to place the library where it wants after building the dynamic library) 

**TL;DR** Travis wont allow multiple OS environments, you can only force the OSX environment if falsely set the language to 'objective-c' - With this flag set it builds correctly in OSX and passes the tests

am11:
@onedayitwillmake, perhaps you can get some pointers from Ruby repos' .travis.yml:

https://github.com/ruby/ruby/blob/trunk/.travis.yml

onedayitwillmake:
Well specifically they mention that, currently if you want multiple OS support you /can/ request it - perhaps that is the route they took?

> This feature needs to be enabled manually. If you wish to take advantage of this feature, please send email to support@travis-ci.com, indicating which repository you would like to have it enabled

> We are at capacity with Mac workers in both .org and .com environments. Consequently, we are not accepting requests for the multi-OS feature at this time. As more capacity is added, we will resume accepting requests.

onedayitwillmake:
I did give those parameters a try (simply `osx` as well as `linux) in the `.travis.yml`, however it does not trigger an OSX build.

This was the resulting output: 
https://travis-ci.org/onedayitwillmake/libsass/builds/55528078

In fact, setting the host OS has no effect, you will only get an OSX VM if the language is set to objective-c.

http://docs.travis-ci.com/user/ci-environment/#CI-environment-OS
> Travis CI virtual machines are based on Ubuntu 12.04 LTS Server Edition 64 bit, with the exception of Objective-C builds, which runs on Mac OS X Mavericks.

am11:
I once sent them email for osx support for node-sass, and they denied cutely:

>I took a look at this repository. I am afraid we are unable to turn the multi-OS feature for you at this time, as the repository requires multiple Node.js runtimes that are not available.
>
 For currently supported language configurations, please consult https://github.com/travis-ci/travis-ci/issues/2320.
>
 We are working to beef up our Mac infrastructure, and planning to enable this feature for all repositories.
>
Thank you for your understanding.

Perhaps libsass gets the pass.

xzyfer:
I've requested OSX support for the sass/libsass repo.

xzyfer:
We now have osx support enabled. In theory it should "just work".

xzyfer:
@onedayitwillmake if you revert your latest travis changes and just add the osx os it should work now with the objective c hack.

onedayitwillmake:
Yay! 🎊 - will do in the morning!

onedayitwillmake:
In order to iron out any kinks, should I request OSX support for my own fork? Or was that part of their green light?

onedayitwillmake:
@am11 @xzyfer This /should/ do the trick now.

I've set the language:
```
language: 
  - objective-c
  - cpp
``

Which should trigger the OSX environment. 
I've also temporarily excluded the shared build on OSX, it compiles correctly but `sassc` cannot locate it.
I would consider that a separate issue for now

**Comment by onedayitwillmake** [comment]
@xzyfer Looks like a new different error is showing up:

```
Using /home/travis/.rvm/gems/ruby-1.9.3-p551
0.01s/home/travis/build.sh: line 42: /usr/local/bin/actool: Permission denied
```

I wonder if it is a combination of linux environment, and the language being set to `objective-c` in order to trigger the OSX dual environment

**Comment by onedayitwillmake** [comment]
I tried swapping the language setting, but this forced the linux only environment causing OSX builds to fail of course.

Perhaps someone can jump in with some advice, but it appears.

* Setting language 'objective-c' causes linux builds to fail (even though they are in fact ran from a linux worker), potentially a bug in Travis.
* Setting language to 'cpp' prevents OSX host from being created causing Xcode builds to fail.

**Comment by onedayitwillmake** [comment]
@mgreter Any advice, it appears that for now we cannot satisfy Travis (cannot trigger OSX unless lang is set as objective-c, not setting the language to cpp causes build failure for linux)?

**Comment by xzyfer** [comment]
I've emailed the Travis team to confirm that building against multiple OSes
is actually possible. Will update when I get a reply.

**Comment by onedayitwillmake** [comment]
@xzyfer It's definitely picking it up, it's just that we have to set the language to Objective-C which is causing the other builds to fail for some reason

**Comment by xzyfer** [comment]
@onedayitwillmake yeah I get that. However ideally we want to be able to run both Ubuntu and OSXCI builds. 

Correct me if I'm wrong, but triggering OSX builds is current an all or nothing deal right?

**Comment by onedayitwillmake** [comment]
@xzyfer I don't believe so, as on the top it's stating that it's using a linux worker:
`Using worker: worker-linux-3da96de7-1.bb.travis-ci.org:travis-linux-6`

**Comment by mgreter** [comment]
It's pretty confusing indeed, I can see that job 1379.5 has an "Apple" icon, while the others have a "tux", but is still reported as "worker-linux-3da96de7-1.bb.travis-ci.org:travis-linux-6" !?

**Comment by onedayitwillmake** [comment]
The environment is not correct, but brew is installed by default. If we run brew it will not exist.

The only way to trigger OSX environment, is to set the language to `objective-c`, other than being obviously wrong. It causes the reverse, and linux builds fail. Lemme do that now, maybe we can dissect it.

When the environment is OSX brew exist, and is one of the steps in my modified ci script

**Comment by mgreter** [comment]
My tests indicate that it works as intended: https://travis-ci.org/sass/libsass/builds/56759363

Here the `travis.yml` it is using:
```bash
language: cpp

os:
  - linux
  - osx

compiler:
  - gcc
  - clang

env:
  - AUTOTOOLS=no COVERAGE=yes BUILD=static
  - AUTOTOOLS=yes COVERAGE=no BUILD=shared

matrix:
  exclude:
    - compiler: clang
      env: AUTOTOOLS=yes COVERAGE=yes BUILD=static

script: uname
```

onedayitwillmake:
@mgreter Yes, I see your test - looks very promising, maybe I was confusing myself with travis because i'm not familiar with it. *My fork*'s does not allow dual OS builds, but on these merge request builds this project does. I reset the `.travis.yml` file to match more like the one above

onedayitwillmake:
Strange, it looks like it just timed out - but the OSX build was never started

onedayitwillmake:
@mgreter It looks like it's working, but with a few caveats:

From the output, https://travis-ci.org/sass/libsass/jobs/56763059, and I noticed this locally as well, it builds using xcode then something else down the line rebuilds the library.

Curiously this happens output:
```
successfully compiled libsass
AUTOTOOLS=no COVERAGE=yes BUILD=static
``` 

Which means it must happen in a subsequent script.

mgreter:
@onedayitwillmake as you may have seen the compilation via Makefiles and Autotools works as expected on OS-X without a hitch (#1024). So I wonder what the benefit of this xcode build is? Some of the files you added seem to contain md4 hashes of the source files? Or are they static (like the GUIDS in MSVC build)? My main concern here is that we have to maintain these files after all if we include it in our CI process, and I'm currently not sure what the benefit of it would be, since we have Travis-CI and OS-X now working with clang and the existing build scripts.

onedayitwillmake:
@mgreter The files added are part of the Xcode's self contained project file, much as how imagine a .sln file is under the hood. They are just GUID references to the files.

I think if you are using a different OS, then you're right there is no benefit of including the Xcode project from a compilation standpoint for the computer. It builds on OSX using make just fine.

The benefit is more for users on OSX, to be able to contribute to the project using Xcode along with the LLDB debugger by simply pulling the repo down and opening the project file.

However, to your point if that is not a strong enough reason because it would have to be updated when new files are added/removed the way that VS solution file has to be, then I can close the pull request.

mgreter:
Does xcode use clang under the hood? If so, I really don't see any benefits in including it in the CI process. But we may still include them in the contrib directory or you could create a separate repository on github to include them (you may even be able to add your own CI process there). I just don't understand why the source files are mentioned multiple times, seems silly to me, but I guess xcode is not really for console lovers :wink: @xzyfer what's your opinion here?

onedayitwillmake:
@mgreter Fully agree, perhaps in contrib directory is best.

Also out of curiosity which is the spot you're referring to, with the multiple times? The project file? I want to make sure I did not think you were talking about something else when I answered

mgreter:
Ah I see, they are all in comments but one. So that clears this up. Now the only thing that is that they are actually referenced multiple times. Well for MSVC we have to add them in two places/files, so it seems doable, but still quite a bit more difficult, as it seems. I'm not sure what the best place would be to put these files in. They should IMO all be in a separate folder and we should probably add a readme there indicating that we do not really keep those files in sync etc.

xzyfer:
I'm all for this, largely for this reason

>The benefit is more for users on OSX, to be able to contribute to the project using xcode along with the LLDB debugger by simply pulling the repo down and opening the project file.

I know I have tried to get this working myself. My xcode-fu is terrible. We essentially do this already for MSVC. As for where the files go I don't any opinions as long as it _just works_. Ideally xcode is treated as a first class citizen like MSVC. 

If this works I would use is immediately.

mgreter:
@xzyfer If you are using the files yourself I see it much more positive, since I hope in that case they will not get stale very soon. I guess you personally don't get any benefit if we add it to the CI process? So I think it should go either into `./xcode/` or `./contrib/xcode/`. The first one would be more similar to how we have MSVC builds in the tree. And for their alphabetical neighbourhood, I'm actually in favor of that. IMO the tree structure is a mess already, so no harm done :smile:

xzyfer:
:-1: to `./xcode/` for consistency

onedayitwillmake:
@mgreter @xzyfer I think im going to re-do this as a new pull request, to skip over the ugly stuff in the middle that I did while testing out options

PR #971, opened 2015-03-19
Make @error fatal.  Resolves #967.

### LINKED ISSUE #967: @error is non fatal (should be fatal, is inconsistent with custom functions)
I'm running against current master: c15b384d4ca047b414551312ae6cbc834b48e7db

The following:

```c++
#include "sass_context.h"
#include <cstring>
#include <iostream>


int main() {
    // ??? in 3.1.0 I could just call sass_make_data_context("a literal")
    // now I need to put it on the heap? and `free`ing it causes double free?
    // (unrelated but made it harder to make this bugreport)
    char* src = (char*)malloc(sizeof(char) * 50);
    strcpy(src, "a { @error 'foo'; }");
    struct Sass_Data_Context* context = sass_make_data_context(src);
    sass_compile_data_context(context);
    struct Sass_Context* ctx = sass_data_context_get_context(context);
    std::cout << "Error status:" << std::endl;
    std::cout << sass_context_get_error_status(ctx) << std::endl;
    sass_delete_data_context(context);

    return 0;
}
```

Produces:

```
$ ./main 
Error: foo
	on line 0 of stdin

Error status:
0
```

However this code (using custom functions):

```c++
#include "sass_context.h"
#include <cstring>
#include <iostream>


static union Sass_Value* custom_func(const union Sass_Value* args, void* cookie) {
    return sass_make_error("hello world");
}


int main() {
    char* src = (char*)malloc(sizeof(char) * 50);
    strcpy(src, "a { content: foo(); }");
    struct Sass_Data_Context* context = sass_make_data_context(src);
    struct Sass_Options* options = sass_data_context_get_options(context);
    Sass_C_Function_List fn_list = sass_make_function_list(1);
    Sass_C_Function_Callback fn = sass_make_function("foo()", custom_func, NULL);
    sass_function_set_list_entry(fn_list, 0, fn);
    sass_option_set_c_functions(options, fn_list);
    sass_compile_data_context(context);
    struct Sass_Context* ctx = sass_data_context_get_context(context);
    std::cout << "Error status:" << std::endl;
    std::cout << sass_context_get_error_status(ctx) << std::endl;
    sass_delete_data_context(context);

    return 0;
}
```

Produces:

```
$ ./main_custom_func 
Error status:
1
```

The same is reproduced for `@warn` as well (fatal in custom functions, nonfatal outside).

Prior art:
- ruby sass produces a failure for `@error` nodes and does not fail on `@warning` nodes

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9728652-error-is-non-fatal-should-be-fatal-is-inconsistent-with-custom-functions?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asottile:
Manual testing as requested!
```
$ ./bin/sassc test.scss 
Error: hai
        on line 1 of test.scss
>> a { @error 'hai'; }
   ----^
$ echo $?
1
```

### DIFF
--- a/eval.cpp
+++ b/eval.cpp
@@ -265,10 +265,7 @@ namespace Sass {
     }
 
     string result(unquote(message->perform(&to_string)));
-    Backtrace top(backtrace, e->pstate(), "");
-    cerr << "Error: " << result;
-    cerr << top.to_string(true);
-    cerr << endl << endl;
+    error(result, e->pstate());
     return 0;
   }
 


xzyfer:
Am I correct in remembering that `error` give us the backtrace trace that you've removed?

asottile:
ah yeah I've made an unused variable.  I'll remove that.  You can see the sample output in the description of the PR

asottile:
fixed!

xzyfer:
Yeah cool just wanted to make sure backtraces were maintained. It's hard to tell from the example since there's not nesting to unwind.

I have confirmed this is the case - https://github.com/sass/libsass/blob/master/error_handling.cpp#L14-L21

Great work! Thanks!

PR #973, opened 2015-03-23
Use autoconf to detect -dl

saper:
Do not hardcode -ldl, -lstdc++ and linker flags.
Simplify automake/autoconf configuration

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,30 +1,16 @@
 ACLOCAL_AMFLAGS = -I m4
 
-AM_LDFLAGS = -lstdc++ -lm
-AM_CFLAGS = -Wall
-AM_CXXFLAGS = -Wall
-
-if COMPILER_IS_MINGW32
-  AM_CXXFLAGS += -std=gnu++0x
-else
-  AM_CFLAGS += -fPIC
-  AM_CXXFLAGS += -fPIC
-  AM_CXXFLAGS += -std=c++0x
-  AM_LDFLAGS += -ldl
-endif
-
-AM_CFLAGS += -DLIBSASS_VERSION="\"$(VERSION)\""
-AM_CXXFLAGS += -DLIBSASS_VERSION="\"$(VERSION)\""
+AM_COPT = -Wall -O2
+AM_COVLDFLAGS =
 
 if ENABLE_COVERAGE
-	AM_CFLAGS += -O0 --coverage
-	AM_CXXFLAGS += -O0 --coverage
-	AM_LDFLAGS += -O0 --coverage -lgcov
-else
-	AM_CFLAGS += -O2
-	AM_CXXFLAGS += -O2
-	AM_LDFLAGS += -O2
+	AM_COPT = -O0 --coverage
+	AM_COVLDFLAGS = -lgcov
 endif
+
+AM_CFLAGS =   $(AM_COPT)
+AM_CXXFLAGS = $(AM_COPT) -std=c++0x
+AM_LDFLAGS =  $(AM_COPT) $(AM_COVLDFLAGS)
 
 EXTRA_DIST = \
 	COPYING \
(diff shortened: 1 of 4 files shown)


xzyfer:
This appears to be failing CI

QuLogic:
You know, I'm not even sure what makes `dlopen` necessary...

QuLogic:
> You know, I'm not even sure what makes dlopen necessary...

Ah, my `master` checkout was outdated. `plugins.hpp` uses `dlopen`, so it's a necessary thing. Probably just add the `-dlopen` flag to `libsass_la_LDFLAGS`?

saper:
Yes, it does not yet work as expected. We probably should be able to get `-fPIC` flags from `libtool` as well.

mgreter:
LGTM and CI seems to agree! @saper maybe you could give me some background info how you acquired your autotools wizardry :wink: Would make me even more confident to merge this in before the final 3.2 release! I also would love the have some feedback from @QuLogic, since he was basically the only guy that seemed to have an actual clue about that kind of stuff so far. I just went pretty much by best guesses plus a lot of try and error when I refactored the autotools build :smile:

saper:
This is magic. I decided to overcome my angst w/autotools and I can recommend @daivd-a-wheeler video as a first step (http://www.dwheeler.com/autotools/).

Frankly I am learning as we go.

There are still few things to improve (axe few more items in Makefile.am).

I was wondering if we would like to include some proper `gcov.m4` macro - there is one which is GPL - [http://bazaar.launchpad.net/~allanlesage/coverage-tutorial/trunk/files](coverage-tutorial on LaunchPad).

Next step (not in this pull request) will be to detect some C++ features and get rid of gcc 4.4.* branch!

Makefile.am
```diff
@@ -87,9 +75,7 @@ libsass_la_SOURCES = \
 	utf8_string.cpp utf8_string.hpp \
 	util.cpp util.hpp
 
-libsass_la_CFLAGS = $(AM_CFLAGS)
-libsass_la_CXXFLAGS = $(AM_CXXFLAGS)
-libsass_la_LDFLAGS = $(AM_LDFLAGS) -no-undefined -version-info 0:9:0
+libsass_la_LDFLAGS = -no-undefined -version-info 0:9:0
```
QuLogic: According to [this](http://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html):

> Automake always uses two of these variables when compiling C sources files. When compiling an object file for the mumble target, the first variable will be mumble_CPPFLAGS if it is defined, or AM_CPPFLAGS otherwise. The second variable is always CPPFLAGS.

So, you should still need `$(AM_LDFLAGS)` here.

saper: Correct. I need to see what we can do with test coverage.

Unfortunately I think we might want to pass `-O2` to the linker too.

QuLogic: > -O level
> >    If level is a numeric values greater than zero ld optimizes the output. This might take significantly longer and therefore probably should only be enabled for the final binary. At the moment this option only affects ELF shared library generation. Future releases of the linker may make more use of this option. Also currently there is no difference in the linker's behaviour for different non-zero values of this option. Again this may change with future releases.

So yes, it appears to make a difference; not sure how much.

mgreter:
Unfortunately this breaks my build on windows with msys and autotools.
Will check now if I can find the culprit and see if there is a workaround.

mgreter:
I have cloned and [ammended] [1] your branch so it compiles again on mingw/windows. @saper Maybe you want to give it a review? Then I could either create a new PR or you update your branch with mine.

[1]: https://github.com/mgreter/libsass/commit/5793876228305426242c7080f4b7f6b3c872e919

saper:
I am already half-done with `autoconf` stuff. I have now macros to detect range loops, nullptr as well as to clean up <random> madness.

Work in progress in in http://github.com/saper/libsass/tree/cxx11

Now working on variadic templates ...

saper:
@mgreter your changes are of course fine (minus one little nitpick)

saper:
@mgreter Can you help me in getting prelexer work on old compilers (I try to use the old code if C++11 does not fully work - http://github.com/saper/libsass/tree/variadic0). I seem to be getting error in `parser.cpp`:

https://gist.github.com/saper/69435ad853b3f7a1fc10

I see that previously `peek` was used differently, but why this fails only in those two places? Any ideas how to fix it without too much `#ifdef`ing ?

PR #966, opened 2015-03-18
Implement using & in SassScript

xzyfer:
This PR implements using & in SassScript.

Fixes #548. Specs added https://github.com/sass/sass-spec/pull/103, https://github.com/sass/sass-spec/pull/280.

Successor to #965, #959

/cc @ekskimn

PR #975, opened 2015-03-23
Handle only the required nodes in Listize

xzyfer:
This PR does some clean up of the recent `&` in SassScript feature https://github.com/sass/libsass/pull/966.

/cc @ekskimn

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -316,7 +316,7 @@ namespace Sass {
       register_c_function(*this, &tge, c_functions[i]);
     }
     Contextualize contextualize(*this, &tge, &backtrace);
-    Listize listize(*this, &tge, &backtrace);
+    Listize listize(*this);
     Eval eval(*this, &contextualize, &listize, &tge, &backtrace);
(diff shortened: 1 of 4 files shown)


onedayitwillmake:
Slightly tangential question, what does listize do? When is it required and when is it not?

xzyfer:
Listize turns selectors into Sass lists. It's needed when selectors are being used as Sass values i.e. variables, function/mixin arguments etc.

ekskimn:
Hi @xzyfer , I took at look at the PR and it makes sense to me. =) :thumbsup:

mgreter:
LGTM :+1: ! Maybe we want to include this in the next beta? IMO you can still take it from there!

mgreter:
Merged it beside the concerns given by @xzyfer. IMO this seems to be an improvement and we can add a more precise implementation later. Just want to have this included in the next release!

xzyfer:
:+1: I have more work to do here but this is a good start

Issue #976, opened 2015-03-23
Parser is correctly some built-in `@directives`


xzyfer:
Original report https://github.com/sass/sassc/issues/103.

```scss
.debug {
  @debug-this {
    foo: bar;
  }
}
```

Results in the following error as the parser assumes `@debug-this` is in fact `@debug -this`.

```
Error: non-terminal statement or declaration must end with ';'
        on line 9 of test.scss
>>   @debug-this {
   ---------^
```

mgreter:
I IMO have a valid [fix] [1] for this in https://github.com/mgreter/libsass/tree/bugfix/issue_976

[1]: https://github.com/mgreter/libsass/commit/84b1ab0907953177e4e2ca24a34d37a6dab3b846

mgreter:
This is fixed by https://github.com/sass/libsass/pull/990, hope you don't mind @xzyfer :wink:

PR #977, opened 2015-03-23
Parse at-directives to the word boundary

xzyfer:
This PR fixes the parsing of at-directives making sure we take into account the whole directive.

Fixes https://github.com/sass/libsass/issues/976. Needs spec.

### Todo

- [ ]  fix whitespace in output

/cc @mgreter

### DIFF
--- a/parser.hpp
+++ b/parser.hpp
@@ -174,6 +174,14 @@ namespace Sass {
       return peek< mx >(peek < css_comments >(start));
     }
 
+    // at-directives that aren't just prefixes
+    template <prelexer mx>
+    const char* peek_directive(const char* start = 0)
+    {
+      // now peek a token (skip comments first)
+      return peek< sequence< mx, negate< identifier > > >(start);
+    }
+
 #ifdef __clang__
 
 #pragma clang diagnostic pop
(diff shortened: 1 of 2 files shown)


parser.hpp
```diff
@@ -174,6 +174,14 @@ namespace Sass {
       return peek< mx >(peek < css_comments >(start));
     }
 
+    // at-directives that aren't just prefixes
+    template <prelexer mx>
+    const char* peek_directive(const char* start = 0)
+    {
+      // now peek a token (skip comments first)
+      return peek< sequence< mx, negate< identifier > > >(start);
```
mgreter: Maybe it would be more correct to have a [word-boundary] [1] match (like regex has with `\b`)?
Other than that I fully agree that this should be a step in the right direction!

[1]: http://www.regular-expressions.info/wordboundaries.html

mgreter: I actually just saw that we already have the word boundary check implemented and in use for `extend` keywords. I just re-factored it a little bit and added a `word` prelexer to match a full word by a given kwd constant. Spec tests all pass and it should be more robust this way.
-> https://github.com/mgreter/libsass/commit/2a05583cde482c42bd78edf148cb4a53703e8d48

mgreter:
IMO this is superseeded by https://github.com/sass/libsass/pull/990. That PR implements word boundary assertions for all keywords. I also refactored the prelexer quite a bit and was able to improve the performance quite a bit. I guess your pretty busy right now, but I hope we can get this included in 3.2!

mgreter:
@xzyfer I hope you don't mind, but this should be fixed by #990 !

Issue #980, opened 2015-03-23
Unable to append variable-length parameters after optional parameter


davidkpiano:
A function that is called with an optional parameter followed by a variable-length parameter throws the error `"Optional parameters may not be combined with variable-length parameters"` despite being valid in Ruby Sass.

```scss
@function foo($value, $default: 13, $args...) {
  $res: $value + $default;
  @if length($args) != 0 {
    $res: $res + nth($args, 1);
  }
  @return $res;
}

.test {
  value: foo(3); // expected: 16
  value: foo(3, 4); // expected: 7
  value: foo(3, 4, 5, 6); // expected: 12
}

// RUBY SASS
.test {
  value: 16;
  value: 7;
  value: 12;
}

// LIBSASS
// Throws "optional parameters may not be combined with variable-length parameters"
```

mgreter:
I guess this has changed in ruby sass without us noticing. I just had to remove the existing check and everything seems to work correctly (#981). I took the liberty to adjust the test case from @davidkpiano above and I hope someone would volunteer to create a PR on sass-specs for this, thanks!

xzyfer:
We need to be careful here. Optional and variable-length parameters can be mixed, but only in specific orders i.e. variable-parameters can follow optional parameters but not the other way around.

ncoden:
> variable-parameters can follow optional parameters but not the other way around.

@xzyfer Why not ? I understand that when passed inline we cannot split optional parameters from variable-parameters, but it would work well with named parameters, and would be really useful to provide additional "second-hand" options to a function.

For example:
```scss
@function please-call-this-for-me($func, $args..., $warn: true) {
    @if type-of($func) == 'function' and function-exists($func) {
        @return call($func, $args...);
    }
    @else if $warn == true {
        @warn 'Function #{$func} does not exists'   
    }
}

$_: please-call-this-for-me(foo, bar, baz);

$_: @function please-call-this-for-me(foo, bar, baz, $warn: false);
```

That's a stupid example I know, but we can see that the "main job" of `please-call-this-for-me` is to call a function with arguments (this is why using variable-parameters seems legit to me, instead of a list) and that we need beside this "main job" to customize it with optional arguments that will not be used often but that we still need.

PR #981, opened 2015-03-23
Remove obsolete check for rest args after optionals

mgreter:
Fixes https://github.com/sass/libsass/issues/980

### DIFF
--- a/ast.hpp
+++ b/ast.hpp
@@ -1667,9 +1667,6 @@ namespace Sass {
         if (has_rest_parameter_) {
           error("functions and mixins cannot have more than one variable-length parameter", p->pstate());
         }
-        if (has_optional_parameters_) {
-          error("optional parameters may not be combined with variable-length parameters", p->pstate());
-        }
         has_rest_parameter_ = true;
       }
       else {


ast.hpp
```diff
@@ -1667,9 +1667,6 @@ namespace Sass {
         if (has_rest_parameter_) {
           error("functions and mixins cannot have more than one variable-length parameter", p->pstate());
         }
-        if (has_optional_parameters_) {
```
xzyfer: Is this correct? I believe this is still a valid error.\

The error case mentioned in the issue is 

> [when] optional parameter followed by a variable-length parameter

This change appears to affect the case 

> [when] a variable-length parameter is followed an optional parameter

Which is a legal scenario.

If this fixes the reported issues I believe the bug is due to something before this check.

mgreter: IMO it should be spot on now. The test I removed is obviously wrong, since rest parameters can come after a parameter with optional (default) value (as you can see `has_optional_parameters_` is set when we had a preceding `default_value`. Or can you come up with a sample that doesn't behave correctly?

Issue #982, opened 2015-03-24
C-API missing sass_context_take_included_files()


rodneyrehm:
Context data can (usually) be accessed in two ways `sass_context_get_output_string()` and `sass_context_take_output_string()` - the latter detaching the data from libsass' memory cleaning. This the case for pretty much any string hosted on the context, except for the list of files, where we have `sass_context_get_included_files()` but no `sass_context_take_included_files()`.

Is this an oversight or a deliberate omission?

mgreter:
Pretty much an oversight. Maybe you're willing to create a PR in this regard yourself? Should be pretty easy to add this function. I will probably not come around to do it before the weekend!

PR #983, opened 2015-03-24
feature(sass_context): adding sass_context_take_included_files()

rodneyrehm:
fixing #982

### DIFF
--- a/sass_context.cpp
+++ b/sass_context.cpp
@@ -795,6 +795,7 @@ extern "C" {
   IMPLEMENT_SASS_CONTEXT_TAKER(char*, error_file);
   IMPLEMENT_SASS_CONTEXT_TAKER(char*, output_string);
   IMPLEMENT_SASS_CONTEXT_TAKER(char*, source_map_string);
+  IMPLEMENT_SASS_CONTEXT_TAKER(char**, included_files);
 
   // Push function for include paths (no manipulation support for now)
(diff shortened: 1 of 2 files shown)


mgreter:
@rodneyrehm I guess you already have tested it? I guess it should be ok, but I'm a bit worried because it's a `char**` in this case ...

rodneyrehm:
Yes, I've successfully tested in in my context - but I haven't written any tests for it (as coveralls already reported).

I don't understand why you're worried, it is a `char**` in the first place and considering all the function does is null the pointer so it doesn't get cleaned later. I didn't see a difference to how `char *` is handled. That said, I have **no** idea what I'm doing - I'm rather happy I'm able to spell "C" ;)

mgreter:
I come to the same conclusion as you with `char**`, it's just that it is slightly different than all the others and it often tends to bite me in the ass if I just blindly trust my logic :wink: Going to merge this since you confirmed that it at least works in your context. Impacts are pretty slim with this function anyway!

Thanks for the contribution @rodneyrehm !
BTW. On my todo list also to update sass.link.js :smile:

rodneyrehm:
Thanks for merging!

> BTW. On my todo list also to update sass.link.js

I'm currently looking at [API Sass Importer](https://github.com/sass/libsass/wiki/API-Sass-Importer) and how I can integrate that into sass.js - that should go a long way to simplify your sass.link.js, no? I'll open an issue for that wiki page, as I have questions to ask / discoveries and opinions to share.

Issue #613, opened 2014-11-03
Support Sass 3.3 scoping


xzyfer:
Following on from https://github.com/sass/libsass/issues/278 opening this issue to track work on the new scoping model.

Although the Libsass changeset should be easy enough, sass-spec will probably need some significant work.

mgreter:
As a side note: A proper implementation should release the memory of the variable as soon as it goes out of scope (I guess we don't have to deal with references and counting). AFAIK my last tests showed that ruby sass also does not do this, and libsass of course neither yet! This could IMO lead to potential DOS, since a simple very big `@for` loop will allocate at least one additional variable on each iteration. I thought about a memory limit for [ctx.mem] allocations, but for that we would need to have stats about all allocated memory sizes.

PR #985, opened 2015-03-25
Performance and Memory improvement for loops

mgreter:
Added some performance improvements for some loops. Currently libsass allocates a new variable in a `for` loop for each iteration. This is not the best strategy, as it costs us performance and memory (since all these allocations would not be freed until context is freed). I have added two different optimizations and we should try to find more spots which could optimize the same way.

- Added Memory Manger to Environment, so stuff that really only lives in a certain Env can be re-claimed when the Environment goes out of scope (mostly variables).
- Changed `for` loop to re-use the same variable as iterator (maybe we should make sure that we reset the unit in case the user has changed it, since we only change the actual value ... can we even iterate over pixels?).

Made some benchmarks with a very big for loop:

```CSS
@for $i from 1 through 20000000 {}
// from: 10 wallclock secs ( 9.30 usr +  0.67 sys =  9.97 CPU) and ~ 3GB RAM
// to  :  0 wallclock secs ( 0.05 usr +  0.00 sys =  0.05 CPU) and no ram inc
```

### DIFF
--- a/environment.hpp
+++ b/environment.hpp
@@ -5,7 +5,9 @@
 #include <string>
 #include <iostream>
 
+#include "ast_fwd_decl.hpp"
 #include "ast_def_macros.hpp"
+#include "memory_manager.hpp"
 
 namespace Sass {
   using std::string;
@@ -20,6 +22,7 @@ namespace Sass {
     ADD_PROPERTY(Environment*, parent);
 
   public:
+    Memory_Manager<AST_Node> mem;
     Environment() : current_frame_(map<string, T>()), parent_(0) { }
 
     map<string, T>& current_frame() { return current_frame_; }
--- a/eval.cpp
+++ b/eval.cpp
@@ -100,7 +100,9 @@ namespace Sass {
     double start = static_cast<Number*>(low)->value();
     double end = static_cast<Number*>(high)->value();
     Env new_env;
-    new_env[variable] = new (ctx.mem) Number(low->pstate(), start);
+    // only create iterator once in this environment
+    Number* it = new (new_env.mem) Number(low->pstate(), start);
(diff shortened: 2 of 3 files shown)


akhleung:
Ooh, pretty cool! It'll be interesting to see the impact on large scss projects.

am11:
:+1: 
Perhaps this is related: I have observed some significant performance decrease when running the [huge test](https://github.com/sass/sass-spec/tree/master/spec/scss/huge) with v3.2.0-beta.2 (compared to v3.1.0) via node-sass test runner.

drewwells:
Seeing ~0.8% speedup on project compile time vs 3.2.0-beta.2

mgreter:
@drewwells I'm actually quite surprised that you see such an improvement. My benchmark test was really not a real-work case and I was more concerned with the memory usage there (DOS). I would also like to implement memory limits for allocations as it should be trivial inside memory manager.

drewwells:
Yeah I thought 0.8% was really impressive. The performance graph does show
a lot of time is spent compiling and freeing the compiler context vs OS
tasks like file I/O.

mgreter:
Btw. related issue https://github.com/sass/libsass/issues/613

mgreter:
I actually have another optimization pending in https://github.com/mgreter/libsass/tree/refactor/prelexer. With that branch, my spec runner reports a pretty nice improvement :smile: So stay tuned!

```
in 14 wallclock secs (14.24 usr +  0.47 sys = 14.71 CPU)
in 12 wallclock secs (10.94 usr +  0.58 sys = 11.51 CPU)
```

drewwells:
rock on :100: !

PR #986, opened 2015-03-26
Implement `!global` assignments

mgreter:
Implemented correct `!global` and `!default` handling. Was actually not that hard once I "understood" how it all ought to work, but `@for` and `@each` threw me off a bit, since they don't seem to be "real" blocks (they don't seem to have their own env scope). Added a few additional spec tests which should cover some of the more advanced use cases (which haven't been covered yet). https://github.com/sass/sass-spec/pull/282

Oh and btw. this fixes https://github.com/sass/libsass/issues/613

### DIFF
--- a/debugger.hpp
+++ b/debugger.hpp
@@ -69,7 +69,7 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
         case Complex_Selector::ADJACENT_TO: cerr << "{+}"; break;
         case Complex_Selector::ANCESTOR_OF: cerr << "{ }"; break;
       }
-    cerr << " <" << prettyprint(selector->pstate().token.ws_before()) << "> X <" << prettyprint(selector->pstate().token.ws_after()) << ">" << endl;
+    cerr << " <" << prettyprint(selector->pstate().token.ws_before()) << ">" << endl;
     debug_ast(selector->head(), ind + " ", env);
     debug_ast(selector->tail(), ind + "-", env);
   } else if (dynamic_cast<Compound_Selector*>(node)) {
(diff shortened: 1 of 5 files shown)


mgreter:
@xzyfer maybe you want to review? IMO it's ready to be merged!

xzyfer:
Sorry I've been offline running a conference. I'll catch up today.

functions.cpp
```diff
@@ -145,7 +145,9 @@ namespace Sass {
     static mt19937 rand(static_cast<unsigned int>(GetSeed()));
 
     // features
-    static set<string> features;
+    static set<string> features {
+      "global-variable-shadowing"
```
xzyfer: We should remember to mention this in the changelog.

debugger.hpp
```diff
@@ -69,7 +69,7 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
         case Complex_Selector::ADJACENT_TO: cerr << "{+}"; break;
```
xzyfer: Cool. I'd fixed this in a branch that I haven't PR'd yet :)

xzyfer:
LGTM, nice one

debugger.hpp
mgreter: Looks like your using `debug_ast` too :wink: I hope you like it as much as I do :smile: It actually decreased my time hunting bugs dramatically! I also thought about creating a nicer `debug_env` function, but I was able to use existing one (altough the output could be better there).

xzyfer: I saw you post debub_ast output in an issue and fell in love.

PR #992, opened 2015-03-27
Simplify autotools build

mgreter:
- Generate sass_version.h
- Use $(EXEEXT) for executables
- No need for -lm and C++ library.
- They should be provided by libtool.

Work done by @saper !

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ Makefile.in
 /m4/ltsugar.m4
 /m4/ltversion.m4
 /m4/lt~obsolete.m4
+/script/ar-lib
 /script/compile
 /script/config.guess
 /script/config.sub
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,29 +1,21 @@
 ACLOCAL_AMFLAGS = -I m4
 
-AM_LDFLAGS = -lstdc++ -lm
-AM_CFLAGS = -Wall
-AM_CXXFLAGS = -Wall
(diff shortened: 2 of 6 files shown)


saper:
Can we also have `AM_PROG_AR` on mingw32? Otherwise autoreconf complains out:

````
/usr/local/share/automake-1.15/am/ltlibrary.am: warning: 'libsass.la': linking libtool libraries using a non-POSIX
/usr/local/share/automake-1.15/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
Makefile.am:28:   while processing Libtool library 'libsass.la'
```

saper:
Looks like [_AC_PROG_CC_G](http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/c.m4;h=4c87d2121ad03b06e19115dfd6b6eb7684bbbd06;hb=HEAD#l504) autoconf test sets `-O2 -g` as the default compiler flags on gcc-likes (including clang).

drewwells:
just as a sanity check, b/c this has been changing a lot lately. Here's the params I pass to build libsass

    -lsass -lsdc++ -lm -ldl -std=c++11

lm is removed by this PR and I think ldl was removed in another. Which of these params should I be passing now?

saper:
Can you try running `autoreconf` and then `./configure` ? libtool should find them

Issue #452, opened 2014-07-26
Support for Shadow DOM styling [$5]


trevordmiller:
When trying to use the `/deep/` Shadow DOM selectors for Web Components, the libsass compiler throws an error:

```bash
Warning: /Users/trevordmiller/Sites/github/frontendtv.github.io/app/styles/main.scss:41: error: invalid top-level expression
```

Can support be added for Shadow DOM / Web Components / Polymer related selectors like `/deep/`, `::shadow` and ':host` etc.?

http://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-201/

peterblazejewicz:
:+1:

hackingbeauty:
Can this please get fixed and merged?  I'm not able to use libsass in my Polymer project.

peterblazejewicz:
ref: https://sass-compatibility.github.io/#shadow_dom_styling

Panya:
BTW `/deep/` was changed to `>>>` (http://dev.w3.org/csswg/css-scoping-1/#deep-combinator)

JosefJezek:
https://code.google.com/p/chromium/issues/detail?id=446051

xzyfer:
This is more involved than than the current implementation. I've started work on this but got distracted - see https://github.com/xzyfer/libsass/pull/1

>For intents and purposes the /deep/ combinator acts like the direct sibling combinator except for the shadow dom. [3.2.4 Selecting Through Shadows: the /deep/ combinator](http://dev.w3.org/csswg/css-scoping/#deep-combinator).

Essentially `/deep/` needs to be treated exactly the same as `+`. 

This is a relevant specs - https://github.com/sass/sass-spec/blob/master/spec/libsass-todo-tests/scss-tests/175_test_reference_combinator_with_parent_ref/input.scss

****

Also `/deep/` is a [reference combinator](http://www.w3.org/TR/selectors4/#idref-combinators) as defined in CSS Selectors Level 4.

> The reference combinator consists of two slashes with an intervening CSS qualified name, and separates two compound selectors, e.g. A /attr/ B.

We need to handle generic reference combinators, not just `/deep/`.

This is a relevant spec - https://github.com/sass/sass-spec/blob/master/spec/libsass-todo-tests/scss-tests/117_test_selector_interpolation_in_reference_combinator/input.scss

mgreter:
@xzyfer I'm confused. The spec test seems to be passing!? Can this be closed again?

[...]

PR #994, opened 2015-03-28
Implement `/deep/` shadow DOM selector

mgreter:
Fixes https://github.com/sass/libsass/issues/452 (spec-tests: https://github.com/sass/sass-spec/pull/287)

### DIFF
--- a/constants.cpp
+++ b/constants.cpp
@@ -78,6 +78,9 @@ namespace Sass {
     extern const char webkit_calc_kwd[]  = "-webkit-calc(";
     extern const char ms_calc_kwd[]      = "-ms-calc(";
 
+    // css selector keywords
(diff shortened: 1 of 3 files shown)


xzyfer:
@mgreter I looked into this a while ago, it's much more involved than what has been done here - see https://github.com/xzyfer/libsass/pull/1

>For intents and purposes the /deep/ combinator acts like the direct sibling combinator except for the shadow dom. [3.2.4 Selecting Through Shadows: the /deep/ combinator](http://dev.w3.org/csswg/css-scoping/#deep-combinator).

This is a relevant spec - https://github.com/sass/sass-spec/blob/master/spec/libsass-todo-tests/scss-tests/175_test_reference_combinator_with_parent_ref/input.scss

PR #995, opened 2015-03-28
autoconf macros to improve C++11 compat

saper:
Introduce autoconf macros to detect
TR1 and C++11 features.

Don't hardcode -std=c++0x for all compilers.
Detect various variants of `<unordered_map>`
and provide workaround if `at()` is missing.

unordered_map related code should be now compatible
with gcc 4.2 (has no C++11 features)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,8 @@ AC_TYPE_SIZE_T
 AC_FUNC_MALLOC
 AC_CHECK_FUNCS([floor getcwd strtol])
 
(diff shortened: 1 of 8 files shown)


xzyfer:
@saper is this still relevant? If so please rebase off master.

saper:
We need to resolve C++ namespacing first, and then this can possibly be resolved quickly.

glebm:
We now use `-std=c++11` throughout and require C++ 11 support (at least the subset supported by VS2013, for now).

PR #996, opened 2015-03-28
Work around C++11 nullptr

saper:
Provide CXX11_NULLPTR autoconf macro

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,8 @@ AC_TYPE_SIZE_T
 AC_FUNC_MALLOC
 AC_CHECK_FUNCS([floor getcwd strtol])
 
+CXX11_NULLPTR
+
 # Checks for testing.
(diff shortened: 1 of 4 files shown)


xzyfer:
@saper is this still relevant? If so please rebase off master.

xzyfer:
Ping ^

saper:
Rebased for @drewwells for experiments with limited compiler environment (Android, MacOS X 10.6 SDK, etc.)

mgreter:
IMO it's easier to just replace `nullptr` with `0`, I see no real need for it, we use `0` all over the place.

xzyfer:
I think agree with @mgreter. We don't have the issues `nullptr` is intended to fix.

saper:
> I think agree with @mgreter. We don't have the issues `nullptr` is intended to fix.

Let's use at least NULL instead :)

xzyfer:
Tracking in https://github.com/sass/libsass/issues/1657

Issue #962, opened 2015-03-17
Add importer registry to have multiple custom importers


mgreter:
Currently we only support one custom importer, that has the job to "convert" import urls to import-data (basically loading the content). To support multiple importers, each importer should have a function to determine if it wants and can handle a given import url. This way we could have multiple importers loaded, one for http traffic, another for ftp etc.

### Idea

- each importer implements a `canHandle` function (plus optional priority)
- libsass asks each importer in priority order if it `canHandle` the import url
- call importer that `canHandle` the url and repeat until first "valid" result

### Some open questions

- There is a chance that multiple importers can handle the same type of url (or one handles more specific urls than the other, like loading from a subdirectory). For this we probably need some kind of priority system and also an option to try other importers if current one gave an error. We could also accumulate all imports that get returned by multiple importers, but I don't really see a use case for that.

- We could implement this on top of the currently existing API or change the existing API to support this kind of feature. Since it is still marked experimental, I'm propably for the second option.

### Expose more stuff for importers (#984)

- Expose function to resolve filenames with libsass (by using include paths)
- Expose options of the active context (maybe pass `ctx` and use option API?)

This is open for discussion and more ideas!

am11:
@xzyfer also mentioned this for node-sass on Gitter chat. 

In JavaScript, the end-user can either use promises (such as async#waterfall: https://www.npmjs.com/package/async#waterfall, example usage: [Bartinger/grunt-github-release-asset/github_release_asset.js#L42](https://github.com/Bartinger/grunt-github-release-asset/blob/e44ab430b62b622fbbba228d79453e592e039c86/tasks/github_release_asset.js#L42)) or just a if/else, switch/case state machine to call appropriate function (to read data from database, web-service, file-systems etc.). For this, IMO, there is no need for API support and offloading the implementation design choice to final implementer would make sense: one generic entry point would suffice.

I may be missing the point. Please correct me if there is a use-case which cannot be handled without multiple importers.

mgreter:
[...]

Issue #960, opened 2015-03-17
Custom Mixins


drewwells:
libsass has the ability to specify custom functions, but not custom mixins (that I'm aware of). Are there plans to enable specifying custom mixins?

I'm hacking custom mixins by either injecting mixins at the beginning of source string or using the custom importer interface. The second one is fine, but does require the developer to include special `@import` to enable the custom mixin.

Here's an example mixin that gets injected into the source:

```
@mixin sprite-dimensions($map, $name) {
  $file: sprite-file($map, $name);
  height: image-height($file);
  width: image-width($file);
}
```

mgreter:
This basically boiled down to have custom headers. I implemented something experimental in #1000 to support custom headers. I will paste the relevant parts from the linked issue:

Custom headers (#960) act like importers, but only run once at the beginning of each compilation. They are primarily meant to define custom mixins (via plugins or context options). Beside that they might be usefull for other cool things I currently cannot think of :smile: Opposite to custom importers, all custom headers will be executed in priority order and all imports will be accumulated (so many custom headers can add various custom mixins or css-code).

drewwells:
That is AWESOME! Exactly what I was looking for!

mgreter:
Merged #1000 which contains a first experimental implementation!

PR #1000, opened 2015-03-28
Implement support for multiple importers

mgreter:
This PR now implements all options I envisioned in #962. At this point I'm not sure if we want to include this in 3.2.0 or wait for 3.2.1, since it does bring a few breaking changes with custom importers and custom functions (not that many, but still noticeable). @am11 what do you think, do you have time to adjust this in node-sass before 3.0? It basically just involves setting a list of importers instead of a single one (plus passing an additional priority value on creation). Other than that everything can stay as it is. 

### How do custom importers work now?

For each `@import` libsass will call importers in priority order (highest priority first). If the importer returns `0`, libsass will skip to the next one. If no importer returns anything, libsass will continue to load the import on its own. Once an importer returns something (either a list of imports, an empty list of imports or an error), the chain is aborted and only the given return is handled. Additionally I fixed the behaviour when only a filename is returned to be exactly as with plain libsass imports.

### Breaking C-API changes

To create custom importer list:
```C
double priority = 0; // default
void* cookie = (void*) pointer_to_anything;
Sass_Importer_List importers = sass_make_importer_list(1);
importers[0] = sass_make_importer(sass_importer, priority, cookie);
```

New function signatures for custom functions and importers:
```C
Sass_Import_List sass_importer(const char* cur_path, Sass_Importer_Entry cb, struct Sass_Compiler* comp)
union Sass_Value* call_sass_function(const union Sass_Value* s_args, Sass_Function_Entry cb, struct Sass_Options* opts)
```

The options that have previously been passed can now be fetched indirectly:
```C
void* cookie = sass_importer_get_cookie(cb);
struct Sass_Import* previous = sass_compiler_get_last_import(comp);
const char* prev_path = sass_import_get_path(previous);
void* cookie = sass_function_get_cookie(cb);
```

You should now be able to access all context options via API calls.
```C
struct Sass_Context* ctx = sass_compiler_get_context(comp);
struct Sass_Options* opts = sass_context_get_options(ctx);
double precision = sass_option_get_precision(opts);
```

Renamed a lot of typedefs regardings function and importer API. They should now make much more sense. Not sure how big of an impact this has on implementers though, but it should make it a lot more future-proof. Signatures for functions and importers should not need to change again, since they can now access all options available indirectly via the passed arguments.
```C
// Typedef helpers for import lists
typedef struct Sass_Import (*Sass_Import_Entry);
typedef struct Sass_Import* (*Sass_Import_List);
// Typedef helpers for custom importer lists
typedef struct Sass_Importer (*Sass_Importer_Entry);
typedef struct Sass_Importer* (*Sass_Importer_List);
// Typedef defining importer signature and return type
typedef Sass_Import_List (*Sass_Importer_Fn)
  (const char* url, Sass_Importer_Entry cb, struct Sass_Compiler* compiler);

// Typedef helpers for custom functions lists
typedef struct Sass_Function (*Sass_Function_Entry);
typedef struct Sass_Function* (*Sass_Function_List);
// Typedef defining function signature and return type
typedef union Sass_Value* (*Sass_Function_Fn)
  (const union Sass_Value*, Sass_Function_Entry cb, struct Sass_Options* options);
```

Renamed structs/typedefs (probably incomplete):
```C
Sass_C_Function -> Sass_Function_Fn
Sass_C_Function_List -> Sass_Function_List
Sass_C_Function_Callback -> Sass_Function_Entry
Sass_C_Function_Descriptor -> Sass_Function
Sass_C_Import_Fn -> Sass_Importer_Fn
struct Sass_Importer** -> Sass_Importer_List
Sass_C_Import_Callback -> Sass_Importer_Entry
Sass_C_Import_Descriptor -> Sass_Importer
struct Sass_Import** -> Sass_Import_List
struct Sass_Import* -> Sass_Import_Entry
```

### New experimental feature

Custom headers (#960) act like importers, but only run once at the beginning of each compilation. They are primarily meant to define custom mixins (via plugins or context options). Beside that they might be usefull for other cool things I currently cannot think of :smile: Opposite to custom importers, all custom headers will be executed in priority order and all imports will be accumulated (so many custom headers can add various custom mixins or css-code).

### Include in 3.2.0 !?

From a strict versioning stand-point, this should be included in 3.2 since it contains quite a few breaking changes in the API. The new plugins API would also benefit from it. Adding it to 3.2.1 seems wrong and since it is only stabilizing experimental features, I re-scope this to 3.2 (open for further discussions). I touches quite some code though to be added that late in the beta phase. IMO we should do at least another beta round to let node-sass catch-up to these changes.

//CC @xzyfer @am11 @chriseppstein


drewwells:
Possibly related. Is it possible to get a list of files used by the file compiler? This would be super handy for implementor building a file watcher. That rebuilds correctly based off imported file changes.

mgreter:
@drewwells you are probably looking for `included_files`, which is already documented in the wiki!

mgreter:
This PR now pretty much implements all options I envisioned in #962. So at this point I'm not sure if we want to include this in 3.2.0 or wait for 3.2.1. Since it does include a few breaking changes with importers (not that many, but still noticeable). @am11 what do you think, you have time to adjust this in node-sass before 3.0? It basically just involves setting a list of importers instead of a single one (plus passing an additional priority value on creation). Other than that everything can stay as it is.

### How do imports work now?

For each `@import` libsass will call importers in priority order (highest priority first). If the importer returns `0`, libsass will skip to the next one. If no importer returns anything, libsass will continue to load the import on its own. Once an importer returns something (either a list of imports, an empty list of imports or an error), the chain is aborted and only the given return is handled. Additionally I fixed the behaviour when only a filename is returned to be exactly as with plain libsass imports.

am11:
This looks quite interesting @mgreter. :+1: 

I will try to wrap my head around how to provision the priority thing in JS and/or decent way to capture the intent of user that which kind of import they want to handle with certain importer, so to translate that in API call.

Currently, I am under the impression that we will allow user to make their decision and return null; whether certain importer wants to handle certain kind of URI? In which case, the user will return `null` or `undefined` from JS code, we will return nullptr to libsass and libsass will call the next importer in pipe. If there is no next-importer in queue, libsass will fallback to default mechanism.

Please CMMIW.

+1 to add it in v3.2. We already have fistful of breaking changes coming in node-sass v3, why not another! :sunglasses:

mgreter:
@am11 yup, spot on. Also check out custom headers, which are registered exactly the same way as importers. Only difference is that all custom headers will be called and their imports get accumulated. It's really just an array of importers that get called at the start of each compilation. As for the priority, you may be able to attach a priority property to the js function? Like so:

```js
function importer(...) {
  return ['filename', 'div { ... }'];
}
importer.priority = 99;
var options = {
  importers => [ importer ]
}
```
In perl I opted for a simple array ref:
```perl
headers => [
  header1, // will get default = 0
  [ header2, 1] // specific prio
]
```

What priorities people give to their importes is pretty much out of this scope, but I could propose these very personal guidelines, which are probably really just a bad first guess:

```
>0 : headers that only define stuff
<0 : headers that output css code
1 - 100 : for personal stuff
100 - 1000 : for framework stuff
1000 + : for distributed frameworks
```

am11:
This is brilliant! :zap: 

I think we can let node-oriented users return the priority as member of object, instead of doing something fancy/tricky like `this.priority=..`, they will just `return {file: ..., contents: ..., priority: ..}` (or the async variant `done({file: .., contents: ..., priority: ...})`).

As for the custom headers, I will try to add it to v3, but I am not sure if we would have enough time after this gets merged and libsass v3.2.0 stable is released? Will there be another beta or rc, so we have time to include these features and be able to align node-sass v3.0 release with libsass 3.2?

mgreter:
Be aware that the priority has to be passed to libsass before the importer gets called (you cannot include it in the return of the importer). And yes, I could merge this PR and create another beta, so node-sass can catch up and do another beta on its own, so things can get settled! IMO you should be able to copy pretty much all code from custom importers for custom headers.

am11:
> has to be passed to libsass before the importer gets called

I just envisioned it how it will look, and let me tell you this much: this is going to be fun :sunglasses: :sunny: 

We will need to provide the following variants:

* importer: function
  * the priority 0 will be sent to libsass from node-sass binding
* importer: { priority, function }
  * the user-defined priority will be sent to libsass
* importer: [function1, function2, ..]
  * the array index will be sent as priority
* importer: [{priority, function}, {priority, function}, ..]
  * the user-defined priority will be sent.
* importer: [{priority, function}, {priority, function}, {function /* no priority here! */}..]
  * the user-defined priority will be sent, and the index of array for the missing priority.
  * not sure what will happen if there is a tie in priorities (user-defined vs. index-decuded)

Same goes for the headers.

Aside: we really need help from someone to document all that libsass via node-sass has to offer to node.js paradigm.

am11:
After all, this is issue #:100:0 !

am11:
@mgreter, the CI builds are mostly failing due to the ongoing DDOS attach on GH: https://twitter.com/githubstatus/with_replies.

Separately, would it be possible to emit libsass version info in `.vcxproj` as well?

drewwells:
Also, `sass_context_get_included_files_size (struct Sass_Context* ctx)` added to API!

drewwells:
I'm getting cannot find errors on `sass_importer_set_list_entry`. I'm kinda dumb, so I grepped to find answers.

`sass_make_importer_list` has an entry in sass_functions.cpp, but `sass_importer_set_list_entry` is that why it's not working?

mgreter:
It looks like it was renamed to `sass_import_set_list_entry` ...

mgreter:
Maybe helpful: [commit needed for `perl-libsass` to bring it on par with 3.2.0-beta.4] [1]!

[1]: https://github.com/sass/perl-libsass/commit/85fbc0573609ec5e22270ba0160c6fb8fc41800b

drewwells:
That is helpful! The code indexes the C array directly rather than use a `sass_importer_set_entry` (which seems to be different than `sass_import_set_entry`): https://github.com/sass/perl-libsass/commit/85fbc0573609ec5e22270ba0160c6fb8fc41800b#diff-5b390435e30d16e259818ad6b65d1dafR576

chriseppstein:
I am opposed to the custom headers feature. One of the philosophies that Sass has is that we do not support things "magically being available". Only explicit declarations can and should bring in code dependencies. This is why there's no way to set a global variable from the sass command line despite numerous requests. Custom importers are a great way to bring in code that is constructed on the fly. That is how compass sprites worked. They imported the sprite images. IMO, we don't need a new abstraction here, the custom importer concept is enough.

chriseppstein:
Load paths is just a simple way to make a filesystem importer. In this way, importers are actually the first class citizens and it allows importers and filesystem paths to be intermixed. Furthermore, the resolution order should just be a list IMO. I don't understand why we need priority.

drewwells:
Custom functions is the new way of supporting the import overrides in Ruby
Sass. These work very well and less confusing in my opinion. From code I've
seen, it was quite rare for anybody to actually understand the various code
imports Compass supported. They just @import "compass" and hope for the
best. I think the explicit overriding of functions is a far better way to
support these mechanisms.

Since libsass has no custom mixin support, headers fill that gap quite
nicely. I understand your hesitation about custom headers. It does place
magic into the end user code that may have undesirable affects. See the
discussion on https://github.com/sass/libsass/issues/489. True people could
put code in there, but it's not a developer detail it's an implementor
detail. Removing custom headers wouldn't prevent implementors doing it
another way like I've had to for a while now. I am quite happy to now have
custom headers to do this with.

mgreter:
We have two ways to add custom functions, headers and importers in libsass. Via the C-API options for implementers or by loading native libsass plugins. So the priority is needed since we have more than one array to consider. From the libsass implementation it's just a slightly different use of custom importers, so the feature was quite cheap and I do see the benefits in regard to native plugins.

chriseppstein:
I get that custom functions violate the "magically available" principle that I mentioned. As such, I've been proposing that we move away from that pattern (https://github.com/sass/sass/issues/1608#issuecomment-72289559) so that even functions have to be loaded explicitly to be available in a sass file's compile.

I get that you find it useful, but it's moving libsass in a direction that sass itself is moving away from. I don't understand why you think libsass needs "custom mixins". Mixins come from imports. If you want to provide a mixin, you either provide a sass file or you provide a custom importer. The reason we need custom functions is to do things that pure sass cannot do, like filesystem access, or integration with build systems, etc. These functions can return whatever information the mixins need.

chriseppstein:
@mgreter OK, I get that native plugins have this issue, but the priority value is really unintuitive to me. There's no agreement what any priority value means and any scheme that may arise would likely be better implemented in a less granular way. IMO, the priority from native plugins should be limited to `HIGH` and `LOW` where high priority plugins are guaranteed to be considered before the load path and low priority plugins are guaranteed to be considered after the load path. Then, any further resolution should belong to the implementer's API.

drewwells:
Loading specific functions would be rather tedious. Maybe packages that
define sets of functions would be an affective way of making functionality
available. The imports are a good idea for a library of functionality. The
only issue I see is that errors would not give any indication of what went
wrong. Function not available couldn't be tracked back to a missing package
as libsass would no idea what package the function comes from.

Thinking of this problem like most modern language. IDEs provide support
for manipulating the import selection based on the code. This isn't
something Sass users can lean on. Calling "package-function" would add the
necessary compass import. It would actually be easier to track these things
down if rather than @function {function name} was referenced from a package
@function {package}.{function name}. This connects the functions back to
packages and relies less on the 'just works' principle that custom headers
and custom functions provides.

mgreter:
@chriseppstein I agree that it is probably over engineered to expose the priority to the "options" C-API (which is what implementers use). From our point it is just convenient to have exactly the same transfer C objects for both C-APIs. IMO it doesn't make much sense to create new functions in the C-API just to maneuver around this. Instead we should indeed give some guidance which ranges are for what use. I already posted something in that direction:

```
>0 : headers that only define stuff
<=0 : headers that output css code
1 - 100 : for personal stuff
100 - 1000 : for framework stuff
1000 + : for distributed frameworks
```

We also could enfore certain ranges for plugins. But I don't know what you mean by "before the load path". IMO the use case where you simply have a list of importers/headers is possible, as libsass should execute same priorities in registration order (but I have to confess, I didn't really test this).

@chriseppstein is this a showstopper for you for the upcoming 3.2.0 release? The API is still marked highly experimental, so IMO it should be ok for now. IMO we can further discuss this to get to an agreement on a `high` and `low` value after we released a first "draft".

xzyfer:
Apologies if I've missed something, I'm just catching up on all the recent changes.

I think we may have prematurely over-engineered the extensibility points. I think it's worth stepping back and questioning do we really need custom functions, importers, and headers? And what are the real world usecases they solve?

I'm struggling to see what the usecase for custom headers is? As @chriseppstein has mentioned custom mixins don't really make sense.

xzyfer:
For the time being it's my opinion we should disable custom headers in 3.2.0. 

Let's open an RFC to discuss the intention, usecases, and implementation of this feature. The C API will be a major feature of Libsass and it's worth doing the due diligence on changes and features - this is especially important in cases where there is no Ruby Sass / Compass equivalent.

It's probably worth also opening an RFC regarding importer weights. I agree that it should be a list, rather than weight. The pragmatist in me sees this quickly becoming a race to the top ala z-indexes.

xzyfer:
Since this has been merged lets continue the custom import discussion to https://github.com/sass/libsass/issues/1034

PR #1001, opened 2015-03-29
[WIP] add length of included_files array

drewwells:
This extends the API to add the length of the included_files array. Without this, it is difficult to determine the length of the C array exposed via the API interface.

### DIFF
--- a/context.cpp
+++ b/context.cpp
@@ -89,6 +89,7 @@ namespace Sass {
     if (input_path == "") input_path = "stdin";
     if (output_path == "") output_path = "stdout";
 
+    num_included_files = 0;
     include_paths.push_back(cwd);
     collect_include_paths(initializers.include_paths_c_str());
     collect_include_paths(initializers.include_paths_array());
@@ -396,9 +397,9 @@ namespace Sass {
       return includes;
   }
 
-  int Context::get_num_included_files()
+  int Context::get_num_included_files(size_t skip)
   {
-    return num_included_files;
+    return num_included_files - skip;
   }
 
   string Context::get_cwd()
--- a/context.cpp
+++ b/context.cpp
@@ -220,6 +220,7 @@ namespace Sass {
   {
     sources.push_back(contents);
     included_files.push_back(abs_path);
+    num_included_files = num_included_files + 1;
     queue.push_back(Sass_Queued(load_path, abs_path, contents));
     emitter.add_source_index(sources.size() - 1);
     include_links.push_back(resolve_relative_path(abs_path, source_map_file, cwd));
@@ -393,6 +394,11 @@ namespace Sass {
       // the skip solution seems more robust, as we may have real files named stdin
       // includes.erase( std::remove( includes.begin(), includes.end(), "stdin" ), includes.end() );
       return includes;
+  }
+
+  int Context::get_num_included_files()
+  {
+    return num_included_files;
   }
 
   string Context::get_cwd()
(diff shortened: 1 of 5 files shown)


mgreter:
IMO it's actually not very hard :wink: 

```C
char** incs = sass_get_included_files;
while (*incs) { process(*incs); ++incs; } 
```

drewwells:
for perl :p. cgo isn't so gentle!
https://github.com/golang/go/wiki/cgo#turning-c-arrays-into-go-slices

drewwells:
I thought about doing:

```
for i := range sass_get_included_files {
  if sass_get_included_files[i] == nil {
    length = i
  }
}
```

but this might cause bad runtime exceptions

mgreter:
If you need to compute the length beforehand you could use plain C:
```C
size_t l = 0; while (*incs) { ++l; ++incs; } 
```

drewwells:
Yeah, but this makes it available in the API which felt like a good service to others coming across this. Also, num_included_files was already in the interface but not used anywhere: https://github.com/sass/libsass/blob/master/sass_interface.h#L58

mgreter:
`Sass_Interface` is obsolete and will be removed soon! I'd rather have a function return the result from the snippet I posted in the API, than keeping a counter in sync with the null terminated array. I don't see any benefit to have this value cached on our side!

mgreter:
A much shorter and more future-proof (but slightly slower) equivalent would be:
```C
ADDAPI size_t ADDCALL sass_context_get_num_included_files (struct Sass_Context* ctx);
size_t ADDCALL sass_context_get_num_included_files (struct Sass_Context* ctx)
{ size_t l = 0; auto i = ctx->included_files; while (*i) { ++i; ++l; } return l; }
```

mgreter:
Please check #1000 as I have a commit there which includes the code from my latest comment!

am11:
This might be of your interest: [node-sass/src/binding.cpp#L135-L148](https://github.com/sass/node-sass/blob/dc237d98ff119b362fcce7458309e42c967cdcdb/src/binding.cpp#L135-L148):

```c++
  char** included_files = sass_context_get_included_files(ctx); 

  if (included_files) { 
    for (int i = 0; included_files[i] != nullptr; ++i) { 
      // do something
    }
  }
```

QuLogic:
You don't even need the `if`:
```c++
for (; *included_files != nullptr; included_files++) {
 // do something
}
```

am11:
@QuLogic, thanks for the suggestion. It is fixed by 8a194d0.  :+1: 

However, we need the iterator index to insert the values in v8 array.

QuLogic:
No, wait, actually, there might be a case where the array is `NULL`, if parsing fails. Might be safer for you to leave the condition in then, if you're still checking it in all cases.

am11:
Yes we are checking it in all cases, but is there state, when libsass returns `NULL` for `included_files`?

I think it always returns the valid pointer, since it converts C++ vector ([context.hpp#L43](https://github.com/sass/libsass/blob/77c1b23cd76005291cbbb9efbc6cf3c33cee4313/context.hpp#L43)) to array of C strings ([sass_context.cpp#L475](https://github.com/sass/libsass/blob/6a740b1fa57fb79b71fb4a9f392b372f413b83ea/sass_context.cpp#L475)), while allocating memory in all cases: [sass_interface.cpp#L35](https://github.com/sass/libsass/blob/bb10765929c5ca730668d4f4c16282799cacb77e/sass_interface.cpp#L35). 

Should we consider it safe?

mgreter:
Libsass doesn't guarantee it (ie. if you call it too eraly, before compilation has taken place). I would say play safe here, as it doesn't really imrpove the performance anyway!

drewwells:
Closing in favor of #1000. Thanks @mgreter

PR #1005, opened 2015-03-30
libtool: Build PIC objects only

saper:
By default automake/libtool
produce -fPIC shared library libsass.so
and a non-fPIC static library libsass.a

To enable including of libsass.a
statically into bindings like node-sass
that need to produce a shared object
make sure object files in .a are also
compiled with -fPIC.

libsass has previously hardcoded -fPIC
-dPIC in the Makefile.am

This feature is broken when using
libtool 2.4.2.

While here, restore AM_PROG_AR

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,8 +16,10 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([no])])
 
 # Checks for programs.
 AC_PROG_CXX
+AM_PROG_AR
 AC_LANG([C++])
-LT_INIT([dlopen])
+LT_PREREQ([2.4.3])
+LT_INIT([dlopen pic-only])
 
 # Checks for header files.
 AC_CHECK_HEADERS([unistd.h])


QuLogic:
libtool 2.4.3 is _really_ new. I think you'll have to reduce the requirement there.

saper:
Yep, but unfortunately the `pic-only` option is broken in the earlier versions. I think we can let this to sit here for a while until there is a solution or at least @travis-ci upgrades

mgreter:
This also fails in mingw (I know that `AM_PROG_AR` will not work for me)!

QuLogic:
Travis is the least of the problems; since most distros don't even have 2.4.3, it much much too early to be depending on it.

saper:
Yes... In the good ol' days normally "release" tarballs are shipped with a working `./configure` script which generates `libtool` script. The user should not normally need autotools at all to use our library (even if compiling from source and using `./configure`).

But I agree it is a bit early to require this.

PR #1006, opened 2015-03-30
Code: Fixes dynamic exception as per C++11 specs

am11:
* See http://stackoverflow.com/a/13841791/863980.
* Also fixes a case where non-throw exception<br/>
  was throwing in `sass_context.cpp`.


am11:
@xzyfer, this is pure C++11 std compliance (deprecating the cpp11's deprecated feature) and nothing to do with WindOS specific :smiley_cat:

xzyfer:
:+1:

PR #1010, opened 2015-03-31
Enable url function overloading

### LINKED ISSUE #674: Users cannot safely redefine url()
It seems like the problem is isolated to the case where you pass more than 
one argument to @function url($arg1, **$arg2: ''**) { ... } 

My tests shows that:
- attr() works with multiple arguments
- omitting the second argument works
- local function variables and global works  

**Note!** When passing the second argument the last character of the first argument is removed

````SCSS

$base-path:'../images/';
$base-attr:'data-';

@function url($src, $path:''){
  @return unquote('url('+$base-path + $path+ $src +')');
}
@function url2($src, $path:''){
  @return unquote('url('+ $base-path + $path+ $src +')');
}
@function attr($arg1, $arg2:''){
  @return unquote('attr('+$base-attr + $arg1 + $arg2 +')');
}

div {
    background: url('image.png');
    background: url('image.png','img/');
    background: url2('image.png','img/');
  
  &:after {
    content: attr(value);
    content: attr(value, -extra);
    content: url('icon.png');
    content: url('icon.png','gfx/');
    content: url2('icon.png','gfx/');
  }
}


````

````CSS
/* Output */
div {
  background: url(../images/image.png);
  background: url(../images/image.pn','img/); /* <= Fails */
  background: url(../images/img/image.png); 
}
div:after {
  content: attr(data-value);
  content: attr(data-value-extra);
  content: url(../images/icon.png);
  content: url(../images/icon.pn','gfx/); /* <= Fails */
  content: url(../images/gfx/icon.png); 
}


/* Expected */
div {
  background: url(../images/image.png);
  background: url(../images/img/image.png);
  background: url(../images/img/image.png);
}
div:after {
  content: attr(data-value);
  content: attr(data-value-extra);
  content: url(../images/icon.png);
  content: url(../images/gfx/icon.png);
  content: url(../images/gfx/icon.png);
}




<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6478672-users-cannot-safely-redefine-url?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by mgreter**
@xzyfer maybe give it a review and we probably also want to include in 3.2.0? I don't really see any reason that tests shouldn't have catched any obvious regressions this could introduce!
And it does fix #674 (the refactor made the code actually a bit slimer)!

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -266,8 +266,18 @@ namespace Sass {
           import_single_file(imp, lexed);
         }
       }
-      else if (peek< uri_prefix >()) {
-        imp->urls().push_back(parse_value());
+      else if (lex< uri_prefix >()) {
+        Arguments* args = new (ctx.mem) Arguments(pstate);
+        Function_Call* result = new (ctx.mem) Function_Call(pstate, "url", args);
+        if (lex < uri_value >()) { // chunk seems to work too!
+          String* the_url = parse_interpolated_chunk(lexed);
+          *args << new (ctx.mem) Argument(the_url->pstate(), the_url);
+        }
+        else {
+          error("malformed URL", pstate);
+        }
+        if (!lex< exactly<')'> >()) error("URI is missing ')'", pstate);
+        imp->urls().push_back(result);
       }
       else {
         if (first) error("@import directive requires a url or quoted path", pstate);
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by mgreter** [comment]
CI seems finally happy! IMO this is ready to be merged!

**Comment by xzyfer** [comment]
Could you add a spec for this to prevent future regressions?

File: parser.cpp [hunk 1]
```diff
@@ -1118,10 +1130,10 @@ namespace Sass {
   {
     Expression* conj1 = parse_conjunction();
     // if it's a singleton, return it directly; don't wrap it
-    if (!peek< sequence< kwd_or, negate< identifier > > >()) return conj1;
+    if (!peek_css< kwd_or >()) return conj1;
```
xzyfer: Does peek_css force word boundaries? Otherwise this is a regression.

mgreter: No, but [`kwd_or`] [1] is forced to have word boundaries :wink: 

[1]: https://github.com/sass/libsass/blob/master/prelexer.cpp#L526

xzyfer: Aha that's new. Cool.

mgreter:

xzyfer:
I think this can wait. It's an edge case case. Now is probably a good time
for a feature freeze, thoughts?

mgreter:
I wanted to have the real feature freeze when the next node-sass beta is out. As I have written in #1020, we should create a new beta once node-sass is ready to release its next beta. IMO we can add a few more bugfixes until then. But from then on I would say we close the window and really only commit bugfixes. BTW. the original issue was marked as a bug, so technically this is a bugfix :wink:

xzyfer:
Lol technicalities. Ship it.

PR #1012, opened 2015-03-31
[WIP] Remove dead code paths in output

mgreter:
We seem to carry around quite a lot of dead code paths as coveralls is showing!


mgreter:
I thought I was going insane, since I kept removing dead code paths, but somehow the coverage score still managed to decrease. Turns out pretty much the whole isPrintable code is actually not needed for the spec tests. Maybe we are missing some tests or I don't see the reason why so much code seems to be unnecessary, but results seem to be valid!?

PR #1024, opened 2015-04-01
Enable Travis OS X Support

mgreter:
Basically just uncommented the prepared osx support in `.travis.yml` and it worked!
Additionally removed gcc compiler target for osx since it is only an alias to clang!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: cpp
 
 os:
   - linux
-  # - osx
+  - osx
 
 compiler:
   - gcc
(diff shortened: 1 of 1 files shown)


mgreter:
Actually I'm tempted to just merge this now, since it compiles correctly on Mac OS-X with native Makefiles and Autotools build :sunglasses: Actually we should probably try to disable gcc build, since gcc seems to be just an alias for clang on OS-X (which makes sense). //CC @xzyfer

xzyfer:
Nice!

xzyfer:
> Actually we should probably try to disable gcc build, since gcc seems to be just an alias for clang on OS-X (which makes sense).


Agreed.

Issue #592, opened 2014-10-29
Deep @extend specificity problems


mzgol:
The following SCSS input:
```scss
%a {
    &::-webkit-scrollbar {
        color: green;
    }
}
%b {
    @extend %a;
}

.a {
    .b {
        @extend %a;
    }

    .c {
        .b {
            @extend %b;
        }
    }
}
```
is compiled via Ruby Sass (3.4.5) to:
```css
.a .c .b::-webkit-scrollbar, .a .b::-webkit-scrollbar {
  color: green; }
```

and via libsass (using node-sass 1.1.4) to:
```css
.a .b::-webkit-scrollbar {
  color: green; }
```

mgreter:
This is directly related to https://github.com/sass/libsass/issues/823

xzyfer:
Looks like this made it into 3.2.0. Updated the milestone.

PR #1032, opened 2015-04-01
Fix `@extend` specificity problems

mgreter:
Fixes https://github.com/sass/libsass/issues/592


mgreter:

xzyfer:
Seems to work :+1:

Issue #823, opened 2015-01-07
Nested @extends possible regression


LaurentGoderre:
This might be related to #592, the following code is compiling fine in libsass 2.1-beta but no in 3.x

```scss
%test {
    > {
        .red {
            color: #F00;
        }
    }
}

p {
    @extend %test;

    > {
        a {
            @extend %test;
        }
    }
}
```

In 2.1-beta I get

```css
p > .red, p > a > .red {
  color: #F00; 
}
```

and in 3.x I get

```css
p > .red {
  color: #F00; 
}
```

torkiljohnsen:
I am experiencing something similar. I set up a CodePen snippet where you can compare how CodePen compiles and how LibSASS compiles (added the LibSASS compiled source in a comment). Click the eye icon to toggle compiled CSS view:

http://codepen.io/torkil/pen/BydJQr?editors=110

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/260.

PR #1033, opened 2015-04-02
Fix `is_superselector_of` function

mgreter:
Fixes https://github.com/sass/libsass/issues/823

### DIFF
--- a/ast.cpp
+++ b/ast.cpp
@@ -336,30 +336,34 @@ namespace Sass {
     // check for selectors with leading or trailing combinators
     if (!lhs->head() || !rhs->head())
     { return false; }
-    Complex_Selector* l_innermost = lhs->innermost();
-    if (l_innermost->combinator() != Complex_Selector::ANCESTOR_OF && !l_innermost->tail())
-    { return false; }
-    Complex_Selector* r_innermost = rhs->innermost();
-    if (r_innermost->combinator() != Complex_Selector::ANCESTOR_OF && !r_innermost->tail())
-    { return false; }
+
     // more complex (i.e., longer) selectors are always more specific
     size_t l_len = lhs->length(), r_len = rhs->length();
     if (l_len > r_len)
     { return false; }
 
     if (l_len == 1)
-    { return lhs->head()->is_superselector_of(rhs->base()); }
+    {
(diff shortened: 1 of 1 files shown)


mgreter:

xzyfer:
Maybe this is best left for when selector functions lands.

mgreter:
I actually thought it would make sense to have it merged before selector functions, since it could possibly have an impact there (they may suffer from the same bug). So I thought it would be good to already have it merged. Otherwise chances are that this bug will be fixed twice !?

xzyfer:
Cool, go for it. I just thought it might be a risky change but it seems fine

mgreter:
Pinged @onedayitwillmake so he maybe can verify this changes to be sane. I guess he should have a few additional spec tests by now, so maybe he can test if this brakes any existing behavior.

mgreter:
OK, just made some tests with the WIP branch and it seems alter results for the new selector function tests. So I will leave this open for now!

mgreter:
OK, lets wait for selector-functions and the new spec tests which should cover this topic better.
IMO the selector functions cannot be merged without a proper superselector handling anyway!

onedayitwillmake:
@mgreter I agree, i can merge that in. The selector function, is essentially parsing the selectors, returning the result of calling that method

mgreter:
There is a very strange and IMO wrong logic at work with ruby sass and `is_superselector`. I fail too see why ruby sass produces different results for these two tests:

```css
test-a {
  // false
  $a: "a + c + d";
  $b: "a + b + c + d";
  a: is_superselector($a, $b);
}

test-b {
  // true
  $a: "a + c + d + e";
  $b: "a + b + c + d + e";
  a: is_superselector($a, $b);
}
```

IMO I found the code part in ruby sass that is mainly responsible for this behavior:

```ruby
# .foo > .baz is not a superselector of .foo > .bar > .baz or .foo >
# .bar .baz, despite the fact that .baz is a superselector of .bar >
# .baz and .bar .baz. Same goes for + and ~.
return if seq1.length == 3 && seq2.length > 3
```

With this info I might be able to re-produce the IMO faulty behavior ...

mgreter:
This is included in my WIP branch for 3.3!

PR #1047, opened 2015-04-05
Fix `str-slice` function to work with utf8 strings

mgreter:


functions.cpp
```diff
@@ -906,28 +906,25 @@ namespace Sass {
       string newstr;
       try {
         String_Constant* s = ARG("$string", String_Constant);
-        Number* n = ARG("$start-at", Number);
-        Number* m = ARG("$end-at", Number);
+        double start_at = ARG("$start-at", Number)->value();
+        double end_at = ARG("$end-at", Number)->value();
```
QuLogic: Does it really need to be `double`, and not something integral?

mgreter: No, it will be used as a `size_t` anyway. But the value passed by `value()` is a double, so I thought this makes it more clear that there is a conversion (and we may have to add correct behavior according to ruby sass what happens if non integers are passed in).

PR #1050, opened 2015-04-06
Fix separator preserving with rest args

mgreter:
Makes it pass these todo spec tests:
- libsass/var-args
- scss-tests/071_test_mixin_splat_args_with_var_args_preserves_separator
- scss-tests/090_test_function_splat_args_with_var_args_preserves_separator


xzyfer:
:+1:

PR #1055, opened 2015-04-06
Relax compact error to warning

mgreter:
This relaxes the error to a warning. Otherwise the function cannot be used by user-code. It will now just spit out a warning, but behavior in output matches ruby sass. Makes a few spec tests to pass now:

- libsass/eq
- libsass/compact

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -1619,6 +1619,10 @@ namespace Sass {
   {
     lex< identifier >();
     string name(lexed);
+    if (name == "compact") {
+      warn("`compact` has been removed from libsass because it's not part of the Sass spec", pstate);
+    }
+
     ParserState call_pos = pstate;
     Arguments* args = parse_arguments(name == "url");
     return new (ctx.mem) Function_Call(call_pos, name, args);
--- a/context.cpp
+++ b/context.cpp
@@ -512,7 +512,6 @@ namespace Sass {
     register_function(ctx, index_sig, index, env);
     register_function(ctx, join_sig, join, env);
     register_function(ctx, append_sig, append, env);
-    register_function(ctx, compact_sig, compact, env);
     register_function(ctx, zip_sig, zip, env);
     register_function(ctx, list_separator_sig, list_separator, env);
     // Map Functions
--- a/error_handling.cpp
+++ b/error_handling.cpp
@@ -8,6 +8,18 @@ namespace Sass {
   : type(type), pstate(pstate), message(message)
   { }
 
+  void warn(string msg, ParserState pstate)
+  {
+    cerr << "Warning: " << msg<< endl;
+  }
+
+  void warn(string msg, ParserState pstate, Backtrace* bt)
+  {
+    Backtrace top(bt, pstate, "");
+    msg += top.to_string();
+    warn(msg, pstate);
+  }
+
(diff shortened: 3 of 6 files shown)


xzyfer:
My preference is we simply just drop this error/warning completely. Mentioning it as a breaking change should be enough IMO. I don't see the benefit in keeping this code debt around.

mgreter:
How about relaxing the error in 3.2.1 to the warning and removing it completely in 3.3.0 (3.2.2)?

xzyfer:
Sure thing!

xzyfer:
Thinking about this does it makes sense to go form error -> warning -> gone?

Ship this in 3.2.0 instead?

mgreter:
When did we introduce that error message (was a version already released with that message included)? IMO it does make sense to make people aware that the behavior has changed! But personally I don't really have a preference here ...

xzyfer:
I believe it was introduce in 3.2.0. 

IMO error -> warning -> gone make no sense. Either
- warn -> error -> gone
- error -> gone

mgreter:
> I believe it was introduced in 3.2.0. 

So we haven't even released the "deprecated" error yet. Problem is that the error will disallow the use of custom functions named "compact". But we IMHO still want to inform the user that his code will no longer work the way it did. The errors seems to be a bit too much, but warnings might get overlooked by users. So IMHO it makes sense to first use an error (failing hard), then emit a soft warning and finally ignore it altogether. This should make a nice transition for users that use this function, but an error also disallows any "polyfill" implementation, so maybe going with a warning is the best option?

Actually I don't see why "warn > error > gone" would make any sense at all?

xzyfer:
>So we haven't even released the "deprecated" error yet.

Correct.

>Actually I don't see why "warn > error > gone" would make any sense at all?

warn > error > gone and error -> gone are patterns used by the Ruby Sass team.

>an error also disallows any "polyfill" implementation

This makes erroring a non-starter for me. I had assumed native (libsass) functions would have the lowest priority in the function resolution process.

>maybe going with a warning is the best option

I think warning -> gone is the best option here unless we change the resolution order of function calls.

parser.cpp
```diff
@@ -1608,6 +1608,10 @@ namespace Sass {
   {
     lex< identifier >();
     string name(lexed);
+    if (name == "compact") {
```
xzyfer: Would not incorrectly warn users even if they have defined their own `compact` function? Could we move the warning into the functions.cpp?

mgreter: I also saw that, but wasn't sure how to better address it. But we surely can move that to some more appropriate place. IMO we might be able to add a specific if clause just before internal functions are resolved, and if the name is `compact` emit this warning!?

xzyfer:
@mgreter I think I have misunderstood the issue here. 

> Otherwise the function cannot be used by user-code

I thought you meant the current master made it impossible for users to define their own `compact` method. This does not appear to be the case as this code execute fine.

```csss
@function compact() {
  @return foo; }

foo {
  bar: compact(); }
```

> Makes a few spec tests to pass now

The specs you've mentioned are invalid and should be deleted as `compact` has never been a supported function in Ruby Sass.

***

I think we should leave the current behaviour and remove the error completely 3.2.1/3.3. Thoughts?

Issue #1060, opened 2015-04-07
@elseif is not recognized


mzgol:
The following SCSS:
```scss
div {
    @if true {
        display: block;
    } @elseif true {
        display: flex;
    }
}
```
is translated via libsass `3.2.0-beta.5` to:
```css
div {
  display: block; }

@elseif true {
  div {
    display: flex; } }
```
Ruby Sass translates it to:
```css
div {
  display: block; }
```
Changing `@elseif` to `@else if` resolves the issue.

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/306

xzyfer:
This is fixed and will be 3.2.0

mzgol:
Thanks! :)

PR #1062, opened 2015-04-07
Fix @elseif keyword not being parsed

xzyfer:
This PR fixes the `@elseif` keyword not being parsed.

Fixes https://github.com/sass/libsass/issues/1060. Specs added https://github.com/sass/sass-spec/pull/306.

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -1630,12 +1630,13 @@ namespace Sass {
     if (!peek< exactly<'{'> >()) error("expected '{' after the predicate for @if", pstate);
     Block* consequent = parse_block();
     Block* alternative = 0;
-    if (lex< kwd_else_directive >()) {
-      if (peek< exactly<if_after_else_kwd> >()) {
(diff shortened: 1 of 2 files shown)


mgreter:
@xzyfer out of curiosity. Did that happen because of `word` matcher? I guess it was parsing `sequence < if_kwd, else_kwd >`. If so I actually like it since it shows that it works as intended :wink:

xzyfer:
> Did that happen because of word matcher?

Yes. Essentially Sass allows both `@else if()` and `@elseif ()`. Due to the word boundary the old check failed.

```
if (lex< kwd_else_directive >()) { //@else
  if (peek< exactly<if_after_else_kwd> >()) { // should match @elseif but only matches @else if
```

xzyfer:
Now that I'm looking at it again (and it's not 2am) the `alternatives` in `elseif_directive` should be removed since the semantics of `@else` are different to `@else if`

PR #1066, opened 2015-04-07
Fix backslash multiline string with win linefeed

mgreter:
OK, that was a dumb coding error which I obviously did not test :cry:

Fixes https://github.com/sass/libsass/issues/1067


am11:
This is super fast! :smiley_cat: :tada:

mgreter:
It can't get much faster than opening the PR to fix the issue before the actual issue is opened :grinning: :lollipop:

PR #1069, opened 2015-04-07
Improve `lex_css` to restore old token on non-match

mgreter:
We still parse and throw away any possible comments!
This is done only once for multiple calls (performance)!


xzyfer:
:+1:

PR #1071, opened 2015-04-07
[WIP] Created an Xcode project to build libsass as a static-libary

onedayitwillmake:
This is a simple pull request to be able to build / work on the project from Xcode on OSX taking advantage of the debugger to step through the code.

It is built as a static-library, which is placed in $PROJECT/lib (as in the Makefile), with a simple attached project that contains a simple main function to allow stepping through the code when debugging.

Cleaned up this PR after using https://github.com/sass/libsass/pull/968 as testing ground

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,25 @@ support/libsass.pc
 # Cloned testing dirs
 sassc/
 sass-spec/
+
+# Xcode
+
+xcode/static-library/build/
+xcode/dev/build/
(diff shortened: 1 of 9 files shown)


onedayitwillmake:
@mgreter Although it's building correctly, I'm having a bit of trouble with linking to the installed .dylib on travis.

Any advice? Imo i think it would also be fine to simply not build the shared library version on OSX/Xcode since it's less used

PR #1072, opened 2015-04-08
Fix incorrect semantics of @else

xzyfer:
This PR fixes a regression in the semantics of `@else` introduced (by me) in #1062.

Spec updated https://github.com/sass/sass-spec/pull/308.

### DIFF
--- a/prelexer.cpp
+++ b/prelexer.cpp
@@ -262,10 +262,9 @@ namespace Sass {
       return word<else_kwd>(src);
     }
     const char* elseif_directive(const char* src) {
-      return alternatives< word<else_kwd>,
-                           sequence< exactly< else_kwd >,
-                                     optional_css_whitespace,
-                                     word< if_after_else_kwd > > >(src);
+      return sequence< exactly< else_kwd >,
+                                optional_css_whitespace,
+                                word< if_after_else_kwd > >(src);
     }
 
     const char* kwd_for_directive(const char* src) {


prelexer.cpp
```diff
@@ -262,10 +262,9 @@ namespace Sass {
       return word<else_kwd>(src);
     }
     const char* elseif_directive(const char* src) {
-      return alternatives< word<else_kwd>,
-                           sequence< exactly< else_kwd >,
-                                     optional_css_whitespace,
-                                     word< if_after_else_kwd > > >(src);
+      return sequence< exactly< else_kwd >,
+                                optional_css_whitespace,
+                                word< if_after_else_kwd > >(src);
```
mgreter: Shouldn't it be (untested):
```
return word <
         sequence<
           exactly< else_kwd >,
           optional_css_comments,
           exactly< if_after_else_kwd >
         >
       >(src);
```

The `word` matcher only checks if it is not followed by a `alnum` character. So it doesn't care if the thing inside has any spaces (I guess this might be confusing). IMO your solution does exactly the same, but IMO indentation is off. And btw. I don't know, but I actually prefer to write each lexer on its own line, it's much better readable to me, specially with the correct indentation. Just wanted to ask as I tend to re-write them and wanted to get consens!?

xzyfer:
> Shouldn't it be [..] `word<>`

Good question. I wonder what's more in line with Ruby Sass. As it stands this is behaves correctly. but the error message when something like `@elseif-foo` is encountered is worth comparing.

> IMO indentation is off

Yep my bad.

>I actually prefer to write each lexer on its own line [..] Just wanted to ask as I tend to re-write them and wanted to get consens!

Have you got an example? The style you had with the IE filter refactor is generally what we do. I'm aim for more lines rather than less. Code can never been too readable IMO.

mgreter:
Example from prelexer.cpp:

current:
```C
return sequence< optional<sign>,
                       optional<digits>,
                       exactly<'n'>,
                       zero_plus < space >,
                       sign,
                       zero_plus < space >,
                       digits >(src);
```

I would write it as:
```C
return sequence<
         optional < sign >,
         optional < digits >,
         exactly < 'n' >,
         zero_plus < space >,
         sign,
         zero_plus < space >,
         digits
       >(src);
```

I mainly mean two things. The `optional<sign>` goes on its own line and indentation is two from previous matcher/lexer. If more than one arguments is in `<...>`, they shall go on separate lines.

xzyfer:
Sounds good. The current convention is to line up children with first child. I find this weird.

```
return sequence< optional<sign>,
                 optional<digits>,
                 exactly<'n'>,
                 zero_plus < space >,
                 sign,
                 zero_plus < space >,
                 digits >(src);
```

It's gets super unwieldy with mildly complex patterns like `identifier`. :+1:

Issue #1056, opened 2015-04-06
sourcemap regressions (3.2 beta.5)


Snugug:
As part of our test cases for [gulp-sass](https://github.com/dlmanning/gulp-sass/tree/2.x) we have a test to ensure Source Maps work. In `3.2.0-beta.2` of Libsass (`node-sass 3.0.0-beta.4`), our source map tests work as expected [with this test file](https://github.com/dlmanning/gulp-sass/blob/2.x/test/scss/inheritance.scss). In `3.2.0-beta.5`, it appears as if some `source_index` numbers are now returning `-1`, which is wreaking all sorts of havoc because I'm not entirely sure `-1` is a valid source index. If it helps we're passing in a string through `node-sass` instead of the file name.

**3.2.0-beta.2**
```
{ version: 3,
  file: 'inheritance.scss',
  sources:
   [ 'inheritance.scss',
     'includes/_cats.scss',
     'includes/_dogs.sass' ],
  sourcesContent: [],
  mappings: 'ACGA,IAAI,CAAC;EACH,UAAU,EAAE,IAAK,GADb;;ACAN,MAAM,CAAC;EACL,UAAU,EAAE,GAAI,GADV;;AFAR,MAAM,EAAE,SAAS,CAAV;EACL,MAAM,EAAE,IAAK;EACb,UAAU,EAAE,IAAK,GAFX;;AAKF,MAAM,AAAA,UAAU,EAAE,UAAU,AAAA,SAAS,CAA1B;EACf,SAAS,EAAE,KAAK;EAChB,WAAW,EAAE,IAAK,GAFF;;AAKlB,SAAS,CAAC;EAER,YAAY,EAAE,GAAG,GAFR',
  names: [] }
```

**3.2.0-beta.5**
```
{ version: 3,
  file: 'inheritance.scss',
  sources:
   [ 'inheritance.scss',
     'includes/_cats.scss',
     'includes/_dogs.sass' ],
  sourcesContent: [],
[...]

PR #1073, opened 2015-04-08
Fix source-map regression

mgreter:
Once again `debug_ast` to the rescue :crystal_ball: 

I really hope this fixes the reports in #1056! It does not fix source-maps in general, only bugs introduced during the re-factoring for 3.2.0 release! My own initial test suite with 3 very simple tests shows one difference to 3.1 (but it only points to something more specific, not to something wrong). Other than that I can confirm that with this patch it works a lot better and hopefully as before.

We are aware that source-maps are not working very well currently. A lot of work already has been put into the necessary refactorings to correctly implement source-maps in the long run. The main problem is that libsass parses selectors and other stuff twice or more (ie. when expanding rulesets), so this makes the whole thing even more complicated than it already is.

### DIFF
--- a/debugger.hpp
+++ b/debugger.hpp
@@ -32,28 +32,45 @@ inline string longToHex(long long t) {
   return is.str();
 }
 
+inline string pstate_source_position(AST_Node* node)
+{
+  stringstream str;
+  Position start(node->pstate());
+  Position end(start + node->pstate().offset);
+  str << (start.file == string::npos ? -1 : start.file)
+    << "@[" << start.line << ":" << start.column << "]"
+    << "-[" << end.line << ":" << end.column << "]";
+  return str.str();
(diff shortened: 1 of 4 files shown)


am11:
:+1: :100:

And this is one of the **huge** advancement in libsass IMO. Source-maps are very much promising that what we had few months ago. I remember when I reverse engineered the libsass generated source-map to fix the bugs for [WE2013 in C++++ (#)](https://github.com/madskristensen/WebEssentials2013/blob/c87673d8dbad6a4013cc04325fe830d2dc4b275b/EditorExtensions/Shared/Helpers/CssSourceMap.cs#L115-L223), which I think should be removed now. :smile_cat:

xzyfer:
:+1:

Issue #1074, opened 2015-04-08
Regression in selector interpolation


xzyfer:
The follow selector interpolation worked in 3.1.0 but fails with 3.2.0-beta.1 and above

```scss
$i: 1;
.foo#{-$i} { a:b }
```

3.2.0-beta.2 and above produce the following error
```
Error: invalid selector after .foo-
        on line 1 of test.scss
```

3.2.0-beta.1 also fails but reports a different line. A git bisect reports this regression was introduced in https://github.com/sass/libsass/commit/7877cfa872b036d55344db5d12ea908bb70f3705

The following variations work in all tested releases
```scss
$i: 1;
.foo-#{$i} { a:b }
.foo#{-1} { a:b }
.foo-#{1} { a:b }
```

Spec added https://github.com/sass/sass-spec/pull/309

Can you take a look @mgreter ?

mgreter:
Error seems to happen in expand, since it seems to be parsed correctly:

```
String_Schema 0x2c322b0 (0@[0:6]-[1:0]) 4
 String_Quoted : 0x2a41dc0 (0@[0:6]-[1:0]) [.foo] <>
 Unary_Expression 0x2a41e40 (0@[0:6]-[0:7]) [1]
  operand: Variable 0x2c84be0 (0@[0:7]-[0:9]) [$i]
```

My guess is that the unary expression is not correctly or not at all evaluated!

PR #1076, opened 2015-04-08
Fix interpolation with unary expressions

mgreter:
Fixes https://github.com/sass/libsass/issues/1074

Btw. we might be able to reduce or completely get rid of the interpolation function. It was mainly added to not "step on any toes" while doing the refactoring. For now it makes it easier to keep the code separated from the regular eval visitors. IMO it makes sense to revisit this once interpolation has proven to do the correct thing (I still know a few very esoteric edge-cases).


xzyfer:
Agreed. That function is scary. Maybe even breaking it into an interpolation visitor could be an elegant solution.

I also agree it can wait til 3.4+

xzyfer:
Merged this because I need it for my stuff. Great work @mgreter

Issue #1075, opened 2015-04-08
Rest parameters not expanded with call


xzyfer:
The following fails in 3.2.0-beta.5

```scss
$args: ("color": #ff0000, "amount": 10%);
foo {
  bar: call(lighten, $args...);
}
```

with the following error

```
Error: required parameter $amount is missing in call to function lighten
        on line 3 of test.scss
>>   bar: call(lighten, $args...);
```

However using rest parameters without `call` works as expected

```scss
$args: ("color": #ff0000, "amount": 10%);
foo { 
  bar: lighten($args...);
}
```

The expected output is

```css
foo {
  bar: #ff3333; }
```

Specs added  https://github.com/sass/sass-spec/pull/310

xzyfer:
I need this for my own work so I'm going to squeeze it into 3.2.0

PR #1077, opened 2015-04-08
Fix argument handling for dynamic `call` function

mgreter:
Pass argument flags (keyword/rest) to called function! (Fixes https://github.com/sass/libsass/issues/1075)


xzyfer:
:+1:  thanks!

xzyfer:
Merged this because I need it for my stuff. Great work @mgreter

PR #1078, opened 2015-04-08
Backtrace lines are off-by-one

saper:
As reported in sass/node-sass#843

### DIFF
--- a/backtrace.hpp
+++ b/backtrace.hpp
@@ -41,7 +41,7 @@ namespace Sass {
              << "\t"
              << (++i == 0 ? "on" : "from")
              << " line "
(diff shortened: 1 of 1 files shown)


saper:
As reported in sass/node-sass#843

mgreter:
LGTM! IMO I have changed the positions internally to start from 0, as it always is in C. Was pretty confusing otherwise and we don't want to waste a single bit :smile:

saper:
Well I started digging into the code and it seems that backtrace gets constructed also in `static int handle_errors(Sass_Context* c_ctx)` manually, but I think `node_sass` gets one from the Backtrace object.

Issue #1086, opened 2015-04-10
Map keys are not evaluated


OnkelTem:
This code produces error
```sass
$map: (-1px: 12);
```
Error message: 
```
  undefined:undefined  _Map_base::at
```
Also note 'undefined:undefined' standing for the line and column number — they are missed. 

Using node-sass v3.0.0-beta.5 (pointing to libsass v3.2.0-beta.5)

I believe this temporary workaround can be used:
```sass
$minus1px: -1px;
$map: ($minus1px: 12);
```

mgreter:
This is actually not really a regression, although that exact case was working before. The reason for this is, that things like the following did not work in libsass 3.1.0:

```scss
$map: (0 - 1px: 12);
foo { bar: inspect($map); }
```

We changed the way we parse negative numbers to wrap them into an unary_operator. This is more in line with the actual meaning and solved problems in other places, but unfortunately brings up another problem to the surface here. IMO we need to eval the keys before "using" them.

mgreter:
This specific issue (and a few more) should be fixed by #1090. I'm quite sure that we have more subtile problems with map handlings, but this should at least fix quite a lot of the more common use-cases!

PR #1090, opened 2015-04-10
Improve hash map key handling

mgreter:
Implement a few other key types (Addresses https://github.com/sass/libsass/issues/1086)


eval.cpp
```diff
@@ -10,6 +10,7 @@
 #include "backtrace.hpp"
 #include "prelexer.hpp"
 #include "parser.hpp"
+#include "debugger.hpp"
```
xzyfer: remove this?

xzyfer:
:+1:

Issue #1098, opened 2015-04-13
Erroneous space inserted in { property: value-as-digit\9; }


atdrago:
Adding a `\9` after a property value that ends in a digit adds an erroneous space before the `\`.

For example,
```sass
div { opacity: 1\9; }
```
Should output
```css
div { opacity: 1\9; }
```
But instead outputs
```css
div { opacity: 1 \9; }
```

Ruby sass compiles the value without the inserted space.

xzyfer:
Thanks for the report @atdrago. Could you please also provide a test case in sass-spec? Here's a good example of one https://github.com/sass/sass-spec/pull/315.

@mgreter this is a regression from 3.1.0. I'll do a bisect shortly.

atdrago:
Thanks for the response, @xzyfer. Added a PR sass/sass-spec#327

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/331. Turns out this regression is more server then first suspected.

```scss
div { 
    opacity: 1\9;
    width: 500px\9;
    color: #f00\9\0\;
}
```
Ruby Sass
```css
div {
  opacity: 1\9;
  width: 500px\9;
  color: #f00\9\0\;; }
```
Libsass 3.1.0
```css
div {
  opacity: 1\9;
  width: 500px\9;
  color: #f00 \9\0\;; }
```
Libsass 3.2.0-beta.5
```css
div {
  opacity: 1 \9;
  width: 500px \9;
  color: #f00\9\0\\; } // extra /
```

xzyfer:
@mgreter I've started a WIP https://github.com/sass/libsass/pull/1111 before I noticed the other error cases.

I'm not tied to this PR so if you think you can fix this quickly go for it. I'm really want to ship the final beta asap :tada:

mgreter:
IMO it's related to our way to parse every list as a space or comma separated list. Not sure how ruby sass handles this internally. Since this syntax is used as a IE css hack, we could make sure that the way we output it currently will still be valid for those old IE UAs. If that is the case, it is IMO not a real blocker for the next release. We do allow escaped sequences in identifiers (`static_value`), but all the above cases are not valid identifiers. IMO a hotfix for certain use-cases is possible, but to support this correctly we need the non-spaced list parsing (which is IMO not possible in the scope of the current beta).

xzyfer:
>IMO it's related to our way to parse every list as a space or comma separated list.

This is true.

>IMO not a real blocker for the next release

[...]

Issue #1092, opened 2015-04-10
Missing space after a semi-colon if interpolation used


mzgol:
The following SCSS:
```scss
a {
    $s: "";
    transform: translateZ(0) #{$s};
}
```
Ruby Sass 3.4.13 transforms it to:
```css
a {
  transform: translateZ(0) ; }
```
libsass 3.2.0-beta.5 transforms it to:
```css
a {
  transform: translateZ(0) ;}
```
Note the missing space after a semi-colon.

If interpolation (`#{}`) is removed, the space appears.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/323

hcatlin:
I can confirm that this issue has been solved in the latest release.

hcatlin:
Aaaand, I'm mistaken.

PR #1111, opened 2015-04-16
Fixes a regression in handling common IE property hacks

xzyfer:
This PR fixes a regression in handling common IE property hacks.

Fixes https://github.com/sass/libsass/issues/1098. 
Specs added https://github.com/sass/sass-spec/issues/331

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -1396,7 +1396,8 @@ namespace Sass {
     Token str(lexed);
     --str.end;
     --position;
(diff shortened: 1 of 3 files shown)


xzyfer:
I believe this this correct fix, it's however failing due an issue in the code generation phase erronously added a space before the ending `;` similar https://github.com/sass/libsass/issues/1092.

PR #1118, opened 2015-04-20
Fix interpolation edge cases with escapes

mgreter:
Fixes https://github.com/sass/libsass/issues/1115

### DIFF
--- a/debugger.hpp
+++ b/debugger.hpp
@@ -344,7 +344,8 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
     cerr << " (" << pstate_source_position(node) << ")";
     cerr << " [name: " << block->name() << "] ";
     cerr << " [type: " << (block->type() == Sass::Definition::Type::MIXIN ? "Mixin " : "Function ") << "] ";
-    cerr << " [signature: " << block->signature() << "] ";
+    // this seems to lead to segfaults some times?
+    // cerr << " [signature: " << block->signature() << "] ";
     cerr << " [native: " << block->native_function() << "] ";
     cerr << " " << block->tabs() << endl;
     debug_ast(block->parameters(), ind + " params: ", env);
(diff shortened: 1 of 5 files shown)


xzyfer:
:+1:

xzyfer:
I've created the spec for this issue - https://github.com/sass/sass-spec/pull/340

This PR is an improvement on the current behaviour, however it still fails this spec.

xzyfer:
So I've spent all weekend getting my head around our string handling code, and feel it's already over complicated. I'm super reluctant to add another flag to our String AST nodes i.e. `String::sass_fix_1291`

I wonder if we can tackle this case entirely in the Eval phase instead?

mgreter:
@xzyfer I fully agree, but the handling by ruby sass is pretty complex and AFAIR I was able to get rid of quite a few flags with my last refactoring. But I pretty much gave up to understand how ruby sass internally handles interpolations and escape sequences (We keep finding new edge cases all the time). So for the time beeing I really only see the way to implement specific code paths for these cases, until we have a good enough spec suite to tackle another refactoring. My last commit should solve the reported issues, but it is far away from beeing pretty. I'm having problems coming up with ideas how to name all those different escape/unescape functions, but I currently don't see a way to get rid of them.

mgreter:
Btw. I have used this extended test case:

```scss
foo {
    bar: "x\79";
    baz: "#{x}\79";
    bar: "x\0\1\2\3\4\5\6\7\8\9\80";
    bar: "#{x}\0\1\2\3\4\5\6\7\8\9";
    bar: "x#{\0\1\2\3\4\5\6\7\8\9}";
    bar: "x\a\A\b\B\c\C\d\D\e\E\f\F\g\G";
    baz: "#{x}\a\A\b\B\c\C\d\D\e\E\f\F\g\G";
}

baz {
    // another regression
    baz: "#{x}\79#{\7a}";
}
```

Issue #1106, opened 2015-04-15
Empty declarations sometimes being output


xzyfer:
```scss
.foo {
  foo: unquote(null); }
```
Ruby Sass 3.4.12
```css
// no output
```
Ruby Sass 3.4.12
```css
DEPRECATION WARNING: Passing null, a non-string value, to unquote()
will be an error in future versions of Sass.
        on line 2 of test.scss
```
Libsass 3.1.0
```css
.foo { }
```
Libsass 3.2.0-beta.5
```css
.foo {
  foo: ;}
```

Spec added https://github.com/sass/sass-spec/pull/329

xzyfer:
Still seeing this in 3.0.0-beta.6. I'll work on a reduced test case today.

xzyfer:
```scss
$foo: red;
$bar: unquote(null);

.baz {
    foo: $foo;
    foo: $bar; }

.bax {
    foo: $bar; }

.bay {
    foo: $foo; }
```

Ruby Sass
```css
.baz {
  foo: red; }

.bay {
  foo: red; }
```

Libsass 3.1.0
```css
.baz {
  foo: red; }

.bax { }

.bay {
  foo: red; }
```

Libsass 3.2.0-beta.6
```css
.baz {
  foo: red;
  foo: ;}

.bay {
  foo: red; }
```

xzyfer:
This appears to be due a difference in `unquote`

```scss
foo { 
  bar: type-of(unquote(null)) }
```
Ruby Sass
```css
foo {
  bar: null; }
```
Libsass
```css
foo {
  bar: string; }
```

xzyfer:
Although the current patch solves the issue of not showing empty rulesets, due to https://github.com/sass/libsass/issues/1124 in the special case of `unquote(null)` this breaks if there another non-null declaration in the ruleset.

```scss
$foo: null;
foo {
  a: $foo;
  b: $foo; }
bar {
  a: b;
  b: unquote($foo); }
```
Ruby Sass
```css
bar {
  a: b; }
```
Libsass 3.1
```css
bar {
  a: b; }
```
Libsass 3.2.0-beta.6
```css
bar {
  a: b;
  b: ;}
```

Issue #1124, opened 2015-04-23
unquote(null) returns a string instead of null


xzyfer:
```scss
foo {
  a: type-of(null);
  b: type-of(unqoute(null)); }
```
Ruby Sass
```css
foo {
  a: null;
  b: null; }
```
Libsass
```css
foo {
  a: null;
  b: string; }
```

This in itself is not a regression, but due to other changes surrounding generating CSS output this causing regressions - https://github.com/sass/libsass/issues/1106#issuecomment-95375320

Specs added https://github.com/sass/sass-spec/pull/335

xzyfer:
Fixed by #1125

PR #1125, opened 2015-04-23
Unquoting null should return null

xzyfer:
This PR fixes `unquote(null)` ensuring it returns `null`.

Fixes https://github.com/sass/libsass/issues/1106, https://github.com/sass/libsass/issues/1124.
Specs added https://github.com/sass/sass-spec/pull/338, https://github.com/sass/sass-spec/pull/339.

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -760,6 +760,10 @@ namespace Sass {
     Signature unquote_sig = "unquote($string)";
     BUILT_IN(sass_unquote)
     {
+      if (dynamic_cast<Null*>(env["$string"])) {
+        return new (ctx.mem) Null(pstate);
+      }
+
       To_String to_string(&ctx);
       AST_Node* arg = env["$string"];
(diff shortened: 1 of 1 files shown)


functions.cpp
```diff
@@ -760,6 +760,10 @@ namespace Sass {
     Signature unquote_sig = "unquote($string)";
     BUILT_IN(sass_unquote)
     {
+      if (dynamic_cast<Null*>(env["$string"])) {
+        return new (ctx.mem) Null(pstate);
+      }
+
       To_String to_string(&ctx);
       AST_Node* arg = env["$string"];
```
mgreter: Would make more sense IMO to use to `arg` variable and avoid fetching it twice from env ... maybe you want to move the `To_String` inside the if cases that actually uses it ...

xzyfer: Legit - https://github.com/sass/libsass/pull/1129

Issue #1130, opened 2015-04-24
varargs with zero arguments


chriseppstein:
This code should compile:

```scss
@function foo($args...) {
  @return bar($args...);
}

@function bar() {
 @return "hi";
}

.foo {
  result: foo();
}
```

to:

```css
.foo {
  result: "hi";
}
```

But instead I get the following error even in the latest libsass code:

```
function bar only takes 0 arguments; given 1 on line 2 at column 11
```

chriseppstein:
damn you're fast.

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/343

PR #1131, opened 2015-04-24
Allow empty rest arguments in varargs

mgreter:
Fixes https://github.com/sass/libsass/issues/1130


xzyfer:
Add a spec then :ship:

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/343

xzyfer:
Confirmed this passes the spec :shipit:

Issue #1133, opened 2015-04-25
Regression in the call function with a map argument


xzyfer:
When `call` is passed a maps as the `$args` param it is being auto expanded. This should only happen if the map value is followed by `...` i.e. `call('foo', $my-map...)` vs `call('foo', $my-map)`. Originally reported https://github.com/sass/libsass/issues/1128#issuecomment-96031247

```scss
@function foo($map) {
  @return $map;
}

@debug(call(foo, (this: is, my: map));)
```
Ruby Sass + Libsass 3.1.0
```css
test.scss:20 DEBUG: (this: is, my: map)
```
Libsass 3.2.0-beta.6
```css
Error: function foo has no parameter named $this
        on line 20 of test.scss
>> @debug(call(foo, (this: is, my: map)));
   ------------------------------------^
```
Specs added https://github.com/sass/sass-spec/pull/342

PR #1135, opened 2015-04-25
Fix call function incorrectly expanding map arguments

xzyfer:
This PR fixes a bug with `call()` where in it incorrect treated a map argument as a rest argument.

Fixes https://github.com/sass/libsass/issues/1133.
Spec added sass/sass-spec#342, https://github.com/sass/sass-spec/pull/351.

### DIFF
--- a/bind.cpp
+++ b/bind.cpp
@@ -56,6 +56,20 @@ namespace Sass {
           else {
             env->local_frame()[p->name()] = a->value();
           }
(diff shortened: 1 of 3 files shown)


xzyfer:
@mgreter I owe you a beer for `debug_ast`!

xzyfer:
The failure in CI is due to a 404 during setup.

Issue #1080, opened 2015-04-09
@import placed above comment at beginning of output CSS


simonsmith:
Given this `scss`:

``` sass
/** Some comment */

@import url("suitcss-components-grid");

.Tweet {
  clear: both;
  color: red;
}
```

I expect it to compile to:

``` css
/** Some comment */
@import url("suitcss-components-grid");
.Tweet {
  clear: both;
  color: red;
}
```

Which it does with Ruby Sass 3.4.12.

However libass 3.1.0 places the `@import` above the comment:

``` css
@import url("suitcss-components-grid");
/** Some comment */
.Tweet {
  clear: both;
  color: red; }
```

I would expect it to match the Ruby Sass output. I realise it's perhaps a minor issue but in my situation it effects some tooling that expects the comment on the first line.

Above tests were run in sassmeister.com

Thank you

mgreter:
Ruby sass seems to preserve the order of appearance. Seems like we need to have a more generic `vector<Expression*> top_elements` instead of specific ones for each type (comment/import). Since this is not a regression I will assign it to 3.2.1.

```CSS
/** comment 1 */
@import url("import-1");
/** comment 2 */
@import url("import-2");
/** comment 3 */
foo { bar: baz; }
```

ruby sass:
```css
/** comment 1 */
@import url("import-1");
/** comment 2 */
@import url("import-2");
/** comment 3 */
foo {
  bar: baz;
}
```

libsass:
```CSS
@import url("import-1");
@import url("import-2");
/** comment 1 */
/** comment 2 */
/** comment 3 */
foo {
  bar: baz; }
```

simonsmith:
Great, appreciate it

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/321

xzyfer:
Fixed in [3.2.1](https://github.com/sass/libsass/releases/tag/3.2.1)

PR #1138, opened 2015-04-26
Fix order of top nodes (comments and imports)

mgreter:
Fixes https://github.com/sass/libsass/issues/1080

### DIFF
--- a/output.cpp
+++ b/output.cpp
@@ -8,8 +8,7 @@ namespace Sass {
   Output::Output(Context* ctx)
   : Inspect(Emitter(ctx)),
     charset(""),
-    top_imports(0),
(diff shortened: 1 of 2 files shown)


hcatlin:
Interestingly, the test for 1080 passes already... without this commit!

xzyfer:
Might be a sass-spec issue because the output is certainly incorrectly

```
$ sassc test.scss
/** comment 1 */
@import url("import-1");
@import url("import-2");
/** comment 2 */
/** comment 3 */
foo {
  bar: baz; }
```

hcatlin:
hmmm... what the hell is wrong with my setup...

PR #1152, opened 2015-04-30
Don't truncate trailing 0s on integers when precision is 0

### LINKED ISSUE #1140: variables losing zeros
Sorry for the weird title, I'm not sure how to even describe this properly.

given the input

```scss
$line-height: 20px; div {line-height: $line-height;}
```

I get the output

```css
div {
  line-height: 2px; }
```

note how `20px` turned into `2px`. I see the same behavior for `0.1px` turning into `px`.

---

I've tracked the issue down to `sass_option_set_precision(ctx_opt, 0);` being the culprit. Not setting a precision, or setting any precision other than `0` works. I *think* I remember reading something about `0` being the libsass default. But I can't support that by linking to code. Judging from [sass_context.cpp](https://github.com/sass/libsass/blob/3.2.0/sass_context.cpp#L549) and [inspect.cpp](https://github.com/sass/libsass/blob/3.2.0/inspect.cpp#L420) I'd assume the default is `5`.

I'm not sure if this is an issue with code, the documentation, or me being daft (again).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12312593-variables-losing-zeros?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xzyfer:
This PR fixes Libsass incorrectly truncation trailing `0`s on integers when precision is set to zero.

Fixes #1140.

### DIFF
--- a/inspect.cpp
+++ b/inspect.cpp
@@ -424,8 +424,11 @@ namespace Sass {
     // if after applying precsision, the value gets
     // truncated to zero, sass emits 0.0 instead of 0
     bool nonzero = n->value() != 0;
-    for (size_t i = d.length()-1; d[i] == '0'; --i) {
-      d.resize(d.length()-1);
+    size_t decimal = d.find('.');
+    if (decimal != string::npos) {
+      for (size_t i = d.length()-1; d[i] == '0' && i >= decimal; --i) {
+        d.resize(d.length()-1);
+      }
     }
     if (d[d.length()-1] == '.') d.resize(d.length()-1);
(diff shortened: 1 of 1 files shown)


xzyfer:
@mgreter thoughts?

PR #1157, opened 2015-05-01
Add a max stack depth setting

carsonmcdonald:
A segfault happens when a recursive function is created that never ends:

```
@function ex() { @return ex(); }
div { x: ex(); }
```

I've added a max stack depth setting and an error when that setting is exceeded.

### DIFF
--- a/constants.cpp
+++ b/constants.cpp
@@ -3,6 +3,8 @@
 namespace Sass {
   namespace Constants {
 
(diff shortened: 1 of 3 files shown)


xzyfer:
Thanks @carsonmcdonald

mgreter:
LGTM :+1:

Issue #1162, opened 2015-05-04
Segmentation fault on #{0/0}


altmind:
libsass/node-sass(tested with node-sass@3.0.0-beta.7) segfaults when trying to process following line:
`font: #{0/0} a;`
This strange notation is part of bootstrap mixins: bootstrap/lib/mixins/_hide-text.scss

Even if this syntax don't make sense, sass processor should not crash. This syntax worked in previous versions of node-sass(checked node-sass@2.0.1)

```shell
>npm install -g node-sass@3.0.0-beta.7
#...
>echo '.ya{font: #{0/0} a;}'|node-sass
Segmentation fault (core dumped)
```

mgreter:
The PR should solve the problem that libsass segfaults. But the test also shows another issue, since `#{0/0}` returns `Infinity` for libsass, while it stays untouched in ruby sass.

xzyfer:
Tracking the infinity issue in https://github.com/sass/libsass/issues/1206

PR #1165, opened 2015-05-04
Force evaluation of binary expression in interpolation

mgreter:
Fixes https://github.com/sass/libsass/issues/1162


xzyfer:
Add a spec?

xzyfer:
:+1:

Issue #1153, opened 2015-04-30
Incorrectly output integers as float is precision is 0


xzyfer:
Originally reported in https://github.com/sass/libsass/issues/1140#issuecomment-97699462
```scss
$foo: 123px; 
foo { 
  bar: $foo; }
```
Libsass with precision 0
```css
foo{
  bar: 123.0px; }
```
Ruby Sass with precision 0
```css
foo{
  bar: 123px; }
```

mgreter:
Once again, the logix seems to be more complicated then anticipated and again not sure if wanted behavior or just a ruby sass bug? //cc @xzyfer 

input.css
```css
foo { 
  bar: (0);
  bar: (.0);
  bar: (0.0);
  bar: (-0);
  bar: (-.0);
  bar: (-0.0);
  bar: (42);
  bar: (-42);
  bar: (42.0);
  bar: (-42.0);
  bar: (42.84);
  bar: (-42.84);
  bar: (42.24);
  bar: (-42.24);
  bar: (42.845555);
  bar: (-42.845555);
  bar: (42.245555);
  bar: (-42.245555);
}
```

Ruby sass output:
```css
foo {
  bar: 0;
  bar: 0;
  bar: 0;
  bar: 0;
  bar: 0;
  bar: 0;
  bar: 42;
  bar: -42;
  bar: 42;
  bar: -42;
  bar: 43.0;
  bar: -43.0;
  bar: 42.0;
  bar: -42.0;
  bar: 43.0;
  bar: -43.0;
  bar: 42.0;
  bar: -42.0; }
```

xzyfer:
> Seems to be a special case if the value was actually rounded

Yes. When I recently refactored this code I applied to "special cases" always because it was semantically equivalent and the old behaviour wasn't. That's what this issue is referring to.

PR #1166, opened 2015-05-05
Refactor number emitting (match precision of ruby sass)

mgreter:
Fixes https://github.com/sass/libsass/issues/1153

### DIFF
--- a/inspect.cpp
+++ b/inspect.cpp
@@ -416,43 +416,91 @@ namespace Sass {
 
   void Inspect::operator()(Number* n)
   {
+
+    string res;
+
+    // init stuff
     n->normalize();
+    int precision = 5;
+    double value = n->value();
+    // get option from optional context
+    if (ctx) precision = ctx->precision;
+
+    // check if the fractional part of the value equals to zero
+    // neat trick from http://stackoverflow.com/a/1521682/1550314
+    // double int_part; bool is_int = modf(value, &int_part) == 0.0;
+
+    // this all cannot be done with one run only, since fixed
+    // output differs from normal output and regular output
+    // can contain scientific notation which we do not want!
+
+    // first sample
     stringstream ss;
-    ss.precision(ctx ? ctx->precision : 5);
-    ss << fixed << n->value();
-    string d(ss.str());
-    // store if the value did not equal zero
-    // if after applying precsision, the value gets
-    // truncated to zero, sass emits 0.0 instead of 0
-    bool nonzero = n->value() != 0;
-    size_t decimal = d.find('.');
-    if (decimal != string::npos) {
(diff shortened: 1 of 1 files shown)


inspect.cpp
```diff
@@ -416,43 +416,91 @@ namespace Sass {
...
+      ss << fixed << value;
+    }
+
+    const char *it = ss.str().c_str();
+    while (*it && *it != '.' && *it != ',') ++ it;
+    const char *point = it;
+    while (*it && *it == '0') ++ it;
+    bool is_int = (*it == 0);
+    // take a shortcut for integers
+    if (is_int)
```
xzyfer: `bool is_int = d.find('.') == string::npos` ?

inspect.cpp
```diff
@@ -416,43 +416,91 @@ namespace Sass {
...
+
+    // check if we got scientific notation in result
+    if (ss.str().find_first_of("e") != string::npos) {
+      ss.clear(); ss.str(string());
+      ss.precision(max(12, precision));
+      ss << fixed << value;
+    }
+
+    const char *it = ss.str().c_str();
+    while (*it && *it != '.' && *it != ',') ++ it;
```
xzyfer: I find this code really unintuitive and hard to grok. We should aim for expressive code first, and optimize later.

```
bool has_decimal = ss.find_first_of(".,");
```

inspect.cpp
```diff
@@ -416,43 +416,91 @@ namespace Sass {
...
-    if (d == "-0") d.erase(0, 1);
-    // use fractional output if we had
-    // a value before it got truncated
-    if (d == "0" && nonzero) d = "0.0";
-    // if the precision is 0 sass cast
-    // casts to a float with precision 1
-    if (ctx->precision == 0) d += ".0";
-    // append number and unit
-    append_token(d + n->unit(), n);
+      error(res + " isn't a valid CSS value.", n->pstate());
```
xzyfer: Should this be in the output phase? Surely we know if a number in valid in eval?

mgreter: A number can eval to something that is not valid, like:

```
$a: 2px * 2px; // ->invalid
div { margin: $a / 2px; } // ->valid
```

inspect.cpp
mgreter: Unoftunately every bit (even the number 12) seems to be exactly needed this way.  But I guess this one loop can be rewritten with find_first_of ...

mgreter:
Refactored the code and incorporated the feedback from @xzyfer. Made the code much more explicit. I think I cannot really improve much on it anymore. Hope CI build passes this times. IMHO appveyor sometimes fails for one build (with random errors it seems) ...

xzyfer:
:+1:

Issue #1169, opened 2015-05-07
Quoted and unquoted colors in maps produces 'Duplicate key "transparent" in map'


AlbertoElias:
In Ruby Sass, CSS colors in map keys are considered as colours as you can see [here](http://sassmeister.com/gist/911c0887c1fa588737da) instead of as strings. 

It seems to be the same in [Libsass](http://sassmeister.com/gist/58f494fbfad75e32358f), but if I do something similar as the first example, [link](http://sassmeister.com/gist/dd293c1cb143867f4391), when they're set as map keys, they're both considered strings, but if you `map-get`, they are considered as colors if they're not quoted.

I also tried using the recently released node-sass 3.0.0 that uses the stable Libsass 3.2.0 and it still happens.

xzyfer:
Sorry, I'm confused. The example you've given fails on Ruby Sass 3.4.12.

```scss
$colors: (
	"transparent": #000000,
);

.c {
  color: map-get($colors, transparent);
  foo: type-of(map-get($colors, transparent));
}
```
Ruby Sass
```css
.c {
  foo: null; }
```
LibSass
```css
.c {
  foo: null; }
```

What behaviour are you expecting?

xzyfer:
```scss
$colors: (
	transparent: #000000,
	"transparent"": #000000,
);
```
Ruby
```css
Invalid CSS after "	"transparent"": expected ":", was "": #000000,"
```
LibSass
```css
error reading values after "transparent" on line 3 at column 2
```

xzyfer:
```scss
$colors: (
	transparent: #000000,
);

.c {
  color: map-get($colors, transparent);
  foo: type-of(map-get($colors, transparent));
}
```
Ruby
```css
.c {
  color: #000000;
  foo: color; }
```
LibSass
```css
.c {
  color: #000000;
  foo: color; }\
```

xzyfer:
As far as I can tell the behaviour is the same as Ruby Sass.

HugoGiraudel:
```scss
$map: (
  'red': 'foo',
   red: 'bar',
);

.foo {
  content: inspect($map);
}
```

Ruby Sass:

```scss
.foo {
  content: ("red": "foo", red: "bar");
}
```

LibSass:

```scss
Duplicate key "red" in map ("red": "bar"). on line 3 at column 14
```

Discussion: https://twitter.com/HugoGiraudel/status/596267826933014528

kaelig:
[...]

PR #1175, opened 2015-05-07
Add base eval step for map keys while parsing

mgreter:
Keys need to be immutable for the main structure, so we
need to evaluate them while the map is being parsed. We
may defer the map initialization into the eval step too.

Fixes https://github.com/sass/libsass/issues/1169

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -1010,6 +1010,16 @@ namespace Sass {
   {
     To_String to_string(&ctx);
     Expression* key = parse_list();
+    if (String_Quoted* str = dynamic_cast<String_Quoted*>(key)) {
+      if (!str->quote_mark() && !str->is_delayed()) {
+        if (ctx.names_to_colors.count(str->value())) {
+          Color* c = new (ctx.mem) Color(*ctx.names_to_colors[str->value()]);
+          c->pstate(str->pstate());
+          c->disp(str->value());
+          key = c;
+        }
+      }
+    }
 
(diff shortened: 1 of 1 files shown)


mgreter:
This fixes the issues mentioned in #1169, but there are more cases where we differ from ruby sass. IMO we need to refactor Sass_Value handling internally to streamline all related things. Currenty different parts (binary operations, inspection, equal operator, etc) are scattered around our code base. We should put that all under a streamlined API, but I guess this will not be ready before 3.3.

So this is a start, but as I said we have more issues like:
```css
$colors: (
    "transparent": "string",
    transparent: "color"
);

color { content: map-get($colors, transparent); }
string { content: map-get($colors, "transparent"); }
interpolate { content: map-get($colors, #{trans}#{parent}); }
```

ruby sass:
```css
color {
  content: "color"; }

string {
  content: "string"; }

interpolate {
  content: "string"; }
```
libsass:
```css
color {
  content: "color"; }

string {
  content: "string"; }

interpolate {
  content: "color"; }
```

So we seem to evaluate that string somewhere unnecessarily into a color again ...

xzyfer:
:+1:

AlbertoElias:
That was quick! Thanks :D

mgreter:
@AlbertoElias Can you confirm that the PR fixes your issues?

AlbertoElias:
@mgreter I can confirm it fixes it! Cheers!

Issue #1185, opened 2015-05-09
eval.cpp:531:81: error: 'to_string' was not declared in this scope


saper:
Using gcc 4.8.4 on FreeBSD 8.4

````
/bin/sh ./libtool  --tag=CXX   --mode=compile g++48 -DHAVE_CONFIG_H -I.    -Wall -O2  -std=c++0x -pipe -g -Wl,-rpath=/usr/local/lib/gcc48 -fno-strict-aliasing -Wl,-rpath=/usr/local/lib/gcc48 -MT eval.lo -MD -MP -MF .deps/eval.Tpo -c -o eval.lo eval.cpp
libtool: compile:  g++48 -DHAVE_CONFIG_H -I. -Wall -O2 -std=c++0x -pipe -g -Wl,-rpath=/usr/local/lib/gcc48 -fno-strict-aliasing -Wl,-rpath=/usr/local/lib/gcc48 -MT eval.lo -MD -MP -MF .deps/eval.Tpo -c eval.cpp  -fPIC -DPIC -o .libs/eval.o
eval.cpp: In member function 'virtual Sass::Expression* Sass::Eval::operator()(Sass::Function_Call*)':
eval.cpp:531:81: error: 'to_string' was not declared in this scope
         error("Stack depth exceeded max of " + to_string(Constants::MaxCallStack), c->pstate(), backtrace);
````

investigating...

saper:
This is due to 

````
#if ((__cplusplus >= 201103L) && defined(_GLIBCXX_USE_C99) \
     && !defined(_GLIBCXX_HAVE_BROKEN_VSWPRINTF))
````

in `<bits/basic_string.h>`

the same case is described here:

http://stackoverflow.com/questions/19500013/undefined-strtold-when-using-glibcxx-use-c99

xzyfer:
Closed by #1191

PR #1191, opened 2015-05-09
Don't use std::to_string (C++)

saper:
Use streams to convert unsigned long value


saper:
This fixes #1185 for certain gcc 4.x environments

Issue #1188, opened 2015-05-09
Is this error in compiler or not? "unterminated interpolant inside IE function"


mahnunchik:
Hi all,

I faced with the fact what [gridle](https://github.com/olivierbossel/gridle) library can not be compiled because it has following line of code:
```scss
*width: expression((this.parentNode.clientWidth/#{$context}*#{($columns / $name-multiplicator)} - parseInt(this.currentStyle['paddingLeft']) - parseInt(this.currentStyle['paddingRight'])) + 'px');
```

error:
```
{
  "message": "unterminated interpolant inside IE function expression((this.parentNode.clientWidth/#{$context}*#{($columns / $name-multiplicator)",
  "column": 11,
  "line": 244,
  "file": "/Users/nodejs/gridle/sass/gridle/_mixins.scss",
  "status": 1
}
```

```bash
$ node-sass -v
node-sass	3.0.0	(Wrapper)	[JavaScript]
libsass  	3.2.2	(Sass Compiler)	[C/C++]
```

Is this error in compiler or not?

Related issue in gridle repository: https://github.com/olivierbossel/gridle/issues/57

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/372

Issue #1102, opened 2015-04-14
Libsass errors on IE expression syntax


xzyfer:
```scss
foo {
  display:expression("inline",
    (this.innerHTML += (this.innerHTML.indexOf(",") == -1 ? ", " : "")),
    this.runtimeStyle.display = "inline");
}
```
Ruby Sass
```css
foo {
  display: expression("inline", (this.innerHTML += (this.innerHTML.indexOf(",") == -1 ? ", " : "")), this.runtimeStyle.display = "inline"); }
```
Libsass 3.0.0-beta.5
```css
Error: error reading values after 1
        on line 3 of test.scss
>> HTML += (this.innerHTML.indexOf(",") == -1 ? ", " : "")),
   ------------------------------------------^
```

xzyfer:
Related to https://github.com/sass/libsass/issues/690

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/344

am11:
:+1:

---

[Spoiler: Given we have custom functions, it might not be possible. On flip-side, it may be possible since it appears to me that the AST has all the required context to pull this]

IMO, at some point Sass codegens should be able to emit set of the standard CSS, v2 (and above), v3 (and above), v4 (and above) and since so forth. With those set, the compiler should error if the non-conforming pass-through CSS code is encountered. For instance, expression evaluation was ended by IE team in 2008: http://blogs.msdn.com/b/ie/archive/2008/10/16/ending-expressions.aspx and it is no longer supported in standard mode since IE8. Now in a team work, if someone used the expression evaluation, they should get the message if certain compile-time flag was set (say `--css2` or `--css3`). This is similar to how JS superset languages (sweet.js, coffeescript, typescript) are doing it. 

With that said, I think a black-list approach would be more appropriate to sniff out the vendor specific (non-prefix) tokens (such as `expression()`, `filter()` etc.) and throw.

Issue #931, opened 2015-03-10
Compilation warning using parentheses in alpha + opacity filter


jaydiablo:
Ran into an issue where our designer defined a mixin that caused libsass 3.1.0 to throw a warning and halt compilation.  Here's the mixin:

```
@mixin img-opacity($trans) {
          filter : alpha(opacity=($trans * 100));
      -ms-filter : "progid:DXImageTransform.Microsoft.Alpha(Opacity=#{$trans * 100})";
    -moz-opacity : $trans;
  -khtml-opacity : $trans;
         opacity : $trans;
}
```
libsass specifically complained about the "filter" definition with this message:
> error reading values after opacity

Sass 3.2.19 and up process the mixin just fine:

http://sassmeister.com/gist/e68c7a09b749dd1360ac
(you can change to libsass in that UI and see it fail compilation)

With a slight modification to the mixin, libsass handles it just fine:

```
@mixin img-opacity($trans) {
          filter : alpha(opacity=$trans * 100);
      -ms-filter : "progid:DXImageTransform.Microsoft.Alpha(Opacity=#{$trans * 100})";
    -moz-opacity : $trans;
  -khtml-opacity : $trans;
         opacity : $trans;
}
```
http://sassmeister.com/gist/f4261c967d48b9cd9a18

Seems like libsass doesn't like the parentheses there.

PR #1194, opened 2015-05-10
Improve ie property and expression parsing

mgreter:
Fixes https://github.com/sass/libsass/issues/1188
Fixes https://github.com/sass/libsass/issues/1102
Fixes https://github.com/sass/libsass/issues/931

### DIFF
--- a/util.cpp
+++ b/util.cpp
@@ -262,21 +262,15 @@ namespace Sass {
   {
     bool ws = false;
     bool esc = false;
-    char inside_str = 0;
     string text = "";
-    for(auto i : str) {
+    for(const char& i : str) {
       if (!esc && i == '\\') {
         esc = true;
+        ws = false;
         text += i;
       } else if (esc) {
         esc = false;
-        text += i;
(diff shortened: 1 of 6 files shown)


util.cpp
```diff
@@ -264,19 +264,23 @@ namespace Sass {
...
       if (!esc && i == '\\') {
         esc = true;
+        ws = false;
         text += i;
       } else if (esc) {
         esc = false;
+        ws = false;
         text += i;
-      } else if (!inside_str && (i == '"' || i == '\'')) {
+      } else if (false && !inside_str && (i == '"' || i == '\'')) {
```
xzyfer: Remove this branch?

xzyfer:
Spec PR https://github.com/sass/sass-spec/pull/375

util.cpp
mgreter: Good catch, seems valid, can be gone ...

xzyfer:
:+1:

Also I wonder if parsing these ie expressions is worth the effort (and perf hit?). Could we get away with going back to treating these as String_Schemas?

Issue #1187, opened 2015-05-09
Weird behaviour with lists as map keys


HugoGiraudel:
Maybe follow up of https://github.com/sass/libsass/issues/1169?

```scss
$a: 'foo';
$b: 'foo';
$map: (
  $a: 1,
  $b: 2
);

.foo {
  content: $a == $b;
  content: inspect($map);
}
```

Ruby Sass:

```scss
Duplicate key "foo" in map ($a: 1, $b: 2).
```

LibSass:

```scss
.foo {
  content: true;
  content: ("foo": 2);
}
```

The behaviour is really weird because obviously LibSass spots that it is twice the same key since there is only one key in the output. Although, it does not throw a duplicated key error.

Interestingly enough:

```scss
$key: 'foo';
$map: (
  $key: 1,
  $key: 2
);

.foo {
  content: $a == $b;
  content: inspect($map);
}
```

This yields in LibSass:

```scss
Duplicate key "$key" in map ($key: 2). on line 4 at column 10
```

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/373

HugoGiraudel:
You guys rock.

PR #1195, opened 2015-05-10
Fix duplicate key handling for maps

mgreter:
Deferred check from parser to evaluation
Fixes https://github.com/sass/libsass/issues/1187

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -1014,7 +1014,7 @@ namespace Sass {
 
   Expression* Parser::parse_map()
   {
-    To_String to_string(&ctx);
+    ParserState opstate = pstate;
     Expression* key = parse_list();
     if (String_Quoted* str = dynamic_cast<String_Quoted*>(key)) {
       if (!str->quote_mark() && !str->is_delayed()) {
@@ -1035,7 +1035,7 @@ namespace Sass {
 
     Expression* value = parse_space_list();
 
-    Map* map = new (ctx.mem) Map(pstate, 1);
+    Map* map = new (ctx.mem) Map(opstate, 1);
     (*map) << make_pair(key, value);
 
     while (lex_css< exactly<','> >())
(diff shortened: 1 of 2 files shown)


xzyfer:
:+1:

parser.cpp
```diff
@@ -1073,8 +1073,15 @@ namespace Sass {
       (*map) << make_pair(key, value);
     }
 
-    if (map->has_duplicate_key())
-    { error("Duplicate key \"" + map->get_duplicate_key()->perform(&to_string) + "\" in map " + map->perform(&to_string) + ".", pstate); }
+    // Check was moved to eval step
```
xzyfer: Remove these comments?

parser.cpp
```diff
@@ -1073,8 +1073,15 @@ namespace Sass {
...
+    // if (map->has_duplicate_key()) {
+    //   To_String to_string(&ctx);
+    //   error("Duplicate key \"" + map->get_duplicate_key()->perform(&to_string) + "\" in map " + map->perform(&to_string) + ".", pstate);
+    // }
+
+    ParserState ps = map->pstate();
+    ps.offset = pstate - ps + pstate.offset;
+    map->pstate(ps);
```
xzyfer: Can you quickly explain this change?

mgreter: The check for duplicates keys was moved to the eval stage. The reason for this is, ie. when we have variables as keys, we would check like `$a` == `$b` for duplicate keys. But we actually need to check it only after the keys have been evaluated ... I was not sure if we really can get rid of the first check, but it seems so ...

xzyfer: I meant the pstate stuff. Not sure what's happening here.

This check cab definitely go. I knew it become an issue when I originally
added it.

mgreter:
Ah, the pstate stuff makes sure the Map object source map is pointing from parenthese to parenthese, since we don't lex it with one call ...

xzyfer:
For some more context on the original error message. It was added to address the issue that we weren't throwing duplicate key issues, I expected we'd eventually need it in eval as well. It was added here to match the Ruby behaviour.

IIRC Ruby validates duplicate keys when adding keys to a Map AST node. However it only validates static values. The ideal way to handle this would be to have the Map AST either throw the error or have it return an `int > 0` on failure (only for String_Constant and String_Quoted).

xzyfer:
Ah right, it's because the close `)` is lexed by `parse_value` ?

xzyfer:
Spec activated https://github.com/sass/sass-spec/pull/656

Issue #1096, opened 2015-04-11
Permit multiline in URL


am11:
Test:

```scss
@import url("a\
b");
```

LibSass master (0dd65439366bd4b6d9fce42a7f6a45990df5e836) errors:

> URI is missing ')'

LibSass 3.1 outputs:

```css
@import url("a\
b");
```

RubySass v3.4.12 outputs:

```css
@import url("ab");
```

---

Related test:

```scss
// without quotes
@import url(a\
b);
```

RubySass throws the following error:

> Invalid CSS after "`@import` url(a": expected ")", was "\"

LibSass master (0dd65439366bd4b6d9fce42a7f6a45990df5e836) errors:

> URI is missing ')'

LibSass 3.1 outputs:

```css
@import url(a\
b);
```

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/345

am11:
@mgreter, there is still one crazy case which is inconsistent (perhaps not worth fixing?).

(added periods to highlight spaces)

```scss
@import url("a.. .\.. .
 . . b");
```

Ruby Sass:

```css
@import url("a.. ... .\a  . . b");
```

Current libsass master (28ee0884243820a34c3458e26b36fc59ccefd9b3):

```css
@import url("a.. ... .
\a  . . b");
```

Not sure about logic behind ruby-sass implementation. I was expecting something like this:

```css
@import url("a.. ... .\  . . b");
```

xzyfer:
This still fails the spec.

xzyfer:
Apologies, it looks like my local sass-spec is having issues.

xzyfer:
This fix was a false positive due to https://github.com/sass/sass-spec/pull/380. The spec still fails.

xzyfer:
Related https://github.com/sass/libsass/issues/1294

am11:
@xzyfer,

quickly tested current master with the following input:

```scss
/* url-test1: without quotes */
@import url(a
b);

/* url-test1: with quotes */
@import url("a
b");

/* url-test2: with quotes */
@import url("a   \d \\d \\\d \\\\d
    c \t \n \\t \\n \\\t \\\n \\\\t \\\\n");
  
/* url-test2: without quotes */
@import url(a   \d \\d \\\d \\\\d
    c \t \n \\t \\n \\\t \\\n \\\\t \\\\n);
```

LibSass:

```css
/* url-test1: without quotes */
@import url(a b);
[...]

PR #1196, opened 2015-05-10
Improve url parsing (allow line-feeds)

mgreter:
Fixes https://github.com/sass/libsass/issues/1096

### DIFF
--- a/prelexer.cpp
+++ b/prelexer.cpp
@@ -148,7 +148,12 @@ namespace Sass {
         exactly <'\''>,
         zero_plus <
           alternatives <
-            // skip all escaped chars first
+            // skip escapes
+            sequence <
+              exactly < '\\' >,
+              exactly < '\r' >,
+              exactly < '\n' >
+            >,
             escape_seq,
             // skip interpolants
             interpolant,
@@ -167,7 +172,12 @@ namespace Sass {
         exactly <'"'>,
         zero_plus <
           alternatives <
-            // skip all escaped chars first
+            // skip escapes
+            sequence <
+              exactly < '\\' >,
+              exactly < '\r' >,
+              exactly < '\n' >
+            >,
             escape_seq,
             // skip interpolants
             interpolant,
--- a/util.hpp
+++ b/util.hpp
@@ -21,7 +21,7 @@ namespace Sass {
   string comment_to_string(const string& text);
   string normalize_wspace(const string& str);
 
-  string quote(const string&, char q = 0);
+  string quote(const string&, char q = 0, bool keep_linefeed_whitespace = false);
   string unquote(const string&, char* q = 0);
   char detect_best_quotemark(const char* s, char qm = '"');
 
--- a/inspect.cpp
+++ b/inspect.cpp
@@ -641,7 +641,7 @@ namespace Sass {
   void Inspect::operator()(String_Quoted* s)
   {
     if (s->quote_mark()) {
-      append_token(quote(s->value(), s->quote_mark()), s);
+      append_token(quote(s->value(), s->quote_mark(), true), s);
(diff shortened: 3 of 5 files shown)


prelexer.cpp
```diff
@@ -434,6 +434,9 @@ namespace Sass {
           alternatives <
             alnum,
             exactly <'/'>,
+            exactly <'\\'>,
```
xzyfer: Should this be a sequence?

```
sequence< exactly <'\\'>, alternatives< exactly <'\r'>, exactly <'\n'> > >
```

mgreter: Do you really want to be that strict? Seems to work fine and not sure how ruby sass really does it ... this way it also accepts certain escape sequences (which I guess is correct). Feel free to create a more specific spec test ;)

xzyfer: As is stands I believe this would allow cases Ruby Sass does not, and that aren't in the CSS Spec (which is bad)

```scss
@import url("foo\
bar");
@import url("foo
bar");
@import url(foo
bar);
//Errors with Ruby Sass
@import url(foo\
bar);
```
Ruby Sass
```css
@import url("foobar");
@import url("foo\a bar");
@import url(foo bar);
```
With this patch
```css
@import url("foobar");
@import url("foo\abar");
@import url(foo
bar);
@import url(foobar);
```

xzyfer: Also, it looks like Ruby Sass treats the new line in the quoted string as a space delimited list.

mgreter: OK, I see you're point!

util.hpp
```diff
@@ -21,7 +21,7 @@ namespace Sass {
   string comment_to_string(const string& text);
   string normalize_wspace(const string& str);
 
-  string quote(const string&, char q = 0);
+  string quote(const string&, char q = 0, bool flag = false);
```
xzyfer: What does `flag` mean? Is there a better name that will describe what this is for?

xzyfer:
Update the specs https://github.com/sass/sass-spec/pull/374

util.hpp
mgreter: Didn't yet come around to look more into it. But it's the reason why I added WIP again. Again I don't really know if we need to move the code part that uses this flag to another function. Will have another look after work :smile:

mgreter:
I renamed the flag to `keep_linefeed_whitespace`. I have no idea why ruby sass does what it does here. So I really can only describe what the flag is actually doing. It seems to be some discrepancy between interpolation, quoted string, strings to css or something else. And I'm out of clues and guesses.

xzyfer:
Are just handling `url` values poorly? The Ruby Sass behaviour seems intuitive to me. 

If the `url` value is a quoted string and the newline is escaped then it's removed.
If the `url` value is not a quoted string it's a Sass list.

xzyfer:
I noticed another thing

```scss
@import url("foo
bar");
```
Produces a deprecation warning in Ruby Sass
```
DEPRECATION WARNING on line 2, column 13 of test.scss:
Unescaped multiline strings are deprecated and will be removed in a future version of Sass.
To include a newline in a string, use "\a" or "\a " as in CSS.

@import url("foo\a bar");
```
But not Libsass
```css
@import url("foo\a bar");
```

Libsass does however produce the deprecation warning if there is more content after the import (notice the extra newline)

```scss
@import url("foo
bar");

```

xzyfer:
Given this new information it would seem I was incorrect in thinking Ruby turned multi-line strings into lists. It appears to native support multi-line strings but normalises the newline to a space in the output phase.

am11:
:tada: 
Thanks @mgreter! ~,~

PR #1197, opened 2015-05-10
Improve selectors parsing for nth-child binomials

mgreter:
Fixes https://github.com/sass/libsass/issues/593

Implementation may not be the cleanest, but maybe we find more use cases where "compress" can be applied to String Constants. Then we may get a better picture how we can implement it better.

### DIFF
--- a/ast.hpp
+++ b/ast.hpp
@@ -1359,21 +1359,22 @@ namespace Sass {
   ////////////////////////////////////////////////////////
   class String_Constant : public String {
     ADD_PROPERTY(char, quote_mark);
+    ADD_PROPERTY(bool, can_compress_whitespace);
     ADD_PROPERTY(string, value);
   protected:
     size_t hash_;
   public:
     String_Constant(ParserState pstate, string val)
-    : String(pstate), quote_mark_(0), value_(read_css_string(val)), hash_(0)
+    : String(pstate), quote_mark_(0), can_compress_whitespace_(false), value_(read_css_string(val)), hash_(0)
     { }
     String_Constant(ParserState pstate, const char* beg)
-    : String(pstate), quote_mark_(0), value_(read_css_string(string(beg))), hash_(0)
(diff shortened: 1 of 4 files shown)


xzyfer:
My issue with this is that it's unclear when/if compressed should be used. It's another undocumented mystery flag that noone understands. 

IMHO this issues isn't of any importance atm so we can defer it until 3.4 with the rest of the whitespace issues. By then we should have a better idea of if there are more usecases for compressed (I don't think there are from reading the Ruby code).

mgreter:
I somewhat agree, but I can explain quite clearly what that flag does. It removes all whitespace from the string if it is emitted to css in compressed mode. This is not really familiar with any other node. So not sure what to do with it. I thought about a specific ast node for binomials, but I still would need to preserve the full parsed string, since ruby sass really preserves the whitespace (it does not just add one optional space as other nodes). Or it would basically be another string type doing exactly what the new "compress" flag is doing ...

I somewhat disagree to postpone this. The result seems promising to be on par with ruby sass. I tend to search the code for these specific flags to see where they are used. So far this flag is really specific to binomials. As you see I pretty much fixed all other white space issues, so we probably can label the next bugfix release a "white-space" release, so I vote to include this.

Of course I'm always open to have the option named differently and/or add some description!

xzyfer:
The concerns here are too mixed up IMO. Fixing an issue with binomials-like expressions in selectors by changing how we output strings?

At the very least the output logic should be scoped to selectors, at which point we could use context, rather then some mystery flag. This at least signals to the reader what the intent is.

xzyfer:
Even renaming the `flag` to signal that it's about selectors, would be a big improvement.

mgreter:
Renamed the flag to `can_compress_whitespace`. I don't think it case much to do with selectors, I has specifically to do with those binomials. For me it even looks somewhat like a bug in compressed mode, which this flag reproduces (so we match ruby sass). But that is really just me guessing ...

xzyfer:
This is an improvement :+1:

Issue #442, opened 2014-07-12
Interpolation Issue - Combining Values [$15]


grayghostvisuals:
I have a user on my project that [filed an issue in regards to interpolation using libsass w/gulpsass](https://github.com/typeplate/starter-kit/issues/23) so I'm curious if this is a bug within libsass' interpolation as the current method I use works just fine according to the docs and compiling with Ruby.

Here's the code that seems to be the problem…

```scss
@function context-calc($scale, $base, $value) {
  @return ($scale/$base) + $value; // was @return ($scale/$base)#{$value};
}
@function measure-margin($scale, $measure, $value) {
  @return ($measure/$scale) + $value; // was @return ($measure/$scale)#{$value};
}
```

As you can see the commented lines were the original lines that worked just fine when compiling with Ruby otherwise libsass was giving this ``font-size: 32 rem`` as a result.

HugoGiraudel:
Adding to the discussion: appending a unit as a string is a very poor way of converting a unitless number to a length since it results in having a `string` instead of a `number`, preventing any further calculation.

mgreter:
IMHO this is a pretty difficult issue since it involes `parse_list`, which we currently always expect to be at least space separated. I have added a `debugger.hpp` file which contains a `debug_ast` function:

input:
```
foo { bar: (10/10)#{rem}; }
```

output:
```
foo { bar: 1 rem; }
```

debug_ast:
```
Block 0x2630200 0
 Ruleset 0x22652c0 0
  Selector_List 0x26302e0 - -
   Complex_Selector 0x26e6300 - - -> { } <> X < >
    Compound_Selector 0x23993c0 - - <> X < >
     Selector_Reference 0x2991a20 @ref 0
   -Complex_Selector 0x26e61a0 - - -> { } <> X <>
   - Compound_Selector 0x23992d0 - - <> X <>
   -  Type_Selector 0x2991980 <<foo>> - <> X < >
  Declaration 0x2b27dc0 0
   prop: String_Quoted : 0x26e63b0 [bar] < > X <>
   value: List 0x2b27cf0 [Space] [delayed: 0]  [interpolant: 0]
   value:  Binary_Expression 0x26e65c0 [11]
   value:   left:  Textual  [NUMBER]0x26e6460 [10]
   value:   right: Textual  [NUMBER]0x26e6510 [10]
   value:  String_Schema 0x2b30d90 4 {has_interpolants}
[...]

PR #1198, opened 2015-05-11
Fix interpolation edge case in value parsing

mgreter:
Fixes https://github.com/sass/libsass/issues/442 (probably depends on other PR, so let see once they are merged if CI will be happy).


xzyfer:
This would be rad.

xzyfer:
Rebase and :ship:

PR #1064, opened 2015-04-07
Selector functions

onedayitwillmake:
**Please note: The code in this pull request was written by myself, and is does not represent my employer.** 

*updated 4/08/15*
*updated 4/09/15*

I rebased the PR https://github.com/sass/libsass/pull/1019 and removed Xcode references.

I have 8 of the 8 built, I started from the bottom for some reason.
Bold means they are in and passing the test. 

I'm still actively working on this - however being my second PR and new to the library I wanted to make sure anyone could catch any silly things I was doing.

- [x] selector-nest($selectors...)
- [x] selector-append($selectors...)
- [x] selector-extend($selector, $extendee, $extender)
- [x] selector-replace($selector, $original, $replacement)
- [x] selector-parse($selector)
- [x] simple-selectors($selector)
- [x] is-superselector($super, $sub)
- [x] selector-unify($selector1, $selector2)

Some notes:
* Added stub functions
* Passing 'Contextualize' to `Native_Function signature in BUILT_IN(name) Expression*
* Added `ARGSELL` to compliment `ARG`, `ARGR`, `ARGM` since all the selector functions rely on that so heavily 
* Had to do some hairy things to get `Selector_List::selector_unify` and `Selector_List::selector_unify` to work
* Added `public static` methods version of Extend's subweave, *for now* I named it StaticSubweave just to make sure I don't step on any toes, will remove floating one if that seems fine to everyone
* Added a naiveTrim function to Node that removes any duplicate selectors (this is for `Complex_Selector::unify_with` was returning correct results but the final shared selector would sometimes appear twice, and blindly adding everything else in the collection (I could not get node's existing `trim` implementation to work - I Also could not find any references to it be used.
* Added a `public static` version of Extend's extendSelectorList
* Added param `isReplace` to `Extend::extendSelectorList` - which does not add the Selector if an extension takes place. 
* Added `Vectorized<T>::first` to compliment  `Vectorized<T>::back` in order to make `selector_append` code more clear,

Issue #1091, opened 2015-04-10
Deep @extend specificity problems - continuation


mzgol:
Refs #592

The following SCSS:
```scss
.a {
    top: 0;
}

.b {
    .c {
        @extend .a;
    }
}

.d {
    > .e {
        @extend .a;
        @extend .c;
    }
}
```
is converted by Ruby Sass 3.4.13 to:
```css
.a, .b .c, .b .d > .e, .d > .e {
  top: 0; }
```
libsass 3.2.0-beta.5 produces:
```css
.a, .b .c, .d > .e {
  top: 0; }
```

EDIT: I simplified the test case.

mzgol:
Possibly related: #823

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/322

Issue #1063, opened 2015-04-07
Extending a %placeholder containing a parent reference produce incorrect output


timschoch:
Hello
I'm not sure if this problem is new or related to a known one. & Selectors within a placeholder get ignored when the parent selector is generated on the fly. This is a simplyfied version of what we use to generate navigations. (eg. first level horizontally, second megamenu, third vertically, ...)

```sass
%submenu {
  &:hover > ul { display: block; }
}

ul.root {
  $nested: unquote( '' );
  @for $level from 1 through 3 {
    @if length( $nested ) > 1 {
      #{$nested} {
        > li { @extend %submenu; } 
      }
    } @else {
      > li { @extend %submenu; }
    }
    $nested: append( $nested, unquote( ' > li > ul' ) );
  }
}
```
This is what libsass renders
```css
ul.root > li:hover > ul {
  display: block;
}
```
And in comparission the sass output as I would expect it
```css
ul.root > li:hover > ul,
ul.root > li > ul > li:hover > ul,
ul.root > li > ul > li > ul > li:hover > ul {
  display: block;
}
```
Interestingly the behaviour changes when we add a class inside the loop
```scss
@for $level from 1 through 3 {
  @if length( $nested ) > 1 {
    #{$nested}.level#{$level} { // <- added level specific class
      > li { @extend %submenu; } 
     }
  } @else {
    > li { @extend %submenu; }
  }
   $nested: append( $nested, unquote( ' > li > ul' ) );
}

// renderes as expected
ul.root > li:hover > ul,
ul.root > li > ul.level2 > li:hover > ul,
ul.root > li > ul > li > ul.level3 > li:hover > ul {
  display: block;
}
```

And it's only the & Selector thats broken. If I add another selector to the placeholder the new one works.
```scss
%submenu {
  > a { color: blue; }
  &:hover > ul { display: block; }
}

// > a works
ul.root > li > a,
ul.root > li > ul > li > a,
ul.root > li > ul > li > ul > li > a {
  color: blue;
}
// &:hover doensn't
ul.root > li:hover > ul {
  display: block;
}
```

xzyfer:
I can confirm this issue is present in Libsass 3.2.0-beta.5

mgreter:
You can test for regressions with http://sassmeister.com/ (this one is a plain bug)

xzyfer:
I've looked into this and it appears the issue only occurs when the placeholder's selector contains a `:pseudo`.

[...]

Issue #963, opened 2015-03-18
Implement selector functions


xzyfer:
Ruby Sass 3.4 introduced a bunch of selector functions. With #548 pretty much done we're in a good spot to tackle this.

>Complementing the ability to use `&` in SassScript, there’s a new suite of functions that use Sass’s powerful `@extend` infrastructure to allow users to manipulate selectors. These functions take selectors in the fully-parsed format returned by `&`, plain strings, or anything in between. Those that return selectors return them in the same format as `&`.
>
- The [selector-nest($selectors...)](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_nest-instance_method) function nests each selector within one another just like they would be nested in the stylesheet if you wrote them separated by spaces. For example, `selector-nest(".foo, .bar", ".baz") returns .foo .baz, .bar .baz (well, technically (("foo" ".baz") (".bar" ".baz")))`.
- The [selector-append($selectors...)](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_append-instance_method) function concatenates each selector without a space. It handles selectors with commas gracefully, so it’s safer than just concatenating the selectors using `#{}`. For example, `selector-append(".foo, .bar", "-suffix")` returns `.foo-suffix, .bar-suffix`.
- The [selector-extend($selector, $extendee, $extender)](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_extend-instance_method) function works just like `@extend`. It adds new selectors to `$selector` as though you’d written `$extender { @extend $extendee; }`.
- The [selector-replace($selector, $original, $replacement)](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_replace-instance_method) function replaces all instances of `$original` in $selector with `$replacement`. It uses the same logic as `@extend`, so you can replace complex selectors with confidence.
- The [selector-unify($selector1, $selector2)](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_unify-instance_method) function returns a selector that matches only elements matched by both `$selector1` and `$selector2`.
- The [is-superselector($super, $sub)](http://sass-lang.com/documentation/Sass/Script/Functions.html#is_superselector-instance_method) function returns whether or not $super matches a superset of the elements `$sub` matches.
[...]

PR #1202, opened 2015-05-12
Implement `is_superselector` sass function

mgreter:
Based on https://github.com/sass/libsass/pull/1064

This is mainly for me to try out is superselector behavior which is one root problem of our selector deduping issues (which covers pretty much all of the currently still open confirmed bug issues).
It may also be a base to get the work done by @onedayitwillmake merged into master for 3.3.

### DIFF
--- a/ast.cpp
+++ b/ast.cpp
@@ -242,7 +242,23 @@ namespace Sass {
     if (!lbase) // no lbase; just see if the left-hand qualifiers are a subset of the right-hand selector
     {
       for (size_t i = 0, L = length(); i < L; ++i)
-      { lset.insert((*this)[i]->perform(&to_string)); }
+      {
+        Selector* lhs = (*this)[i];
+        if (Wrapped_Selector* wrapped = dynamic_cast<Wrapped_Selector*>(lhs)) {
+          if (
+            wrapped->name() == ":matches(" ||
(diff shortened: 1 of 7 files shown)


mgreter:
By sheer accident I seem to have fixed #823, #1063 and #1091 :grinning: 
Plus we pass a lot more of the `is_superselector` spec tests added in #963

IMO we should be able to include the improvements for 3.2.4, since they are clear bugfixes.
@xzyfer would you want to hold off the `is_superselector` function for 3.3!? I would ship it.

xzyfer:
If you think it's ready we can ship it now. I'm keen to tag 3.2.4 in the
next hour or so.

mgreter:
OK, just a few moments ... let me clean up some white-space and comments ...

mgreter:
Also let's wait for CI here, then activate all spec tests and re-start the CI to check again ...

xzyfer:
Great. I think there are some specs that need to be activated from the
recent PRs as well.

mgreter:
As always, I just create a big activate patch for the next release.
https://github.com/sass/sass-spec/pull/379

mgreter:
Merged spec activation patch and restarting CI rebuild now ...

Issue #1210, opened 2015-05-14
@at-root and #{&} produces incorrect output [$10]


xzyfer:
```scss
@mixin box-sizing-border {
  @at-root {
    #{&} {
      *, *:before, *:after { 
        box-sizing: inherit; }
    } } }

foo {
  @include box-sizing-border; }
```
Ruby Sass
```css
foo *, foo *:before, foo *:after {
  box-sizing: inherit; }
```
Libsass 3.2.4
```css
foo foo *, foo foo *:before, foo foo *:after {
  box-sizing: inherit; }
```

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/382

iparr:
Also struggling a little with this issue.

I've been using it (perhaps unwisely) *without* mixins, for example in this example where the `global-logo` will only adapt a hover color when it is a link (as opposed to a heading, or span):

```
.global-logo {

	@at-root a#{&} {
		&:hover {
			background-color: red;
		}
	}
}
```

This is the one thing stopping me utilising libsass fully on some of my websites so I'm going to watch keenly for a fix!

blackfalcon:
Regardless of use with mixin or not, the issue persists when trying to create a combination selector based off the parent in the Sass using a child selector. In this example, the desired output from rubySass is not the same as libSass.

```scss
.parent {
  width: 100%;
  @at-root .child#{&} {
    width: 50%;
  }
}
```

**rubySass**
```
.parent {
  width: 100%;
}
.child.parent {
  width: 50%;
}
```

**libSass**
```
.parent {
  width: 100%;
}

.parent .child.parent {
  width: 50%;
}
```

Give it a run in [SassMeister](http://sassmeister.com/gist/30b476b8d88c5be1413f) to see.

danielguillan:
I've run some quick tests and It looks like a bug was introduced in 3.2.3 or 3.2.4. `@at-root` was working as expected on 3.2.0 through 3.2.2. I hope this helps.

xzyfer:
Base on @danielguillan new information I dug in a big further and found this is a regression introduced in https://github.com/sass/libsass/commit/13adf2c in #1183.

I'll see about getting a patch for this into 3.2.5 since there's significant demand for it.

xzyfer:
I've had a quick look into this and unfortunately there are a lot of edge cases. We need significantly more coverage in sass-spec before we're able to tackle this.

danielguillan:
[...]

PR #1217, opened 2015-05-15
Refactor contextualize into eval [WIP]

mgreter:
Remove contextualize and contextualize_eval by placing the functionallity into eval. Also refactors usage of Parent_Selector and gets rid of Selector_Reference. We now pretty much only store state on the expander itself (selector_stack, env_stack, etc). First iteration that should not break CI.


xzyfer:
Needs a clean up but generally speaking it looks sane

xzyfer:
Does your parent selector refactor help with https://github.com/sass/libsass/issues/1210 by any chance?

mgreter:
Well, it should help to make more sense out of the actual code. I actually did not understand why it was doing what it was doing. So yes, it could help with that, since it is more obvious where and which parent selector we are referring to IMHO. But it really depends on the underlying problem :)

mgreter:
Took a look at your linked issue and I guess I know how we can fix it. IMHO we need another selector type beside `Parent_Selector`, namely `Root_Selector`. Currently I thought I could just use NULL to represent that, but this would need quite some refactoring, since we have a lot of places where we would need to check for that potential NULL pointer. Parent selectors now only "live" in the selector_stack, so you can more easily tamper with it. I used this technique to implement Parent_Selector rendering (pop one selector off the stack, render recursively and push it back once finished). So in theory we just should need to add a `Root_Selector` to that stack when rendering/expanding root nodes. At least that's what I would guess. I'm currently focusing in another area, which will take quite some time (values/nodes and memory management, plus some basic performance considerations; currently catching up with the latest tricks in mark and sweep GC implementations). So I have absolutely no ETA if/when I will come around to implement this, so I thought I document it here for future references.

xzyfer:
Can this be applied without https://github.com/sass/libsass/pull/1249?

It looks like https://github.com/sass/libsass/pull/1249 does a bunch of this work as well.

Issue #1108, opened 2015-04-15
Serializable libsass context?


saper:
This is more a question than an issue. I know there is a [HAML forum on Google Groups](https://groups.google.com/forum/#!forum/haml) but I couldn't find any mailing list specific to `libsass`.

I am looking for ways to pass libsass context structure between threads/processes - in other words to serialize its contents.

Motivation:

I am considering to implement a kind of co-routine interface to libsass. In general it would work like this:

```
    C -> L  "compile_file" (context)
    C <- L   "call custom importer"
    L -> C   "import this"
    C <- L   "call custom function"
    L -> C   "custom function result"
    C <- L   "compile_file result"
````

Of course this is not possible with the [current API](API-Documentation) as it is right now, but
maybe it could be available in libsass 4.0

Such an interface could have a couple of advantages: 
- detaching the parser from the environment like files,
- implementation of generic functions would be easy (as described by @hcatlin [in the API discussion in issue 631](https://github.com/sass/libsass/issues/631#issuecomment-62288170)).
- ability to provide LibSassaaS :)

Serialized context implementation:

Because we don't offer such interface right now to implement coroutine interface as above I would need to pass context between threads. Some libraries offer public interface to their API structures, so that they can be processed by the client. I've seen some libraries that do not offer structure details, but provide an extra function that just determines structure size (so that the structure can be copied around transparently). Given we have lots of (char *) pointers it wouldn't work probably as-is.

For now I will probably need to serialize data myself and then reconstruct the structure on the `L` side of my connection as I don't want to depend on libsass internals.

I am thinking of a more low-level C-style wire format (struct, length, char[], length, char[], ...), but providing JSON encapsulation could of course be an advantage.

What would be the best way to implement this? I would like to prepare something that lives outside of libsass right now, but maybe the code could be included in libsass in the future.

mgreter:
I created a [collective ticket] [1] to keep track this and other "non-urgent" feature requests, in order to keep the issue tracker a bit more clean for the more 1st aid issues. I would like to encourage everybody interested in this particular feature to add their comments into this closed issue anyway.

Thanks for your understanding and your contribution!

[1]: https://github.com/sass/libsass/issues/1892

Issue #1065, opened 2015-04-07
Expose way to access AST of Sass file


Snugug:
I'm in the process of writing a Sass linter along with @xzyfer (and have a few ideas about some really funky importers that only will import parts of a file, say ignoring selectors or only importing mixins, etc…) and have found a need to have the AST for Sass files available. We've been looking at [Gonzales](https://github.com/tonyganch/gonzales-pe/tree/dev) but realize it would probably be more accurate and future-friendly if the AST was produced by the same thing doing the actual compiling.

With this in mind, I'd like a way for Libsass to expose its AST to developers instead of just compiling CSS files from the AST.

(unrelated to #430)

xzyfer:
Hey @snugug I rules out using the Libsass AST because it's significantly
inadequate for the task of linting. For example we don't keep track of
whitespace at all.

Snugug:
Well maybe it should 😝 

I still think this would be useful. As a side, I think I have some time this weekend to work on this again

mgreter:
@xzyfer each ast node has a parserstate which keeps track of leading whitespace ...

xzyfer:
I don't see a reason not to do this, but it would take significant effort on top of it to address the problem @snugug wants to solve :)

I've started work on a pure js scss parser which will better suit our use case :)

mgreter:
@xzyfer not that complicated IMHO. We have a capable json library at hand to generate some versatile output format. Other than that it is basically just another [`debug_ast`] [2], just to output everything into a json object. If the result is really useable, is a completely different story. @Snugug not sure if you know about libsass internals, but IMO what you're asking for a export of the [ast tree] [1] before it is rendered.

What I see problematic is that we tend to change and add more ast nodes, since we have a lot of internal ones for calculations etc. Most of these types should be expanded/evaluated/cssized, but you currently will get stuff like `Function_Call` in that tree, since libsass is able to render that directly. We do not really compile everything down to a real _css_ ast tree. We probably could add another intermediate step to first produce a _css_ tree, but that would need quite some effort!

I guess those are really the only two options currently available! IMO the second one could lead to a stable API that could actually be used by consumers. We can IMO expose the ast tree as is without that much effort (still quite a bit), but we cannot guarantee any stability with that one!

IMHO it sounds like a good way to start with this (just manage expectations). This way consumers like @Snugug could get some hands on experience with it. Will probably also help to define a first stable export format, once libsass is ready to support it.

[...]

PR #1227, opened 2015-05-20
Pass the compiler to custom functions instead of the options.

tristanlins:
For me it makes a lot more sense to pass the compiler to custom functions, so how is it done also in imports.
With the compiler I can not only access the options, I can access the context and the import stack.
Especially the import stack is very interesting, e.g. when you need to generate relative URLs in a custom function.

### DIFF
--- a/eval.cpp
+++ b/eval.cpp
@@ -300,7 +300,7 @@ namespace Sass {
       To_C to_c;
       union Sass_Value* c_args = sass_make_list(1, SASS_COMMA);
       sass_list_set_value(c_args, 0, message->perform(&to_c));
-      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_options);
+      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_compiler);
       sass_delete_value(c_args);
       sass_delete_value(c_val);
       return 0;
@@ -332,7 +332,7 @@ namespace Sass {
       To_C to_c;
       union Sass_Value* c_args = sass_make_list(1, SASS_COMMA);
       sass_list_set_value(c_args, 0, message->perform(&to_c));
-      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_options);
+      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_compiler);
       sass_delete_value(c_args);
       sass_delete_value(c_val);
       return 0;
@@ -361,7 +361,7 @@ namespace Sass {
       To_C to_c;
       union Sass_Value* c_args = sass_make_list(1, SASS_COMMA);
       sass_list_set_value(c_args, 0, message->perform(&to_c));
-      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_options);
+      Sass_Value* c_val = c_func(c_args, c_function, ctx.c_compiler);
       sass_delete_value(c_args);
       sass_delete_value(c_val);
       return 0;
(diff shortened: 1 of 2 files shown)


xzyfer:
Thanks for the suggestion @tristanlins.

This probably affects @chriseppstein and @drewwells the most. Would also like your feels here @mgreter.

tristanlins:
While testing this, I find out that this wont work.
The `@import` statements are evaluated before any custom function. In result the import stack is always empty when the custom functions are called.

xzyfer:
@tristanlins is a dead end then? Should this issue be closed?

saper:
Probably we should identify the state of the executor vs compiler's state. We should be able to provide a stable interface to the custom function implementors (for example, I'd like to have it serializable #1108). Even if some reflective functions should be made available ( e.g. #1065 ) we may need to provide a stable interface there as well.

Not only custom functions are affected - probably having the same interface for the built-ins would be beneficial as well.

drewwells:
Maybe detailing what use cases this solves would help illuminate why this is necessary. I'm a big believer in only exposing the amount of information needed to extensions to keep them sane. For instance, I do not need to keep track of where I am in the import stack and I don't see a reason custom functions would need that information. The best use case for that is throwing coherent errors. If a custom function returns an error `SassValue`, libsass locates the error and builds a stack trace for me. Not something I'd care to do myself.

tristanlins:
> and I don't see a reason custom functions would need that information.

We need this information because we need to calculate relative paths. Not from the file itself, but from the context the file is defined in. And to know the context, we need to know which file calls the function.
Its a bit complex, because we deliver "virtual" resources with out application and we have a function very similar to the `*-url` compass functions. Expect that our urls depend on the file's context, not a global definable location... it's a lot of magic ;-D

drewwells:
Perhaps there's an alternative way of doing this. https://github.com/wellington/wellington#functions supports all the compass image operations. Just like Compass, file i/o calls like `sprite-map('../')` are based on global configuration, not the sass file in which the function is called. This makes Sass refactoring very simple, since the function calls are independent of the context in which they are called.

It's not entirely clear to me if this level of context information is available to custom functions in Ruby Sass http://sass-lang.com/documentation/Sass/Script/Functions/EvaluationContext.html. It looks like you may able to derive the current template from the http://sass-lang.com/documentation/Sass/Engine.html options passed via the Environment to a Ruby Custom Function.

However, storing files relative to sass is not something I see often. Web projects, especially as they grow, organize similar resources into distinct directories ie. images/, fonts/, or node_modules/eye-glass-module/.

Thoughts?

chriseppstein:
Here's a couple of use cases:

1. sass-based configuration. A sass file should be able to be used to configure a project instead of using an external configuration file. This can only work if you have some concept of where that sass configuration file is on the filesystem.

2. API scoping. In eyeglass, a sass file that is in an NPM module can access the sass files and assets for only those modules for which is has a direct dependency. We are able to control that access for importing because we are given the current sass file that is doing the import, but when functions try to access an asset, I cannot tell whether the asset is accessible to the current sass file unless I know what npm module it is in.

saper:
I see that lots of use cases revolve around mapping "import-identifier" to "actual-imported-resource"

So, in general, what is needed is an array of:

      ('@identifier', 'string')

or

      ('@identifier', sourcefile)

mappings?

Second question:

What about introducing a way to specify such mapping upfront to libsass, something like:
    
      ctx = sass_new_context();
      sass_add_file_source(ctx, 'a.scss', 'sub/a.scss')
      sass_add_file_source(ctx, 'b.scss', 'sub/b.scss')
      sass_add_subtree_source(ctx, 'mixins', 'proj/mixins') /* for subdirectories */
      sass_add_data_source(ctx, 'c.scss', '/* ... */')
      sass_add_custom_source(ctx, 'special.scss', custom_import_func)
      sass_compile();

This might actually reduce the need for custom importers - as those records could be added dynamically before rendering.

mgreter:
Closing this. We are moving in the direction to have more context information available on the C-API. But we need to take step by step. We already have other open issues regarding similar issues, so this PR serve no real use anymore. Exposing the import context is probably one of the next things to do ...

Issue #1160, opened 2015-05-02
Inconsistent import identifiers + memory leak [$100]


saper:
I just spent some time analyzing https://github.com/sass/node-sass/issues/894 (thanks @jameslnewell) which ends up to be a consequence of problems reported also as https://github.com/sass/libsass/issues/1040 https://github.com/sass/libsass/issues/1041 https://github.com/sass/libsass/issues/1035#issuecomment-88968918

Testing done using:

- https://github.com/saper/libsass/commit/ce80c548a20cef247813b41783629a271ffe6eca which is https://github.com/sass/libsass/tree/3.2.2 plus some ad-hoc debug logging
- https://github.com/sass/node-sass/commit/2fce3035ba2374517d42ae4654bd354fa417d20d

Here are the results of my tests:

https://gist.github.com/saper/5263902b8446431ae136

- [issue-894-two-strings](https://gist.github.com/saper/5263902b8446431ae136#file-issue-894-two-strings-out) demonstrates how two string results from the importer override each other;
- [issue-894-two-files](https://gist.github.com/saper/5263902b8446431ae136#file-issue-894-two-files-out) demonstrates how two files get included under the same import name;
- [issue-894-one-file-one-string](https://gist.github.com/saper/5263902b8446431ae136#file-issue-894-one-file-one-string.out) demonstrates how a string-based import cannot override the one from a file;

`libsass` is not consistent about what is being put into `style_sheets` - sometimes it is a short import name, sometimes file base name.

There is also a memory leak in [context.cpp#L330](https://github.com/saper/libsass/blob/ce80c548a20cef247813b41783629a271ffe6eca/context.cpp#L330) - in case of [two-strings.js](https://gist.github.com/saper/5263902b8446431ae136#file-issue-894-two-strings-js) an AST compiled as a result of the first import gets overwritten when we are storing a second AST under the same name `foobar` [issue-894-two-strings.out#L14](https://gist.github.com/saper/5263902b8446431ae136#file-issue-894-two-strings-out-L14).

It should be decided what do the short import names really mean in the context of custom importers (that can return random content every time they are called).

Looks to me that `add_file` and `add_string` need some love. The cleanup there can also help resolve some memory allocation issues hiding in the code (for example only internally allocated strings should end up on the `sources` list, and not those provided by the caller).

Generally I think there too many single-purpose little lists, may be it would be better to have something similar to the parser input queue (with items similar to today's `Sass_Queued`) that contains a directory of all sources:

- source import name
[...]

PR #1228, opened 2015-05-20
ctx.add_source parameter order corrected.

tristanlins:
In `context.cpp` the method add_source has this signature: `Context::add_source(string load_path, string abs_path, ...)`

But the call was `ctx.add_source(file, load_path, source);`.

I would expect the correct call is `ctx.add_source(load_path, file, source);`. Unfortunally this call run into memory violations... but changing to `ctx.add_source(file, file, source);` fix some source map generation issues with custom imports.

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -237,7 +237,7 @@ namespace Sass {
             else error(message, ParserState(message, source, Position(line, column)));
           } else if (source) {
             if (file) {
-              ctx.add_source(file, load_path, source);
+              ctx.add_source(file, file, source);
               imp->files().push_back(file);
             } else {
               ctx.add_source(load_path, load_path, source);


xzyfer:
@tristanlins can you supply an example where this fixes source map issues? From a cursory look I would have expected the current behaviour to be correct.

tristanlins:
> @tristanlins can you supply an example where this fixes source map issues?

If you have a `source.scss` like this:
```css
// foo is imported by a custom importer
@import "foo";
```

and a custom importer that create such an import:
```c
sass_make_import("/some/path/to/foo.scss", "http://example.com", "body { background: red }", "");
```

the expected sourcemap may look like this:
```json
{ "sources": ["/some/path/to/foo.scss"] }
```

but I got:
```json
{ "sources": ["foo"] }
```

I can try to build a simple C example, but I'm not so familiar with C. It's more try-and-error ^^

tristanlins:
> From a cursory look I would have expected the current behaviour to be correct.

Compare the `Context::add_source` declaration at [`context.cpp:250`](https://github.com/sass/libsass/blob/master/context.cpp#L250), with the call at [`parser.cpp:258`](https://github.com/sass/libsass/blob/master/parser.cpp#L258).
For me, it looks as if the parameters are reversed.

saper:
The second argument (`abs_path`) is almost certainly correct, the first argument which serves a kind of generic "imported entity" identifier is up to the interpretation. See also https://github.com/sass/libsass/issues/1160

mgreter:
I'm pretty sure a real solution involves much more work. Closing this since there are other open issues regarding that problem and this PR serves no real purpose anymore.

Issue #1240, opened 2015-05-27
@debug intentionally sometimes produces different output that the css


xzyfer:
We need to refactor some internals for this.

```scss

$var: 1;
$list: 2, 3;
$new-list: append($var, $list);
$nested-list: $var $list;
@debug($var);
@debug($list);
@debug($new-list);
@debug($nested-list);
div {
 a: $var;
 a: $list;
 a: $new-list;
 a: $nested-list;
}
```
Ruby Sass
```css
test.scss:31 DEBUG: 1
test.scss:32 DEBUG: 2, 3
test.scss:33 DEBUG: 1 (2, 3)
test.scss:34 DEBUG: 1 (2, 3)
div {
  a: 1;
  a: 2, 3;
  a: 1 2, 3;
  a: 1 2, 3; }
```
Libsass
```css
test.scss:30: DEBUG: 1
test.scss:31: DEBUG: 2, 3
test.scss:32: DEBUG: 1 (2, 3)
test.scss:33: DEBUG: 1 (2, 3)
div {
  a: 1;
  a: 2, 3;
  a: 1 (2, 3);
  a: 1 (2, 3); }
```

xzyfer:
Related https://github.com/sass/libsass/issues/1246

mgreter:
The DEBUG output is not correct, but the spec test only tests the actual output. Rescoping to 3.3.1.

mgreter:
We now have a passing spec test since #1249 was merged :balloon:

PR #1242, opened 2015-05-27
Fix output of nested lists for in @debug

xzyfer:
This PR fixes the output of nested lists in `@debug`.

Fixes https://github.com/sass/libsass/issues/1240.


xzyfer:
This requires a deep refactor. Putting it on ice for now.

PR #1249, opened 2015-05-30
Refactoring and base work for 3.3

mgreter:
Pushing my latest refactoring work to public since it is currently at a working state that solves all open  `@at-root` issues, plus a few other open issues. ~~PR is still too dirty too actually commit the work to master branch. It also introduced a minor whitespace (linefeed preserving) regressions (from issue 1208), since the fix we added for it doesn't seem to be 100% correct. I will try to tackle the linefeed preserving again at a later stage (the logic for this seems to be clear so far).~~

The following issues should pass with this WIP branch:

- Error with `&` selector interpolation in `@at-root` blocks (#1043)
- & + '' needs to be wrapped in parentheses (#1170)
- @at-root and #{&} produces incorrect output (#1210)
- Unexpected behaviour with @at-root (1236)
- Regression in parsing @at-rules (#1253)
- Dynamic division with interpolation not compatible with libsass (#948)
- Two more tests for `is_superselector` functions (#963)
- libsass-todo-test/38_expression_in_at_directive

This branch also includes quite a few performance optimizations.
It brought my sass-spec test suite runtime from 27s down to 7s!

### TODO

- [x] add attribution to boost for the hash_combine function (https://github.com/boostorg/functional/blob/1bf379ab601eb52c97908cc2cc141fd6026bd708/include/boost/functional/hash/hash.hpp#L209-L213)
- [x] rebase libsass#master
- [x] back out `Improve function call return value handling` (3f2daf3)
  - fixed on master
- [x] back out `Improve deprecation handling in unquote function` (93d3328)
  - fixed on master
- [x] back out `Fix unquote handling for non string types` (f43b881)
  - fixed on master
- [x] back out `Add code improvements for SUN OS compilation` (b523c9e) to it's own PR (#1326)
  - there's still plenty of discussion about the _right_ way to fix this /cc @saper
  - moved to its own PR (#1326)
- [x] back out `Fix memory leaks and improve 'input_source' handling` (af12e9b)
  - there's still plenty of [discussion](https://github.com/mgreter/libsass/commit/af12e9b321e37367a22e6c883b6f5cee29ec6f11#commitcomment-12060509) about the _right_ way to fix this /cc @saper
  - moved to its own PR (#1327)
- [x] back out `Implement initial parts for reference combinators` (2f07ba3)
  - we don't have specs over this yet so we don't know what our requirements are
  - moved to its own PR (#1329)
[...]

PR #1261, opened 2015-06-04
Selector Functions

onedayitwillmake:
#### Please note: The code in this pull request was written by myself, and is does not represent my employer. ####
Redoing https://github.com/sass/libsass/pull/1064.

This is a fresh checkout of master, and then manually went in and re-implemented the selector-functions.
This is a much simpler PR with less files modified, and only a single commit - hopefully that helps make everything more clear when merging 

Adds the selector functions:

- [x] selector-nest($selectors...)
- [x] selector-append($selectors...)
- [x] selector-extend($selector, $extendee, $extender)
- [x] selector-replace($selector, $original, $replacement)
- [x] selector-parse($selector)
- [x] simple-selectors($selector)
- [x] selector-unify($selector1, $selector2)

Regarding the sass-spec test, there are 2 issues when calling `selector-unify` and `selector-nest` and passing in a parent selector reference. 

I believe these are related to a couple of other issues cropping up around `&` - so I think it's safe to merge this in and tackle those as a new issue. 

###### Known Issues from sass-spec libsass issue_963 tests: ######
*selector-unify*
```
selector-unify(&, '.baz, .bang'); // never makes it passed the parser
Error: invalid selector after &
        on line 14 of test_selector_unify.scss
>>   content: selector-unify(&, '.baz, .bang');
   --------------------------^
```

*selector-nest*
```
selector-nest('.foo', '&.bar','baz &"); // '&' before .bar and after .baz are thrown out
```

### DIFF
--- a/functions.cpp
+++ b/functions.cpp
@@ -6,6 +6,7 @@
 #include "constants.hpp"
 #include "to_string.hpp"
 #include "inspect.hpp"
+#include "extend.hpp"
 #include "eval.hpp"
 #include "util.hpp"
 #include "utf8_string.hpp"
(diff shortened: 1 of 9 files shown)


functions.cpp
```diff
@@ -117,6 +118,38 @@ namespace Sass {
       }
       return val;
     }
+    
+#define ARGSEL(argname, seltype, contextualize) get_arg_sel<seltype>(argname, env, sig, pstate, backtrace, ctx)
```
mgreter: `contextualize` doesn't seem to be used?

mgreter:
Hey @onedayitwillmake, already took a quick look yesterday and looks very nice and clean. Unfortunately appveyor doesn't yet pass, which is pretty much the only thing that I see as a blocker before merging this PR as is. @xzyfer I think at this point it will be best to merge this PR to master and I will then realign my WIP branch for the 3.3 release. I guess we need a few beta rounds anyway with 3.3.

Should we create a separate release coordination issue as I did for 3.2?

onedayitwillmake:
@mgreter Thank you! Yes it seems to dislike the `std::function` I used as a custom comparator for creating a `std::set< Complex_Selector*>` - I assumed I could use C++11 feature.

Looking again, i see you already have a Complex_Selector_Pointer_Compare functor I can use instead! 
I will switch to that and commit :)

mgreter:
I have incorporated most of your changes in my latest WIP branch: https://github.com/sass/libsass/pull/1249 - maybe check if I have missed anything! It seems to pass most of the tests and I tried to replace some functionality with more generic functions, which I added during the refactoring!
IMHO we now mostly need to tackle `is_superselector` in regard to `:not` (and other) pseudo selectors!

mgreter:
I guess this can be closed since #1249 has been merged :tada:

Issue #1269, opened 2015-06-11
Cannot join list with arglist


davidkpiano:
**Test:**

```scss
@function push($list, $items...) {
  @return join($list, $items, $separator: auto);
}

.test {
  $list: push(1 2 3, 4, 5);
  list: inspect($list);
  value: nth($list, 4);
}
```

**Expected:** (works in Ruby Sass)

```scss
.test {
  list: 1 2 3 4 5;
  value: 4;
}
```

**Actual:** (crashes/compiles forever in LibSass 3.2.4)

Might be related to #1257 and #1171

PR #1282, opened 2015-06-18
fix memory leak. ?.

VadimGit:
How about remove this loop before 3.3 will be released? IMHO it's really annoying when people try to use code like this https://github.com/sass/libsass/issues/1269 and get this kind of error.

### DIFF
--- a/eval.cpp
+++ b/eval.cpp
@@ -701,12 +701,9 @@ namespace Sass {
     if (result->pstate().file == string::npos)
       result->pstate(c->pstate());
 
-
-    if (result->concrete_type() == Expression::NONE)
-    {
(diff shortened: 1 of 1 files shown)


xzyfer:
We'll take a look @VadimGit. We're not expecting a release before 3.3 next month. There has also been a lot of work refactoring the internals, so it's hard to say if this will still be relevant then.

VadimGit:
I've looked at preparation for 3.3 cursory. And it looks like this will not be actual for 3.3 but like fixing regression for current release.

mgreter:
Checked latest master and this change seems to have gotten into somehow!

PR #1286, opened 2015-06-20
Clone quoted string flag

saper:
On the way to fix:

https://github.com/sass/node-sass/pull/1006


saper:
This is now included in #1287

PR #1287, opened 2015-06-21
Fix quoted strings for C API

saper:
This also introduced a live logging/tracing facility,
used to diagnose problems with dynamic casting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,7 @@ SOURCES = \
 	contextualize.cpp \
 	contextualize_eval.cpp \
 	cssize.cpp \
+	debug.cpp \
 	listize.cpp \
 	error_handling.cpp \
 	eval.cpp \
--- a/Makefile.am
+++ b/Makefile.am
@@ -48,6 +48,7 @@ libsass_la_SOURCES = \
 	context.cpp context.hpp \
 	contextualize.cpp contextualize.hpp \
 	contextualize_eval.cpp contextualize_eval.hpp \
+	debug.cpp \
 	error_handling.cpp error_handling.hpp \
 	eval.cpp eval.hpp \
 	expand.cpp expand.hpp \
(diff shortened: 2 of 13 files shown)


xzyfer:
Thanks @saper.

Please pull the logger into a separate PR. We need the string fix but I'm hesitant to add yet another (undocumented) logger. We should discuss and implement a single debugging interface.

saper:
On Sun, 21 Jun 2015, Michael Mifsud wrote:

> Thanks @saper.
> 
> Please pull the logger into a separate PR. We need the string fix
> but I'm hesitant to add yet another (undocumented) logger. We should
> discuss and implement a single debugging interface.

Yes, but the string fix is most probably incomplete and I need facility to develop
it further... And merging the logging in and out of the code is no fun!

Which debugging interface is documented?  At least this one has
an old DDJ article :)

But here we go:

https://github.com/sass/libsass/pull/1289 <-- string fix

https://github.com/sass/libsass/pull/1288 <-- tracing strings

mgreter:
@saper I guess we currently mostly use `debug_ast` and I recently started to use `debug_node`. IMO none of our debugging stuff is really documented. So let me give you an intro from what I know:

`debug_ast` was introduced by me and is contained in `debugger.hpp` which you explicitly need to include if you want to debug something (and yes, I keep pulling this in and out a lot!). The end PR should not be dependent on this file! You can pass it any `AST_Node` and it will print a debug tree.

`debug_node` is the sister of `debug_ast` and can be used to debug `Node` collections used by the extend facility. Latest version is probably not yet on master!

I guess you already found out how the DEBUG flag works, so I won't go into any details here.

I'm also not really in favor of adding this kind of debug facility. You now added trace calls for every string creation, but why not add it for every type of `AST_Node`? IMO the debug output shortly becomes less and less usefull once we trace more and more stuff. So I'm also hesitant to add this to master.

saper:
Since #1289 got merged and #1288 is still(?) under consideration - closing this one.

PR #1289, opened 2015-06-21
Create quoted strings coming via API

saper:
Make sure quoted strings coming via API
are created as a String_Quoted instance.
Adjust string "quoted" property sent
via API on output.

### DIFF
--- a/sass_values.cpp
+++ b/sass_values.cpp
@@ -322,7 +322,7 @@ extern "C" {
                 return sass_make_color(val->color.r, val->color.g, val->color.b, val->color.a);
         }   break;
         case SASS_STRING: {
-                return sass_make_string(val->string.value);
+                return sass_string_is_quoted(val) ? sass_make_qstring(val->string.value) : sass_make_string(val->string.value);
         }   break;
         case SASS_LIST: {
                 union Sass_Value* list = sass_make_list(val->list.length, val->list.separator);
--- a/to_c.cpp
+++ b/to_c.cpp
@@ -19,7 +19,16 @@ namespace Sass {
   { return sass_make_color(c->r(), c->g(), c->b(), c->a()); }
 
   Sass_Value* To_C::operator()(String_Constant* s)
-  { return sass_make_string(s->value().c_str()); }
+  { 
+    if (s->quote_mark()) {
+      return sass_make_qstring(s->value().c_str());
+    } else {
+	  return sass_make_string(s->value().c_str());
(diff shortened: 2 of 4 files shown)


chriseppstein:
:+1:

sass_values.cpp
```diff
@@ -322,7 +322,7 @@ extern "C" {
                 return sass_make_color(val->color.r, val->color.g, val->color.b, val->color.a);
         }   break;
         case SASS_STRING: {
-                return sass_make_string(val->string.value);
+                return sass_string_is_quoted(val) ? sass_make_qstring(val->string.value) : sass_make_string(val->string.value);
```
mgreter: if/else would be more consitent IMHO

to_c.cpp
```diff
@@ -19,7 +19,16 @@ namespace Sass {
   { return sass_make_color(c->r(), c->g(), c->b(), c->a()); }
 
   Sass_Value* To_C::operator()(String_Constant* s)
-  { return sass_make_string(s->value().c_str()); }
+  { 
+    if (s->quote_mark()) {
+      return sass_make_qstring(s->value().c_str());
+    } else {
+	  return sass_make_string(s->value().c_str());
```
mgreter: Minor indentation / white-space mismatch

mgreter:
LGTM for a first fix! I am aware that we need a better and more complete API for Number and String handling. My goal is to also have operations available to the C-Bindings, so they don't need to implement addition, multiplication, concact etc. themselve. But not sure when/if I or someone else will come around to actually do it ...

saper:
Oh noes... I actually want to use a pure C connector to libsass... please no C++ madness!

mgreter:
@saper not sure what you mean? I'm talking about C functions like `sass_number_subtract` or `sass_string_concat`. So don't see where the problem is?

saper:
I understood wrongly - that you intend to introduce C++ operators. Personally I think that API is already way too large and too bloated. I'd rather have small JavaScript classes handling complex numbers/units or color class than rely on the API. The API should get the data in and out. Of course it does not help that the structures are opaque - I'd love to be able to see them. That would save us a LOT of API calls.

xzyfer:
@saper aside from style cleanups this is ready for 3.3. We'll be starting betas for 3.3. shortly as it's a massive update.

drewwells:
@mgreter :100: to exposing operators, we could very much use that

Issue #1029, opened 2015-04-01
A lost rule when extending inside a double for loop


mzgol:
The following SCSS:
```scss
@for $i from 1 through 2 {
    %shift-#{$i - 1} {
        margin-left: $i * 1px;
    }
    .week-start-#{$i} {
        @for $j from 1 through 2 {
            .month-start-#{$j} {
                @extend %shift-#{($j - $i - 1) % 2};
            }
        }
    }
}
```
libsass from node-sass `3.0.0-beta.4` turns into:
```css
.week-start-1 .month-start-2, .week-start-2 .month-start-1 {
  margin-left: 1px; }
```
Latest Ruby Sass gives:
```css
.week-start-1 .month-start-2, .week-start-2 .month-start-1 {
  margin-left: 1px; }

.week-start-1 .month-start-1, .week-start-2 .month-start-2 {
  margin-left: 2px; }
```

mgreter:
This could be related to https://github.com/sass/libsass/issues/592

mzgol:
Spec submitted: https://github.com/sass/sass-spec/pull/300

mgreter:
This seems to be related to how we eval `Selector_Placeholder` ... seems to involve refactoring!

mgreter:
Analized the test today and found out that the underlying problem is quite simple to solve.
This issue is caused by a discrepancy in the modulo operator, as in this example:

```css
foo { test: #{-1 % 2}; }
foo { test: -1; } // libsass
foo { test: 1; } // ruby sass
```

@xzyfer I will cherry pick this fix and PR it separately later, so this can be included in 3.2.5!

mzgol:
> I will cherry pick this fix and PR it separately later, so this can be included in 3.2.5!

\o/ Thanks!

xzyfer:
@mgreter good news!

xzyfer:
@mgreter did you end up making a PR for this? Looking to get 3.2.5 out the door ASAP.

mgreter:
@xzyfer sorry, no, but the commit could be cherry picked from https://github.com/mgreter/libsass/commit/81a373b3fa0660baece17a83ab61c645a63dfa3d

xzyfer:
Done in https://github.com/sass/libsass/pull/1299

PR #1299, opened 2015-06-29
Fix handling of negative results in module operator

xzyfer:
Fixes https://github.com/sass/libsass/issues/1029

Spec: sass/sass-spec#300


xzyfer:
Superseded by #1313

PR #1300, opened 2015-06-29
Configurable sass-spec environment for CI

xzyfer:
This PR is a POC for being able to configure the sass-spec repo and branch you want run against in CI. This is useful when working on larger features.

/cc @mgreter

PR #1302, opened 2015-06-30
Better configurable sass-spec environment for CI

xzyfer:
This PR is a WIP for a better way to run CI against user defined sass-spec PRs.

I've reverted my `.travis.yml` hack from https://github.com/sass/libsass/pull/1300.

From now on if you have a link to a sass-spec PR well run against that PR in CI. We'll also continue to run against sass-spec master for the time being until we have confidence in this approach.

@mgreter thoughts?


xzyfer:
For clarification this new behaviour **only** happens for Travis Linux builds.

PR #1306, opened 2015-07-03
Fix project to build libsass, not sassc

NickStrupat:
The Debug build of the Visual Studio project was set to include `sassc.c` and compile as an executable. It wouldn't build from a fresh clone. More importantly, this repo should build libsass, not the sassc front-end.


mgreter:
IMHO we would need corresponding VS project files for sassc as we have for libsass. Before that I don't really want to remove sassc, since we need sassc for our CI process.

driekus77:
Perhaps an idea is to compile libsass as a static lib and let sassc as brother/sister project, not as sub library, use this library? 
I already did this in my personal fork: https://github.com/driekus77/libsass.net.mixed. Heir both libsass and sassc are submodules. A checkout of a (main) lib must always build without having todo other thinks first.

mgreter:
Closing this since @am11 added some commits that addresses this issue.

PR #1309, opened 2015-07-07
Avoid global visibility of std:: in units

saper:
Solaris defines SEC in the global namespace

Fixes: https://github.com/sass/libsass/issues/1308

Issue #1308, opened 2015-07-06
SmartOS/Solaris: libsass/units.hpp:39:5: error: expected identifier before numeric constant


djensen47:
```
$ sudo npm install node-sass
-
> node-sass@3.2.0 install /home/admin/src/road-rules-static-website/www/node_modules/node-sass
> node scripts/install.js

Cannot download "https://github.com/sass/node-sass/releases/download/v3.2.0/sunos-x64-14_binding.node": HTTP error 404 Not Found

> node-sass@3.2.0 postinstall /home/admin/src/road-rules-static-website/www/node_modules/node-sass
> node scripts/build.js

Building: /opt/local/bin/node node_modules/pangyp/bin/node-gyp rebuild --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
gyp WARN EACCES user "root" does not have permission to access the dev dir "/root/.node-gyp/0.12.0"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/home/admin/src/road-rules-static-website/www/node_modules/node-sass/.node-gyp"
make: Entering directory '/home/admin/src/road-rules-static-website/www/node_modules/node-sass/build'
  CXX(target) Release/obj.target/libsass/src/libsass/ast.o
In file included from ../src/libsass/ast.hpp:34:0,
                 from ../src/libsass/ast.cpp:1:
../src/libsass/units.hpp:39:5: error: expected identifier before numeric constant
../src/libsass/units.hpp:39:5: error: expected '}' before numeric constant
../src/libsass/units.hpp:39:5: error: expected unqualified-id before numeric constant
../src/libsass/units.hpp:62:3: error: 'SassUnit' does not name a type
../src/libsass/units.hpp:63:30: error: 'SassUnit' was not declared in this scope
../src/libsass/units.hpp:63:30: note: suggested alternative:
../src/libsass/units.hpp:22:8: note:   'Sass::SassUnit'
../src/libsass/units.hpp:64:3: error: 'SassUnitType' does not name a type
../src/libsass/units.hpp:66:34: error: 'string' does not name a type
../src/libsass/units.hpp:66:40: error: ISO C++ forbids declaration of 'parameter' with no type [-fpermissive]
../src/libsass/units.hpp:66:49: error: 'string' does not name a type
../src/libsass/units.hpp:66:55: error: ISO C++ forbids declaration of 'parameter' with no type [-fpermissive]
[...]

PR #1326, opened 2015-07-08
[Proposal] Add code improvements for SUN OS compilation

xzyfer:
This is a proposed fix for https://github.com/sass/libsass/issues/1308 as suggested in https://gist.github.com/jperkin/4b88403da41f556e540d implemented by @mgreter.

An alternate approach is being worked on by @saper in https://github.com/sass/libsass/pull/1309


mgreter:
Closing as the hotfix has been merged. A better fix might still be applied later!

PR #1327, opened 2015-07-08
[Proposal] Fix memory leaks and improve `input_source` handling

xzyfer:
This is a proposed fix for memory leaks and improvements in `input_source` handling implemented by @mgreter.

An alternate approach is being worked on by @saper in https://github.com/mgreter/libsass/pull/5


mgreter:
Superseeded by https://github.com/sass/libsass/pull/1330

PR #1329, opened 2015-07-08
Implement initial parts for reference combinators

xzyfer:
This PR has been extracted from #1249 because it wasn't directly related to the concerns of that PR and due to the lack specs.

**This PR will need to be rebased off master once #1249 is merged**


mgreter:
I'm pretty sure this is superseeded now by https://github.com/sass/libsass/pull/1330

Issue #1305, opened 2015-07-01
Arglist being passed as arguments while empty


HugoGiraudel:
## Test

```sass
.foo {
  content: call('unquote', 'foo', ()...);
}
```

## Expected

```css
.foo {
  content: foo;
}
```

## Actual

```
function quote only takes 1 arguments; given 2 on line 2 at column 12
```

## Observations

Does not matter the function. When passing an empty list as extra arguments, LibSass pass the empty list as another argument instead of discarding it.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/438

Issue #1298, opened 2015-06-29
Malformed URL Error


xzyfer:
```scss
@import url(//fonts.googleapis.com/css?family=Roboto:400,500,700,400italic);
html {
  font-family: roboto, arial, helvetica, sans-serif;
}
```

Sass 3.4.14
```scss
@import url(//fonts.googleapis.com/css?family=Roboto:400,500,700,400italic);
html {
  font-family: roboto, arial, helvetica, sans-serif;
}
```

Libsass 3.2.5
```
URI is missing ')' on line 2 at column 1
```

***Workaround***

Quote the URL.

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/417

Issue #1259, opened 2015-06-03
Error: named arguments must precede variable-length argument


vaanwd:
Hello! I have this error when i try to compile a simple scss file:
```sass
@import "compass"; // Compass 1.0.3

$red : red;

body {
  background-color: $red;
}

h1 {
  @include border-radius(5px);
}
```
and the output from sassc.exe is...
```
Error: named arguments must precede variable-length argument
        on line 152 of d:/Programming/SASS/include/compass/_support.scss
>> e for-legacy-browser($browser, $range..., $threshold: $threshold, $ranges: $br
   ------------------------------------------^
```

xzyfer:
@vaanwd we do not use compass. Please track down the function causing the issue and produce a copy-pastable http://sassmeister.com gist that produces this error.

HugoGiraudel:
Plus, you seem to be using an obsolete/custom/hacked version of Compass as the `for-legacy-browser(..)` function usually looks like this: https://github.com/Compass/compass/blob/stable/core/stylesheets/compass/_support.scss#L148.

HugoGiraudel:
My bad, it comes from here: https://github.com/Compass/compass/blob/stable/core/stylesheets/compass/_support.scss#L152.

HugoGiraudel:
There you have the bug

## Test

```scss
@mixin dummy($a, $b, $c, $d, $e: true) {
  content: $a $b $c $d $e;
}

.foo {
  @include dummy((1 2 3 4)..., $e: false);
}
```

## Expected

```css
.foo {
  content: 1 2 3 4 false;
}
```

## Actual

```
named arguments must precede variable-length argument on line 6 at column 32
```

adambair:
I've run into this too with the recent release of sassc-rails https://github.com/bolandrm/sassc-rails (hit 1.0 yesterday).

```
Error: named arguments must precede variable-length argument
        on line 152 of ../../.rvm/gems/ruby-2.0.0-p451@openbay/gems/compass-core-1.0.3/stylesheets/compass/_support.scss
>> e for-legacy-browser($browser, $range..., $threshold: $threshold, $ranges: $br
```

xzyfer:
Thanks for the report @adambair. Where you using a particular Sass library or framework when you got this error?

PR #1330, opened 2015-07-08
Additional fixes for 3.3 release

mgreter:
This PR contains all comits after the main parser refactoring (#1249)

### DIFF
--- a/ast.cpp
+++ b/ast.cpp
@@ -911,21 +911,38 @@ namespace Sass {
     }
   }
 
-  /* not used anymore - remove?
-  vector<Compound_Selector*> Complex_Selector::to_vector()
+  void Arguments::adjust_after_pushing(Argument* a)
   {
-    vector<Compound_Selector*> result;
-    Compound_Selector* h = head();
-    Complex_Selector* t = tail();
-    if (h) result.push_back(h);
-    while (t)
-    {
-      h = t->head();
-      t = t->tail();
-      if (h) result.push_back(h);
+    if (!a->name().empty()) {
+      if (/* has_rest_argument_ || */ has_keyword_argument_) {
+        error("named arguments must precede variable-length argument", a->pstate());
+      }
+      has_named_arguments_ = true;
     }
-    return result;
-  }*/
+    else if (a->is_rest_argument()) {
+      if (has_rest_argument_) {
+        error("functions and mixins may only be called with one variable-length argument", a->pstate());
+      }
+      if (has_keyword_argument_) {
+        error("only keyword arguments may follow variable arguments", a->pstate());
+      }
+      has_rest_argument_ = true;
+    }
+    else if (a->is_keyword_argument()) {
+      if (has_keyword_argument_) {
+        error("functions and mixins may only be called with one keyword argument", a->pstate());
+      }
+      has_keyword_argument_ = true;
+    }
+    else {
+      if (has_rest_argument_) {
+        error("ordinal arguments must precede variable-length arguments", a->pstate());
+      }
+      if (has_named_arguments_) {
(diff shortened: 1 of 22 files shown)


mgreter:
@xzyfer my other PR #1338 contains these commits. These are IMHO the most important ones. I guess we still can merge @saper un'std branch later (I guess it should fix the SUB OS problems for now). This one should be easy to review, so I hope we can get this into master ASAP!

saper:
I have rebased un'std branch to the current master, so if the new code has `std::vector`, `std::string`, `std::map`, `std::pair`, `std::make_pair()`, `std::min()`, `std::max()` etc. it should merge easily... Would that be ok?

mgreter:
@saper regarding the `AM_PROG_AR`. It works if I also add `m4_pattern_allow([AM_PROG_AR])` which I guess will tell it to ignore the error about not existing `macro `AM_PROG_AR' not found in library`. So does my commit actually fix the original problem? Other than that, I'm not too excited about getting rid of `using namespace std`. It just makes it more tedious to code around ...

saper:
Well `using namespace std;` is a nice shortcut for simple trivial programs, but I am already having trouble to figure out issues with different compilers (Visual Studio redefining min/max https://github.com/saper/libsass/commit/23f89beef745936fbbb82bc95aa1efe19824a682, GCC breaking `std::round()` on non-glibc systems https://github.com/sass/node-sass/issues/1034, and of course the `SEC` problem on Solaris etc.). 

Actually I was surprised that most of the files use `std::string` or `std::vector` rather sparingly, for other `*.cpp` files (but not header files!) it is always possible to do `using std::string; using std::vector` to improve readability. That said - it easier to spot difference between `std::string` and `String`  (meaning `Sass::String`), and we also have a local variable named `string` somewhere...

Another example of related trouble is https://github.com/sass/libsass/commit/5445c456acd1db47608dd20cd267cb3e034c3c62 - you don't really know which `abs()` are you using and how stricter C++ type checking should work...

Only in `ast.hpp` readability is a bit worse - because this "header" file is doing lots of work (instead of `ast.cpp` file for example) and `ast.hpp` gets included almost everywhere, therefore all namespace imports get spoiled onto other files.

Even if this is developer's (slight) inconvenience, people trying to compile libsass on non-Linux platforms will certainly have a simpler job.

mgreter:
@xzyfer This should be merged next for 3.3! Should not be too big for a review ;)

mgreter:
List of fixed open issues:
- https://github.com/sass/libsass/issues/1259
- https://github.com/sass/libsass/issues/1298
- https://github.com/sass/libsass/issues/1305
- https://github.com/sass/libsass/issues/1332

mgreter:
Gave it another reviewed and for me this is ready to :ship:

xzyfer:
I'm not sure about some of the number changes. I need to give it a closer look. The semantics of `strict` aren't clear. It feels like a bit of a hack to me.

xzyfer:
Please pull out f37cd95aef6aba4dd07f41b58b03b6eed1b938d9 into it's own PR. I want to take a closer look. 

The rest looks good to me.

mgreter:
Remerged branch to have requested commit to be on top and created another PR that only includes that specific commit.

xzyfer:
Wow yeah something is up with the commit list. Github think there were two different authors on the commits.

xzyfer:
>Rebased branch to have requested commit to be the first in list

Why make it first? Please just remove the commit from this branch. It'll live on in your other PR.

mgreter:
So this PR is ready once the other PR is merged! And yeah, messed up by using another client, bah.

mgreter:
Removed as requested!

xzyfer:
:ship: when CI is green!

PR #1335, opened 2015-07-10
[WIP] Refactoring/prepare for value op export

mgreter:
This PR is based up on #1249. It is the preparation to have sass value operators exposed on the C-API which needs another decent refactoring, since we might not have the full context available within C-API function calls. It also acts as a preparation to implement better memory handling. Currently we keep all created resources (values) alive until the whole context goes out of scope. In case of thight loops, this wastes a lot of memory. To fix this, we need to be able to allocate values on the right "scope". To implement this correctly, we will need to respect `!global` flags when allocating memory (ToDo)! I also expect a small performance improvement by having the color maps statically implemented.


mgreter:
This is now in a state where I have access to all sass value operations in the C-API via exposed `sass_op` function. But it needs more tests if all operations behave as expected (basic test work so far).

mgreter:
I have it at a state now where we handle memory much nicer. Temporary values allocated in functions and loops are released after their block goes out of scope (or if another variable overwrites its spot in the environment). The following sample now runs at a stable memory usage of ~15MB (via perl-libsass). Previousely it ran pretty fast out of memory (multiple gigs) ...

```css
@for $x from 1 through 5000000 {
  $a: 2px;
  $b: 1*1;
  $c: foobar;
  $d: foo bar;
  $e: "asd"+"qwe";
  $f: "asd"+"qwe";
  $g: (1, asd, "qwe");
  $h: (a: 1, "b": asd, 3: "qwe");
}
```

mgreter:
This is basically superseeded by https://github.com/sass/libsass/pull/1338!
Will leave this open, since the other PR does not include custom memory handling!

drewwells:
Is the memory leak mentioned here https://github.com/sass/libsass/pull/1335#issuecomment-120753718 supposed to be fixed by #1338?

Memory usage is down from 16gb (all memory) to 3-4gb and a nice speed up `Compilation took: 132234.797ms` I think it took 2x as long before.

PR #1341, opened 2015-07-14
[RFC] Change file fallback from `scss` to `sass` syntax

mgreter:
This is in response to https://github.com/sass/libsass/issues/1219.

Ruby sass seems to use indented syntax for all file types except `scss`. libsass currently does the opposite and recognises `sass` file as beeing indented syntax. This PR changes this behavior. It seems to be more inline with ruby sass, but I don't know the implications of this change. I also think we may need to change the force indented syntax flag in this case to force scss syntax.

So this is open for feedback and discussion!

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -336,7 +336,7 @@ namespace Sass {
       }
       for(size_t i=0; i<extension.size();++i)
         extension[i] = tolower(extension[i]);
-      if (extension == ".sass" && contents != 0) {
+      if (extension != ".scss" && contents != 0) {
         char * converted = sass2scss(contents, SASS2SCSS_PRETTIFY_1 | SASS2SCSS_KEEP_COMMENT);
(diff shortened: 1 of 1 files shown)


xzyfer:
I'd like to hear @chriseppstein thoughts on this.

saper:
Maybe we should not rely on extension at all

drewwells:
:100: 

The code won't help, but here's what I do in wellington. It reads the the input stream looking for non-sass tokens `;` `{`, etc, then convert https://github.com/wellington/wellington/blob/master/import.go#L47-L81

The extension matcher wasn't sufficient to handle stdin or http streams.

mgreter:
Ruby sass relies on the extension, so we probably should keep it that way. I agree there are smarter solutions, but so far we always follow the ruby sass way. So in that sense this change would be legit!

PR #1342, opened 2015-07-14
Fix error for multiplication of invalid units

mgreter:
As requested, commit pulled out from https://github.com/sass/libsass/pull/1330!

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -960,8 +960,10 @@ namespace Sass {
       while (true) {
         r = u.find_first_of("*/", l);
         string unit(u.substr(l, r == string::npos ? r : r - l));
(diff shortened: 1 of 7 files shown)


xzyfer:
I'll take a closer look at this over the weekend.

mgreter:
This fixes the error when multiplying values that would result in invalid css values. This check has been moved to the Output phase, but for other operations we still want to error right away. This is what the strict flag is for (it basically is true for DIV and MUL afair). Reason is that multiplications should be valid when assigned to a variable, since it could still be divided so it results in a valid number to output!

xzyfer:
Sorry it took so long. I've dug into this and :+1: .

PR #1344, opened 2015-07-14
Improve Makefile for builds under MSYS/MinGW (Windows)

mgreter:
Add options to build a hopefully fully portable libsass.dll!
Links gcc and stc++ libs statically into the dll itself!

Adds env options `STATIC_LIBSTDCPP`, `STATIC_LIBGCC` and `STATIC_ALL`. They influence how the library (and sassc with https://github.com/sass/sassc/pull/120) is linked against the built library. I was able to build sassc completely static, linked agains a dll that is static (dependency free) and all variations thereoff.

Preparing a little surprise now ...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,48 +33,74 @@ else
 	endif
 endif
 
-ifeq "$(LIBSASS_VERSION)" ""
-	ifneq "$(wildcard ./.git/ )" ""
+ifeq ($(LIBSASS_VERSION),)
+	ifneq ($(wildcard ./.git/ ),)
 		LIBSASS_VERSION ?= $(shell git describe --abbrev=4 --dirty --always --tags)
 	endif
 endif
 
-ifeq "$(LIBSASS_VERSION)" ""
-	ifneq ("$(wildcard VERSION)","")
+ifeq ($(LIBSASS_VERSION),)
+	ifneq ($(wildcard VERSION),)
 		LIBSASS_VERSION ?= $(shell $(CAT) VERSION)
 	endif
 endif
 
-ifneq "$(LIBSASS_VERSION)" ""
+ifneq ($(LIBSASS_VERSION),)
 	CFLAGS   += -DLIBSASS_VERSION="\"$(LIBSASS_VERSION)\""
 	CXXFLAGS += -DLIBSASS_VERSION="\"$(LIBSASS_VERSION)\""
 endif
 
 # enable mandatory flag
 ifeq (MinGW,$(UNAME))
+	ifneq ($(BUILD),shared)
+		STATIC_ALL     ?= 1
+	endif
+	STATIC_LIBGCC    ?= 1
+	STATIC_LIBSTDCPP ?= 1
 	CXXFLAGS += -std=gnu++0x
 	LDFLAGS  += -std=gnu++0x
 else
+	STATIC_ALL       ?= 0
+	STATIC_LIBGCC    ?= 0
+	STATIC_LIBSTDCPP ?= 0
 	CXXFLAGS += -std=c++0x
 	LDFLAGS  += -std=c++0x
 endif
 
-ifneq "$(SASS_LIBSASS_PATH)" ""
+ifneq ($(SASS_LIBSASS_PATH),)
 	CFLAGS   += -I $(SASS_LIBSASS_PATH)
 	CXXFLAGS += -I $(SASS_LIBSASS_PATH)
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -104,6 +126,21 @@ SASS_SPEC_SPEC_DIR ?= spec
...
+
+ifeq (MinGW,$(UNAME))
+	ifeq (shared,$(BUILD))
+		CFLAGS     += -D ADD_EXPORTS
+		CXXFLAGS   += -D ADD_EXPORTS
+		LIB_SHARED  = $(SASS_LIBSASS_PATH)/lib/libsass.dll
+	endif
+else
+	CFLAGS   += -fPIC
+	CXXFLAGS += -fPIC
```
drewwells: I believe you can put this in CPPFLAGS and it covers CFLAGS and CXXFLAGS.

mgreter: I read on that page: CFLAGS for the C **compiler**, CXXFLAGS for C++ **compiler** and CPPFLAGS for the C or C++ **preprocessor** ... seems not to be exactly the same!??

drewwells: Yeah it sends to whatever compiler is being used. So no matter the language, the same params are sent. SO: http://stackoverflow.com/a/495646/217965 

The language in the wikipedia page is super confusing. The CGO docs are way more clear (handles C/C++ compiling/linking for Go)

> All the cgo CPPFLAGS and CFLAGS directives in a package are concatenated and used to compile C files in that package. All the CPPFLAGS and CXXFLAGS directives in a package are concatenated and used to compile C++ files in that package.

mgreter: Be aware we are in a Makefile, so we are in control of these vars and use them ourself. There is no actual connection (beside the name) to gcc until we pass them to the command line:
```
$(CC) $(LDFLAGS) -o $(TARGET) $^ $(LDLIBS)
```

drewwells: True, these are just make variables. It is a nice way of keeping the Makefile dry for duplicated flags like these.

It would require changing the object rules to this:

```
%.o: %.c
	$(CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $<

%.o: %.cpp
	$(CXX) $(CPPFLAGS) $(CXXFLAGS) -c -o $@ $<
```

mgreter: As I said, the current lines are well tested on many systems, so I rather don't change them just to fix cosmetics. But I agree it could be further consolidated and streamlined!

saper:
Sourceforge malware update again? :)

drewwells:
This was happening a lot for me too. I made a mirror of the binary and shaved off a 7minute download from sourceforge :)

https://github.com/wellington/wellington/commit/5bb710ada

Hopefully, somebody can convince appveyor that mingw64 is better than 32bit... http://help.appveyor.com/discussions/questions/372-build-setup-for-a-c-program#comment_34629599

drewwells:
Apparently, SourceForge has had lots of issues with serving content. msys's package manager has detailed tests to make sure SourceForge is serving up packages.

```
error: failed retrieving file 'mingw32.db' from downloads.sourceforge.net : Connection timed out after 10873 milliseconds 
error: failed to update mingw32 (download library error)
error: failed retrieving file 'mingw64.db' from downloads.sourceforge.net : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds 
```

xzyfer:
I have no input on Makefile changes. Do what you think is best @mgreter :rocket: 

I only ask you make the requires upstream changes to the [homebrew formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/libsass.rb) should they be needed.

drewwells:
make works locally on darwin for me. I can give the brew path a run with `brew install --HEAD` test after this is in master.

Fortunately, if any of these static flags bled into darwin, it would just :hankey: the bed. So finding problems should be easy :+1:

mgreter:
So the surprise is that I created a project to create an MSI installer for libsass. Pretty much a rip-off from one of my other projects, so it was not really too difficult to make this happen :)

:package: http://webmerge.ocbnet.ch/libsass/installer/
:slot_machine:  https://github.com/mgreter/libsass-msi-installer

drewwells:
Love seeing packaged built versions of libsass. Hope we start seeing this
for more platforms :)

mgreter:
Would be nice if people who tried it out could report if it works. I'm not 100% certain, since I probably have dlls around for most things. So not sure if it contains everything to work on a new system ...

PR #1345, opened 2015-07-15
Implement color maps as static object

mgreter:
Makes the color tables available to all code parts!
Also saves the run-time initialization of the maps!
I actually used a perl script to generate the source file!


mgreter:
@xzyfer this should be merged next ...

color_maps.cpp
```diff
@@ -0,0 +1,326 @@
...
+  const char* color_to_name(int key)
+  {
+    auto p = colors_to_names.find(key);
+    if (p != colors_to_names.end()) {
+      return p->second.c_str();
+    }
+    return 0;
+  }
+
+  const char* color_to_name(Color* c)
```
xzyfer: The header file defines this as `const char* color_to_name(const Color* c)`

xzyfer:
Looks like a header definition mismatch. Otherwise :ship:

mgreter:
Thanks, good catch! Finally this whole thing should be const, but for now this is not possible. It should be once the whole `sass_op` PR is merged!

mgreter:
Wating for CI :traffic_light: to report back ...

Issue #1328, opened 2015-07-08
Unexpected \a between selectors 


dennisfrank:
I stumbled over following issue:

http://sassmeister.com/gist/51e6f8724f98f49ffcf4

```
@mixin foo($foo: bar) {
  .#{$foo}:before,
  [data-attribute="#{$foo}"]:before,
  [data-another-attribute="#{$foo}"]:before {
    content: "whatever";
  }
}

@include foo;
```

will become

```
.bar:before,
[data-attribute="bar"]:before, \a [data-another-attribute="bar"]:before {
    content: "whatever";
}
````

instead of

```
.bar:before,
[data-attribute="bar"]:before,
[data-another-attribute="bar"]:before {
    content: "whatever";
}
```

**The line-break between the second and third selector gets replaced by `\a` and breaks the CSS.**

If you remove `.#{$foo}:before,` or append a space or another character to the data-attribute value like `[data-attribute="#{$foo} "]:before,` the `\a` disappears but the CSS is wrong.

As a crude workaround you can replace the double quotes of one(!) selector with single quotes:

```
@mixin foo($foo: bar) {
  .#{$foo}:before,
  [data-attribute='#{$foo}']:before,
  [data-another-attribute="#{$foo}"]:before {
    content: "whatever";
  }
}

@include foo;
```

will become

```
.bar:before,
[data-attribute='bar']:before,
[data-another-attribute="bar"]:before {
    content: "whatever";
}
````

When both selectors contain the same quotes (double or single) it breaks again.

**Comment by mgreter** [comment]
This seems already fixed with my latest refactoring branch (//CC @xzyfer)

**Comment by dennisfrank** [comment]
@mgreter @xzyfer If this is already fixed, that would be great! 

BTW same issue with `@each`:

```
@each $item, $value in $items {
	.#{$foo}:before,
	[data-attribute="#{$foo}"]:before,
	[data-another-attribute="#{$foo}"]:before {
		content: "whatever";
  }
}
```

xzyfer:
I've seen a couple reports of this. It'd be good to have a failing test case to ensure this doesn't regress. I'll work it.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/446

mgreter:
it seems that the fix for this is somewhere burried in one of my refactorings ...

mgreter:
IMO this is not fixed yet, waiting for that fix to come up ...

mgreter:
[...]

Issue #1295, opened 2015-06-24
Apostrophes converted to escaped quotes, when interpolation used


marek-saji:
I've encountered a bug when using interpolation to customize inlined SVGs:

```css
div
{
    $size: '10px';
    $color: '%23FF0000';
    background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100% 100%' width='#{$size}' height='#{$size}'><circle cx='50%' cy='50%' r='50%' fill='#{$color}' /></svg>");
}
```

produces:

```css
div {
  background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100% 100%' width='10px' height='10px\"><circle cx='50%' cy='50%' r='50%' fill=\"%23FF0000' /></svg>"); }
```

Note the `\"` in `height` and `fill` attributes.

I'm using libsass 3.2.4:

```
$ node-sass --version
node-sass       3.1.2   (Wrapper)       [JavaScript]
libsass         3.2.4   (Sass Compiler) [C/C++]
```

Minimal case producing the problem I've been able to reproduce:

```css
div
{
    $nothing: null;
    a: "#{$nothing}' %' '#{$nothing}";
}
```

which produces:

```css
div {
  a: "\" %' \""; }
```

xzyfer:
Thanks @marek-saji I've confirmed this is a bug.

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/418

marek-saji:
In case someone have the same problem and finds this issue — temporary workaround for my SVG problem is to use single interpolation:

```css
div
{
    $size: '10px';
    $color: '%23FF0000';
    background-image: url("data:image/svg+xml,#{ "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100% 100%' width='" + $size + "' height='" + $size + "'><circle cx='50%' cy='50%' r='50%' fill='" + $color + "' /></svg>" }");
}
```

PR #1349, opened 2015-07-15
Replace `String_Quoted` with `String_Constant` in parser where possible

mgreter:
Fixes https://github.com/sass/libsass/issues/1328 and https://github.com/sass/libsass/issues/1295

### DIFF
--- a/parser.cpp
+++ b/parser.cpp
@@ -591,7 +591,7 @@ namespace Sass {
       // try to parse mutliple interpolants
       if (const char* p = find_first_in_interval< exactly<hash_lbrace> >(i, end_of_selector)) {
         // accumulate the preceding segment if the position has advanced
-        if (i < p) (*schema) << new (ctx.mem) String_Quoted(pstate, string(i, p));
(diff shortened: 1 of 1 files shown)


xzyfer:
\o/ (hooray!)

xzyfer:
Happy to merge this!

I'd love a quick explanation though as to why this fixes the issue. I am not clearer on when to a quoted string vs a constant.

mgreter:
Basically `String_Constant` are literals, while `String_Quoted` are actual strings with semantics. Internally we store the latter in unquoted form and decide on output if we should quote them again (usefull for comparison to literals or concatenation). AFAIR, we previousely had the flag `needs_unquote`, which basically represented the same as these two string classes. I basically just introduced them for interpolation; but I guess there is a difference between actual sass string values and i.e. static expressions (where ruby sass even preserves block comments).

As you know interpolation is still a mess, and fixing that will definitely help clearing up the mess with the string classes. BTW: the list without spaces is `String_Schema`, but output (escaping) is off to ruby sass ...

xzyfer:
Thanks for the explanation.

dennisfrank:
Thanks guys! :)

Issue #1223, opened 2015-05-18
libsass version is missing when building to a new directory


drewwells:
When configuring libsass from a clean directory, libsass_version always reports "[NA]". Even though, `make install` has the correct version in `$(build_path)/include/sass_version.h`.

So here's what I did...

```
cd src; autoreconf -fvi
cd build; ../src/configure --disable-shared --prefix=$(shell pwd) --disable-silent-rules --disable-dependency-tracking
cd build; make install
```

I guess it is still building the static library from src rather than from the files in build that are laid down by configure. Since configure lays down the files in the current directory rather than source directory, the modified sass_version.h is in build/. However, the archive file appears to have `"[NA]"` baked into it.

Here's my makefile: https://github.com/wellington/go-libsass/blob/master/Makefile#L20

mgreter:
We don't actually release any archives. If you create one, you decouple the source files from git control and have to add a VERSION file. I have this automated with `perl-libsass`, since I also include the full sources with the distribution: http://cpansearch.perl.org/src/OCBNET/CSS-Sass-v3.2.2/libsass/VERSION. We really don't want to have the version baked into git either!!

drewwells:
It is cool that it will read a VERSION file, good to know!

This is not for generating archives, it's for making custom configured versions of libsass. `configure` is designed to allow for this behavior. You could for instance do.

```
cd ../shared-libsass; ../libsass/src/configure --enabled-shared
cd ../static-libsass; ./libsass/src/configure --disabled-shared
```

This already works properly, but the versioning is missing. A build rule or changing the order of things happening would fix this though... This is the recommended way of working on implementation projects that want to maintain a pristine source. However, when you do either of these the `sass_version.h` file still gets modified in the source. It should instead modify the target location.

mgreter:
We need that header at build time, since it will be baked into the library. Thinking about it, I see that we should not really install sass_version header, since it serves no purpose for implementors (you should never include it, always call libsass_version function of the library)!

The file is overwritten by the configure step, so I guess if you modify it after that step, it should be picked up. But I really have a hard time to follow your steps. Your makefile is incomplete and the version target is never called or used as a dependency!? So it's hard to tell what is going wrong, but it should certainly work, as I just tested locally!

drewwells:
src refactor fixed this issue, It is now working. 

```
mkdir -p $(NEWDIR);
cd $(LIBSASS_SRC); autoreconfig -fvi
cd $(NEWDIR); ./configure
```

[...]

PR #1352, opened 2015-07-16
Move source files into `src` subfolder

mgreter:
Move sources into sub folder and other code reorganizations (only last commits are relevant). Waiting for other PRs to be merged, so this can be rebased! For now I also had to use my local sassc repo.


drewwells:
This is going to cause a lot of problems for me. Is there a specific entrypoint to include c/h files? Writing a cpp file like

mega.cpp
```
import ast.cpp
...
import sass_value.cpp
```

Just creates a broken compile time situation. Can as part of this some sort of import source/header that includes all files in the correct order?

mgreter:
The only headers you should include are the ones that will be in `include` subfolder (only C-API headers). I really hope you don't import any of the `hpp` headers, since they are not meant for consumers to be included. Also don't see how this could cause much problems, since you just need to pass the new include path to the compiler (mostly `-I include`).

I guess it will take me 5 minutes to update `perl-libsass` to use the new file structure!

drewwells:
I build libsass from source since it makes the Go binary more portable.

 `go get github.com/wellington/wellington/wt` compiles libsass C/C++, links
C-API to Go, then builds the final Go binary. I've tried amalgamating
libsass, but it ends up with namespace conflicts.

mgreter:
perl-libsass builds it from source too. Can't you just pass another include directory to the invoked compiler? Seems to be one of the most common cases IMO! Or I don't understand your concerns!!

xzyfer:
@mgreter I'd prefer to avoid this for now. WE should come back to this after 3.4 when we have parity. For I the over head this is going to introduce when going through git history to find issues/regressions has more cost than benefits here.

drewwells:
I'll check out your perl-sass setup and see if it helps. Thanks!

mgreter:
@xzyfer this should all be unproblematic!

a) there is one commit that just moves the files. So we don't lose history or anything!
b) it is how pretty much all other projects are organized and it helps keeping root clean!
c) Compilation only needs two changes: new path for sources and additional include directories!

xzyfer:
:ship: when :traffic_light: is ready

mgreter:
Here the commit needed to adjust perl-libsass: https://github.com/sass/perl-libsass/commit/cdb2c4429d84c4331b226a5b5c5e008b41ded0ea

am11:
Looks pretty reasonable to me. :+1:

mgreter:
Merged changes needed for sassc spec runner to pass CI ... let's wait and see :traffic_light:

saper:
Hmm, not sure I love it but I think there are few things to consider:

1) Can we fix the `Makefile` situation? It is highly annoying to have a version-controlled `Makefile` overwritten by `configure`

2) `lib` directory will be confusingly not used in the libtool case, correct? Can we maybe use `src/.libs` instead?

The reason why it may be useful - I am compiling `node-sass` bindings with library paths set to `$HOME/sw/libsass/.libs/libsass.so`. This way I can test changes in `libsass` without touching the `node-sass` bindings. Would love to have this location for *both* autoconf and non-autoconf cases.

3) I think if this reorganization would serve any real purpose, I'd like to have out-of-tree building possible. MSVC is always doing this, we could have this with GNU make.

Some middle-term thoughts:

4) I think `ast.hpp` is too big. What about introducing one source file per type (`number.cpp`, `string.cpp`, ....) to combine all functions related to one node type into one file (and possibly one class). This could have some additional advantages like getting rid out classes like `Output` and probably we could have much less `dynamic_cast`s (if any).

drewwells:
It would be nice if this worked too. Or there was a make task to update it without performing a build. https://github.com/sass/libsass/issues/1223

mgreter:
@saper 

1) I don't see a way around it without getting rid of either one. Since mingw does not support autotools easily (and it is also pretty slow to generate the makefiles, if you ofter start with an empty tree). So I don't want to get rid of it. So I don't see a solution for this annoyance. 

2) which lib directory are you talking about. The lib will be built in `build/lib` and installed to `$prefix/lib`. Not sure what you mean by "not used"? Simply define the correct prefix and do `make install`, then tell node-sass to use the lib from there!?

3) Agree, but one step at a time. I've read this is not that easy with autotools!

4) Also agree, but cannot fix everything at once :wink:

saper:
re 1) What about having hand-generated Makefile in one directory and autotools-generated Makefile in another? Maybe `GNUmakefile` can be used somehow (generated....)?

The reason I raise this I want to avoid another move because this will be easier for some other reason (simplifying the script for example).

Here's a simpler https://github.com/saper/sassc/blob/fixautoconf/ci-build file that actually can be made even more simple. And we should almost never need to change the directory in the script if the Makefiles are right...

mgreter:
@saper I don't really see why 1) is such a big problem. The only annoyance I have with it is that I have to be carefull when changing the makefile without commiting and later running autoreconf (which would overwrite any previous unstaged changes). I also accidentaly commit it once or twice, but nothing a rebase/ammed could fix. Other than that I don't see any issues!?

mgreter:
Pretty sure CI is going :green_heart: in a few minutes and I'll :ship: this. It has been tested on mingw (makefiles via CI) under cmd, mingw under msys (makefiles and autotools), gcc and clang on linux (makefiles and autotools via CI), clang on mac (makefiles and autotools via CI) and MSVC under windows via CI. Also tested building sassc directly under windows in all variations (plus CI tests on linux).

saper:
It is a problem, because `git status` always shows modified `Makefile` and `sass_version.h` - one needs to hand-pick files to commit and be careful not to include those; this contributes to getting `dirty` version indicator even it is not.

mgreter:
I guess not much we can do about that for now, I only consider it an annoyance. Maybe you can tell autotools to build the Makefile somewhere else for you? Or use the plain makefiles!?

BTW: https://github.com/sass/libsass looks much better, doesn't it?

saper:
https://github.com/sass/libsass/pull/1358 moves Makefile.am to GNUmakefile.am (plus fixes some issues with `.gitignore`)

drewwells:
It's also an order of operation problem. Sass version is modified in the
source directory not the build directory as the issue above mentions. So
version doesn't work when targeting a new build directory.

Issue #1322, opened 2015-07-08
Fails to evaluate variables and interpolants in media queries


xzyfer:
```scss
$foo: 400px;
$bar: "min-width:400px";
@import url(foo.css) (min-width:400px);
@import url(foo.css) (min-width:$foo);
@import url(foo.css) (min-width:#{$foo});
@import url(foo.css) ($bar);
@import url(foo.css) (#{$bar});
```
Ruby Sass
```scss
@import url(foo.css) (min-width: 400px);
@import url(foo.css) (min-width: 400px);
@import url(foo.css) (min-width: 400px);
@import url(foo.css) (min-width: 400px);
@import url(foo.css) (min-width: 400px);
```
LibSass 3.2.5
```scss
Error: top-level @import directive must be terminated by ';'
        on line 3 of test.scss
>> @import url(foo.css) (min-width:400px);
```

Spec added https://github.com/sass/sass-spec/pull/431

xzyfer:
https://github.com/sass/libsass/pull/1324 fixes the error, but interpolation fails.

```scss
@import url(foo.css) (min-width: 400px);
@import url(foo.css) (min-width: $foo);
@import url(foo.css) (min-width: #{$foo});
@import url(foo.css) ($bar);
@import url(foo.css) (#{$bar});
```

mgreter:
The error was that we did not evaluate the media queries in import statements. But there is another strange discrepancy. I would like to change the spec test to:

```css
$foo: 400px;
$bar: "min-width: 400px";
@import url(foo.css) (min-width:400px);
@import url(foo.css) (min-width:$foo);
@import url(foo.css) (min-width:#{$foo});
@import url(foo.css) ($bar);
@import url(foo.css) (#{$bar});
```

Just added a space, since libsass really just adds the string from the variable. It seems ruby sass does a re-interpretation when something is inserted there. See this example:

```css
$bar: "min-width: 2+400px";
@import url(foo.css) ($bar);
@import url(foo.css) (#{$bar});
```

Output:

```css
@import url(foo.css) (min-width: 2 (+400px));
@import url(foo.css) (min-width: 2 (+400px));
```

Not sure why ruby sass is doing that!??

xzyfer:
Should this be reopedn @mgreter ?

mgreter:
Fix has been merged! Maybe we should open a new one for the remaining white-space issue!

xzyfer:
Sounds good.

PR #1353, opened 2015-07-16
Evaluate media_queries for import statements

mgreter:
Fixes https://github.com/sass/libsass/issues/1322

### DIFF
--- a/debugger.hpp
+++ b/debugger.hpp
@@ -311,7 +311,7 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
     cerr << ind << "Import " << block;
     cerr << " (" << pstate_source_position(node) << ")";
     cerr << " " << block->tabs() << endl;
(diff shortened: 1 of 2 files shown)


debugger.hpp
```diff
@@ -311,7 +311,7 @@ inline void debug_ast(AST_Node* node, string ind = "", Env* env = 0)
     cerr << ind << "Import " << block;
     cerr << " (" << pstate_source_position(node) << ")";
     cerr << " " << block->tabs() << endl;
-    // debug_ast(block->media_queries(), ind + " @ ");
+    debug_ast(block->media_queries(), ind + " @ ");
```
xzyfer: remove this?

mgreter: Why should this be removed? It debugs the associated media queries?

xzyfer: you're right

PR #1354, opened 2015-07-16
Fix parsing for interpolated reference combinator

mgreter:
Fixes https://github.com/sass/libsass/issues/452


xzyfer:
@mgreter create a PR for the spec activation and put a link to it in the PR description. This way we never forget specs, and CI can assert the spec passes :)

xzyfer:
LGTM

PR #1357, opened 2015-07-19
Add automatic git versions for vcxproj file

mgreter:


am11:
:+1:

xzyfer:
LGTM

saper:
I think we should start generating `sass_version.h` files in any case. 

What happens here if git is unavailable/not in PATH/on strike?

drewwells:
Building Go from source required a VERSION file whenever git was
unavailable. Looks like a decent approach.

`go tool dist: FAILED: not a Git repo; must put a VERSION file in $GOROOT`

mgreter:
@saper if git is not available it will probably fail. We mainly use these files for CI and they are not polished enough anyway to build all possible combinations (static/shared). Heck they even have a hard reference to `sassc.cpp`, since we need that to run the spec test suite.

PR #1359, opened 2015-07-20
Add cmake build files

mgreter:
This is an early WIP PR and supersedes https://github.com/sass/libsass/pull/292! I basically just pushed it here to close #292 and so others could build on that base if they want! I just tested it on windows and MSVC, but I'm confident that it should work with other flavors too, since I would blame cmake otherwise :)


am11:
@mgreter, just wanted to stop by and say thanks for overhauling the build process and opting for cmake. :+1:

xzyfer:
Why was this closed @mgreter ?

Issue #1294, opened 2015-06-24
Incorrect output of line breaks within comments


joejoinerr:
I like to head up my partials with a comment so that I can find the relevant section easily in the generated CSS. Example:

```
/*------------------------------------*\
  #BUTTONS
\*------------------------------------*/

.btn {
  ...
}
```

However, when the SCSS is processed with Libsass, the line break is removed and I am left with this output:

```
/*------------------------------------*  #BUTTONS
\*------------------------------------*/
```

Also, when documenting `property: value` pairs inline, I use this syntax:

```
.btn {
  display: inline-block; /* [1] */
  ...
}
```

Interestingly, when this is processed, a line break is _added_:

```
.btn {
  display: inline-block;
  /* [1] */
  ...
}
```
##

mgreter:
This is related to String_Quoted and also related to https://github.com/sass/libsass/issues/1026. Since we need a different parser for 1026, this will probably be fixed at the same time! The problem is the backslash before the newline.

xzyfer:
The latter declaration example is consistent with Ruby Sass.

xzyfer:
Specs added https://github.com/sass/sass-spec/pull/415

xzyfer:
Related https://github.com/sass/libsass/issues/1096

mgreter:
@xzyfer IMO this is not related to #1096, they seem to trigger two different parse paths. Comments need a new parse path, since block comments undergo a strange indent normalization!

xzyfer:
@mgreter  you're correct. I spent some trying to fix this issue and as you say this requires significant work. This is a major corner case and isn't a priority for 3.3 given the amount of work required.

xzyfer:
Rescoped to 3.4 since this only affects insignificant whitespace and as such doesn't prevent people migrating to LibSass from Ruby Sass.

PR #1364, opened 2015-07-20
Allow linefeeds in urls and remove from output

### LINKED ISSUE #1096: Permit multiline in URL

mgreter:
Fixes #1096

### DIFF
--- a/src/constants.cpp
+++ b/src/constants.cpp
@@ -140,7 +140,7 @@ namespace Sass {
     extern const char arglist_name[]    = "arglist";
 
     // constants for uri parsing (RFC 3986 Appendix A.)
-    extern const char uri_chars[]  = ":;/?!$%&#@|[]{}'`^\"*+-.,_=~";
+    extern const char uri_chars[]  = ":;/?!$%&#@|[]{}'`^\"*+-.,_=~\r\n";
 
     // some specific constant character classes
     // they must be static to be useable by lexer
--- a/src/inspect.cpp
+++ b/src/inspect.cpp
@@ -16,7 +16,8 @@ namespace Sass {
   using namespace std;
 
   Inspect::Inspect(Emitter emi)
-  : Emitter(emi)
+  : Emitter(emi),
+    strip_newlines(false)
   { }
   Inspect::~Inspect() { }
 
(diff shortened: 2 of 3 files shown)


xzyfer:
IMHO this is a little hacky. We should figure a our a proper fix for this which will also address https://github.com/sass/libsass/issues/1294. They're the same issue IMO

mgreter:
We cannot really fix everything correctly at once. This would need a major refactoring of all string types! So we either go with the hacky one or leave it open for the next months or year! It also helps to start to figure out what really are the different use cases. A lot has been cleared up lately because we started to work in the right direction (often with hacky stuff and flags first). Once everything works with spec test a correct refactoring and putting these options in the right place is much easier IMHO!

mgreter:
BTW. this should probably be done when creating `String_Quoted` instances, but currently we also parse comment into a `String_Quoted`, meaning we can only add this to the right place after i.e. the block comments are pares correctly (and probably other objects). That's why I opted for a flag on output. It should be removed once it can be added to the String class itself!

PR #1365, opened 2015-07-21
Remove lib/ directory on clean

saper:
We create this directory during build
so it should be removed by the "clean"
target as well.

Checked-via: git clean -dxn

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ CXX      ?= g++
 RM       ?= rm -f
 CP       ?= cp -a
 MKDIR    ?= mkdir
+RMDIR    ?= rmdir
 WINDRES  ?= windres
 INSTALL  ?= install
 CFLAGS   ?= -Wall
@@ -290,7 +291,8 @@ test_issues: $(SASSC_BIN)
 	$(RUBY_BIN) $(SASS_SPEC_PATH)/sass-spec.rb -c $(SASSC_BIN) $(LOG_FLAGS) $(SASS_SPEC_PATH)/spec/issues
 
 clean-objects:
-	$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
+	-$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
+	-$(RMDIR) lib
 clean: clean-objects
 	$(RM) $(CLEANUPS)
 


xzyfer:
Makes sense to me.

mgreter:
In principle `$(RM) -r` is not portable. Might not be that big of an issue as ie. `mkdir -p`, but what if we have `del` set for RM! Might not be a problem now, but if we use specific options, it makes no sense to use variables for the calls!

saper:
> In principle `$(RM) -r` is not portable. Might not be that big of an
> issue as ie. `mkdir -p`, but what if we have `del` set for RM! Might
> not be a problem now, but if we use specific options, it makes no
> sense to use variables for the calls!

Good point, we can have have $(RMF). Is "del" working with "-f" ?

mgreter:
Del has no such support!

saper:
Well, there is `del /s` :) But I went for `rmdir` as you have suggested, thanks!

mgreter:
Actually `del` did not have this option from the begining. But I see that more and more of my knowledge gets obsolete :older_man:  Seems like win98 was the last win OS that did not implement this flag (I remember we used deltree in the old times). Looks like it got introduces with windows 2000.

Makefile
```diff
@@ -290,7 +291,8 @@ test_issues: $(SASSC_BIN)
 	$(RUBY_BIN) $(SASS_SPEC_PATH)/sass-spec.rb -c $(SASSC_BIN) $(LOG_FLAGS) $(SASS_SPEC_PATH)/spec/issues
 
 clean-objects:
-	$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
+	-$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
+	-$(RMDIR) lib
 clean: clean-objects
```
mgreter: Why is there a dash before the variables/command? And I somehow would rather put the rmdir into the clean target itself, but I guess that's just a personal preference, but maybe you can see where I'm going ...

saper: > >  clean-objects:
> > -	$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
> > +	-$(RM) lib/*.a lib/*.so lib/*.dll lib/*.la
> > +	-$(RMDIR) lib
> >  clean: clean-objects
> 
> Why is there a dash before the variables/command? And I somehow would rather put the rmdir into the clean target itself, but I guess that's just a personal preference, but maybe you can see where I'm going ...

Dash is to ignore error from rmdir. rm -r does not fail when the file
does not exist; rmdir does.

Otherwise "make clean" fails when "lib" exist.

PR #1371, opened 2015-07-22
access npos member constant from string

drewwells:
access npos from the string object. Moving the define out of the global space.


saper:
Oh that was ugly, thanks!

xzyfer:
Thanks @drewwells

PR #1372, opened 2015-07-22
remove default argument from instantiation

drewwells:
The default argument is in the definition https://github.com/sass/libsass/blob/bc802ba94/src/sass_util.cpp#L112 so the instantiation throws this error.

`error: redefinition of default argument`


xzyfer:
Thanks @drewwells

PR #1373, opened 2015-07-22
move methods from directives to functions

drewwells:
This moves the `is_digit` and `is_space` directives to methods.

### DIFF
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -357,9 +357,8 @@ static void to_surrogate_pair(uint32_t unicode, uint16_t *uc, uint16_t *lc)
   *lc = (n & 0x3FF) | 0xDC00;
 }
 
-#define is_space(c) ((c) == '\t' || (c) == '\n' || (c) == '\r' || (c) == ' ')
-#define is_digit(c) ((c) >= '0' && (c) <= '9')
-
+bool is_space        (const char *c);
+bool is_digit        (const char *c);
 static bool parse_value     (const char **sp, JsonNode        **out);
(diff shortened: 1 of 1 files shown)


drewwells:
sourceforge!

xzyfer:
Thanks @drewwells

src/json.cpp
```diff
@@ -357,9 +357,8 @@ static void to_surrogate_pair(uint32_t unicode, uint16_t *uc, uint16_t *lc)
   *lc = (n & 0x3FF) | 0xDC00;
 }
 
-#define is_space(c) ((c) == '\t' || (c) == '\n' || (c) == '\r' || (c) == ' ')
-#define is_digit(c) ((c) >= '0' && (c) <= '9')
-
+bool is_space        (const char *c);
+bool is_digit        (const char *c);
```
mgreter: This declaration doesn't seem to match the implementation?

drewwells: In what way? I declared it as static bool and some of the windows and os x
builds failed. This declaration passed all builds though.

xzyfer: @drewwells 

```
bool is_digit (const char *c)
```
vs
```
bool is_space(const char& c)
```

My mistake for noticing this.

drewwells: Ahh good catch, I'll PR a fix.

xzyfer: I'm working on it in https://github.com/sass/libsass/pull/1395.

There are other inconsistencies I want to fix.

PR #1378, opened 2015-07-24
Fixing multiple MSVC compiler warnings and code analysis

driekus77:
Some fixed code issues. Compiled and tested ok on win64.
Kind regards,
Henry

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1,3 +1,7 @@
+#ifdef _MSC_VER
+#pragma warning(disable : 4503)
+#endif
+
 #include "ast.hpp"
 #include "context.hpp"
 #include "node.hpp"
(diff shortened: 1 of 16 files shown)


saper:
In general - as noted partially in the comments - we need to check whether we need assertion or whether we should somehow handle the situation.

We definitely SHOULD NOT cover up errors with `if()` - the [code should fail fast](http://martinfowler.com/ieeeSoftware/failFast.pdf) when things go wrong -  finding errors will be more difficult.

driekus77:
Thanks for the "fail fast" lesson. Martin Fowler related reading is always ok. 

I will close this pull request and create 2 new ones:
- Fixing Warnings with MSVS 2015
- Fixing Code analysis issues with MSVS 2015

Again thanks for the quick feedback.

PR #1382, opened 2015-07-25
Let Memory_Manager actually manage memory

saper:
It looks like `operator()` never gets called (why
should it be?) and the nodes list is always
empty, so we do not release any resources
when the Memory_Manager is destroyed.

### DIFF
--- a/src/memory_manager.cpp
+++ b/src/memory_manager.cpp
@@ -20,19 +20,11 @@ namespace Sass {
     // release memory for all controlled nodes
     // avoid calling erase for every single node 
     for (size_t i = 0, S = nodes.size(); i < S; ++i) {
+      
       deallocate(nodes[i]);
     }
     // just in case
     nodes.clear();
-  }
-
-  template <typename T>
-  T* Memory_Manager<T>::operator()(T* np)
-  {
-    // add to pool
-    nodes.push_back(np);
-    // return resource
-    return np;
   }
 
   template <typename T>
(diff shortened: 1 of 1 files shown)


mgreter:
Thanks for the report! I messed this up in commit https://github.com/sass/libsass/commit/451980e52fdd11598cb2877f482698b049c3960e#diff-cb34a4b7c469c2c15b7102db29e0daf2L48. (`operator()` was called from there). Unsure if we should go back to how it originaly was (separate allocation and giving memory under its control) ...

saper:
It seems to me this separation is unnecessary... I think it should be kept as simple as possible. Additionally I find overloading of `()` mostly confusing.

mgreter:
I probably need the seperation once I implement the correct memory handling (scoping). At some point we may move items from an inner scope to the global scope. But I agree that it could be a normal method and doesn't have to overload the `()` operator (was there before my time).

saper:
> I probably need the seperation once I implement the correct memory
> handling (scoping). At some point we may move items from an inner
> scope to the global scope. But I agree that it could be a normal
> method and doesn't have to overload the `()` operator (was there
> before my time).

Uh, sounds complicated! I think we still have possibilities
to simplify and improve memory handling in the existing code.

saper:
My very early attempt at better scoping of the heap memory is in #1386 for discussion.

mgreter:
Hope you don't mind but I've opted to leave it separated for now and merged the other fix to revert to as it was before my mistake. Not sure if you have seen https://github.com/sass/libsass/pull/1335 which implementes memory scoping. So memory handling is on the road-map, but I think there needs more base work to support it properly.

saper:
Sure, I didn't know where `operator()` was supposed to be used, so reverting the incorrect change is fine as well.

PR #1383, opened 2015-07-25
Fix memory manager not tracking memory

mgreter:
This is the alternative fix to https://github.com/sass/libsass/pull/1382 that keeps the allocation and adding "memory" to the "pool" separated. The bug got introduced in a [previous refactoring] [1] (I wrongly removed the call to `operator()`).

[1]: https://github.com/sass/libsass/commit/451980e52fdd11598cb2877f482698b049c3960e#diff-cb34a4b7c469c2c15b7102db29e0daf2L48

### DIFF
--- a/src/memory_manager.hpp
+++ b/src/memory_manager.hpp
@@ -33,7 +33,7 @@ namespace Sass {
 
 template <typename T>
 inline void* operator new(size_t size, Sass::Memory_Manager<T>& mem)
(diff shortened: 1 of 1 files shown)


saper:
Side note: it seems that some test is failing on travis from time to time. Can we run it more verbosely to find out which one is this?

xzyfer:
Thanks for surfacing this @saper

PR #1385, opened 2015-07-25
Fix mac CI issues

mgreter:
~~When a signal aborts the blocking read, we should re-start the function (reentrant). Since we are a library, this can happen due to signals setup by implementor. Not sure why Travis CI on Mac exhibits this behavior.~~


xzyfer:
Please add a description to this issue @mgreter

PR #1386, opened 2015-07-26
[DNM][WIP][Experimental] Manage dependent objects w/destructors

saper:
This is just a test for a better, more localized memory
management. Basically in the parse tree "parent" object
should "own" dependent objects. Locally scoped objects
should be attached to Environment (probably).

This series of commits frees 170+ object instances
created on startup of a very simple Sass script
with a custom importer. For example class Parser
instances used to prepare native functions are
freed after they are no longer needed.

This series "leaks memory" by not freeing the
objects attached to a global Context and
Memory Manager (so it is easier to detect, what
has not been destructed yet).

Parser frees its block_stack
Environment frees its local_frame_
Vectorized types free their elements_
Definitions free their parameters_ (which are
Vectorized). 
Each Parameter frees its default_value_ (if available).

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1856,5 +1856,10 @@ namespace Sass {
     return message();
   }
 
+  Definition::~Definition()
+  {
+    delete parameters_;
+  }
+
 }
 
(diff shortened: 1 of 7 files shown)


mgreter:
Closing this since https://github.com/sass/libsass/pull/2222 has been merged. We now use a reference counted memory model, where objects get deleted once nobody keeps hold on to it. This allows for a flexible and elegant memory management, with dynamic life cycles.

PR #1388, opened 2015-07-26
Fix a few MSVC warnings

mgreter:
Fixes the warnings about double to int conversion (allow double to be passed for color maps)!


xzyfer:
Thanks @driekus77

PR #1390, opened 2015-07-26
Hotfix for block comment parsing (dirty)

mgreter:
Fixes https://github.com/sass/libsass/issues/1294

To fix this proberly a String refactoring is needed and this PR should show why. IMHO we need to find the places where we have to apply which behavior. I guess we're getting closer, but still have some basics wrong. Nut sure if we want to ship this PR, as it's a pretty dirty hack. But maybe it helps us identify which strings need what behavior. Since the issue it fixes is not critical, I leave this open for comments.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -1381,8 +1381,8 @@ namespace Sass {
   protected:
     size_t hash_;
   public:
-    String_Constant(ParserState pstate, string val)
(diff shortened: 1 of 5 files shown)


xzyfer:
I also took a look at this issue last week. I agree we need to re-refactor strings for this to work.

As you say this issue is non-critical, and big corner case. I vote for not merging this, until we fully understand the problem.

mgreter:
IMO this could help understand the problem, but I repect your call!

PR #1381, opened 2015-07-25
Fixing Warnings with MSVS 2015

driekus77:
In this commit/push I reversed previous pull request changes and fixed only the warnings as pointed out by the MSVS 2015 C/C++ compiler.

Code analysis issues will follow soon in another request.

PR #1391, opened 2015-07-26
Fix a few obvious code issues

mgreter:
Fixing a few obvious minor code issues found by @driekus77 in https://github.com/sass/libsass/pull/1381


xzyfer:
Thanks @driekus77

PR #1395, opened 2015-07-27
Fix mismatch in function signature in json.cpp

xzyfer:
This PR fixes a function signature mismatch introduced in https://github.com/sass/libsass/pull/1373.

Originally mentioned in https://github.com/sass/libsass/pull/1373/files#r35503582

### DIFF
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -357,8 +357,8 @@ static void to_surrogate_pair(uint32_t unicode, uint16_t *uc, uint16_t *lc)
   *lc = (n & 0x3FF) | 0xDC00;
 }
 
-static bool is_space        (const char& c);
-static bool is_digit        (const char& c);
+static bool is_space        (const char *c);
+static bool is_digit        (const char *c);
 static bool parse_value     (const char **sp, JsonNode        **out);
 static bool parse_string    (const char **sp, char            **out);
 static bool parse_number    (const char **sp, double           *out);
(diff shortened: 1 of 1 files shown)


src/json.cpp
```diff
@@ -357,8 +357,8 @@ static void to_surrogate_pair(uint32_t unicode, uint16_t *uc, uint16_t *lc)
   *lc = (n & 0x3FF) | 0xDC00;
 }
 
-bool is_space        (const char *c);
-bool is_digit        (const char *c);
+bool is_space        (const char& c);
+bool is_digit        (const char& c);
```
mgreter: Why are they not static as the others?

xzyfer: Yep, updated. I'm working through some inconsistencies in this as an experiment.

xzyfer:
I've added a second commit which makes the these methods more consistent with the rest of the file. I have not strong feelings about this particular change since this is essentially a 3rd party library.

However the lack of attribution makes it hard to track back upstream so we essentially own this now and internal consistency would be nice.

drewwells:
:+1: Odd that I was getting some build failures with `static bool` on some OSes. I guess the changes to how the arguments are dereferenced fixed that.

I wasn't sure how much we wanted to change this, but makes sense to own this as it is. Unless we want to locate a different json lib for this tiny portion of the API it is involved in :)

saper:
Could this cause https://github.com/sass/libsass/issues/1415 ? hard to believe

PR #1410, opened 2015-07-28
fix build errors on windows

drewwells:
This addresses a build error occurring on Windows. 

```
units.cpp: In function 'Sass::SassUnit Sass::string_to_unit(const string&)':
units.cpp:87:25: error: return-statement with no value, in function returning 'Sass::SassUnit' [-fpermissive]
     else if (s == "in") return IN;
                         ^
units.cpp: In function 'const char* Sass::unit_to_string(Sass::SassUnit)':
units.cpp:116:14: error: expected primary-expression before ':' token
       case IN: return "in"; break;
              ^
```

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -1,6 +1,10 @@
 #include <stdexcept>
 #include "units.hpp"
 
+#ifdef IN
+#undef IN
+#endif
+
 namespace Sass {
 
   /* the conversion matrix can be readed the following way */


xzyfer:
I think it be better to fully namespace the case conditions. Thoughts?

drewwells:
I got the same error with `Sass::IN`. I'm only seeing these on Windows so my ability to debug this is really slow (appveyor).

```
In function 'Sass::SassUnit Sass::string_to_unit(const string&)':
units.cpp:72:40: error: expected unqualified-id before ';' token
     else if (s == "in") return Sass::IN;
                                        ^
units.cpp: In function 'const char* Sass::unit_to_string(Sass::SassUnit)':
units.cpp:101:20: error: expected unqualified-id before ':' token
       case Sass::IN: return "in"; break;
                    ^
```

xzyfer:
I was thinking something like this https://github.com/sass/libsass/pull/1411/files. It passes AppVeyor. Does it fix your issue?

drewwells:
Nope, still seeing the same error

```
units.cpp: In function 'Sass::SassUnit Sass::string_to_unit(const string&)':
units.cpp:72:46: error: expected unqualified-id before ';' token
     else if (s == "in")   return SassUnit::IN;
                                              ^
units.cpp: In function 'const char* Sass::unit_to_string(Sass::SassUnit)':
units.cpp:101:24: error: expected unqualified-id before ':' token
       case SassUnit::IN:      return "in"; break;
                        ^
```

xzyfer:
Ahh bugger. Something about undef random things makes me uneasy, but it seems pretty standard.

drewwells:
It appears a windows header includes this (I guess that's obvious). Not sure of a better way to get around this though :\

xzyfer:
Yeah I'm not sure of a work around either. I guess we'll ship it and see how it goes. Could you please add a comment above your change so we know why this exists. Please squash the commit.

drewwells:
Ahh, here's the culprit: http://sourceforge.net/p/mingw-w64/code/HEAD/tree/stable/v3.x/mingw-w64-headers/include/minwindef.h#l57

drewwells:
Might have found a better way to handle this, trying it out locally first.

xzyfer:
Legend!

drewwells:
This worked for me! Sort of the same thing, but removes all fake modifiers mingw32 creates.

Side not, very little on the internet about the purpose of these: http://lmgtfy.com/?q=W32_PSEUDO_MODIFIERS

xzyfer:
Do you still need both commits?

drewwells:
It only needs the define.

I wanted to get your thoughts on that before I rebased the other out. The downside here is that this may be a race condition. `_NO_W32_PSEUDO_MODIFIERS` would need to be defined before windows headers are loaded. The first commit will just work no matter the order.

xzyfer:
You're right about the race condition, which isn't great. Now that we understand the origin of conflict I'm fine with your original commit with a comment and link back to the source of the define.

As an aside do you know why AppVeyor wasn't failing on this?

drewwells:
Because `windef` isn't included units.cpp. This error will pop up if units.cpp is included with anything that includes windows headers ie. file.cpp, functions.cpp or plugins.cpp.

Going to clean up the branch and describe the problem better in the commit, then should be ready!

xzyfer:
Thanks @drewwells!

saper:
Does #1310 fix the issue without the #undef?

drewwells:
No, it has to be #undef or not #define in the first place. #undef is a cleaner approach though: https://github.com/sass/libsass/pull/1410#issuecomment-125458730

saper:
> No, it has to be #undef or not #define in the first place. #undef is a cleaner approach though: https://github.com/sass/libsass/pull/1410#issuecomment-125458730

I asked because Solaris #defines SEC and we can fix it by not polluting our namespace with "std".
Would that work for your case as well?

the min() max() windows problem needs to be solved by #define NOMINMAX though - they provide
the possibility to tell windows.h "don't define these". By the way they are defined
prevent me to cure the situation using namespaces.

Can you try #1310 and see if it helps? (Without the undef).

Which compiler are you using?

drewwells:
#1310 doesn't address the use of `IN` and #include <stdlib> doesn't define IN.

`undef` approach is used because #define _NO_WIN32_PSEUDO_MODIFIERS would have to be added in many places and their define serves no purpose for libsass. https://github.com/sass/libsass/pull/1410#issuecomment-125453374

drewwells:
If namespacing was an issue, https://github.com/sass/libsass/pull/1411 would have fixed it (but doesn't).

saper:
Well, there is no build error on our code as-is (just checked with VS2013). This is a consequence of combining all our files into one, which is not a supported build configuration. Unfortunately, Win32 API is full of funny symbols and one of the reasons to split files is to isolate affected files. 

I'd recommend adding an additional file with a workaround code in the additional header file included as the first one, prior to merging the files. Or somewhere in between.

`#define _NOWIN32_PSEUDO_MODIFIERS` does not work Visual Studio, unfortunately `#define IN` is embedded in many core headers and it is impossible to somehow `#define` it out

Having said that, I'm happy to work on streamlining our header files to make compilation process faster, but I don't think we should support that mode of compilation.

drewwells:
Are you experiencing build errors as a result of these changes? If so,
could you reproduce it on the appveyor test suite because all builds are
passing.

saper:
> Are you experiencing build errors as a result of these changes? If so,
> could you reproduce it on the appveyor test suite because all builds are
> passing.

The build is fine with the code as-is (from git master).

If I add #include <windows.h> to "units.cpp" it breaks
because of #define IN in many Windows headers (windef.h,
ncrypt.h, bcrypt.h, wincrypt.h and many others).

You can try this I think also with MinGW32 if you point
to to a full Windows API (as opposed to a stripped down
package MinGW32 provides).

But #undef IN after #include <windows.h> fixes the issue.

drewwells:
Personally, I think we should not use `IN`, `OUT`, `OPTIONAL` because those have special meaning in windows. I initially fixed this by renaming `IN` to `INCH`, but if you give windows an inch it'll take a mile :)

saper:
> Personally, I think we should not use `IN`, `OUT`, `OPTIONAL` because
> those have special meaning in windows. I initially fixed this by
> renaming `IN` to `INCH`, but if you give windows an inch it'll take a
> mile :)

That's reasonable. Windows hijacks lots of words, but we've
had issues with Solaris as well.

PR #1423, opened 2015-07-31
Fixing current build issues on Windows 64 bits with MSVC 12/2013

driekus77:
Hopefully I did everything right this time ;-)

Checkout out current libsass did resolve in build errors and warnings. I did my best to fix them and the result is this PR. 

Only build on Win x64 so other builds have to be checked. I will do this if Appveyor results in errors.

Kind regards,

Henry

### DIFF
--- a/include/sass.h
+++ b/include/sass.h
@@ -5,9 +5,19 @@
 
 #ifdef _MSC_VER
   #pragma warning(disable : 4503)
+
+#ifndef _SCL_SECURE_NO_WARNINGS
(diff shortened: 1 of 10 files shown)


drewwells:
Something is very appealing about this PR...!

saper:
I have submitted https://github.com/sass/sassc/pull/129 to apply the same to sassc (which is used also when building libsass on win32)

xzyfer:
This is great @driekus77! I wonder how we can make AppVeyor pick these issues up for us?

PR #1430, opened 2015-08-10
Return Null value for & if selector stack is empty

### LINKED ISSUE #1415: std::bad_typeid when checking parent selector on base level
This looks like a regression in master

Given the following code:

```
@mixin prepend-foo {
    $parent: &;

    @if $parent {
        .foo & {
            @content;
        }
    } @else {
        .foo {
            @content;
        }
    }
}

@include prepend-foo {
    .bar {
        color: red;
    }
}
```

Ruby Sass:
```
$ sass --version
Sass 3.4.16 (Selective Steve)
$ sass test.scss
.foo .bar {
  color: red; }
```

libsass 3.2:
```
$ sassc --version
sassc: 3.2.1
libsass: 3.2.5
sass2scss: 1.0.3
$ sassc test.scss
.foo .bar {
  color: red; }
```

libsass master:
```
$ sassc --version
sassc: 3.2.5-11-g3565
libsass: 3.2.5-151-g9b4e
sass2scss: 1.0.3
$ sassc test.scss
Error: std::bad_typeid
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25092788-std-bad_typeid-when-checking-parent-selector-on-base-level?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

saper:
A quick'n'dirty "fix" for https://github.com/sass/libsass/issues/1415

Fixes #1415

### DIFF
--- a/src/eval.cpp
+++ b/src/eval.cpp
@@ -1517,10 +1517,14 @@ namespace Sass {
   Expression* Eval::operator()(Parent_Selector* p)
   {
     Selector_List* pr = selector();
-    exp.selector_stack.pop_back();
-    if (pr) pr = operator()(pr);
-    exp.selector_stack.push_back(pr);
-    return pr;
+    if (pr) {
+      exp.selector_stack.pop_back();
+      pr = operator()(pr);
+      exp.selector_stack.push_back(pr);
+      return pr;
+    } else {
+      return new (ctx.mem) Null(p->pstate());
+    }
   }
 
 }


xzyfer:
Seems legit to me. It's what I would have done.

src/eval.cpp
```diff
@@ -1520,7 +1520,10 @@ namespace Sass {
     exp.selector_stack.pop_back();
     if (pr) pr = operator()(pr);
     exp.selector_stack.push_back(pr);
-    return pr;
+    if (pr)
+      return pr;
+    else
+      return new (ctx.mem) Null(p->pstate());
```
xzyfer: Style nitpick.

Remove new lines

```    
if (pr) return pr;
else return new (ctx.mem) Null(p->pstate());
```

or add braces

```
if (pr) {
  return pr;
} else {
  return new (ctx.mem) Null(p->pstate());
}
```

xzyfer:
Please also add this patch @saper 

```
diff --git a/src/functions.cpp b/src/functions.cpp
index 099948d..5ae9ba2 100644
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -1651,10 +1651,14 @@ namespace Sass {
       if( arglist->length() == 0 )
         error("$selectors: At least one selector must be passed", pstate);

+
       // Parse args into vector of selectors
       vector<Selector_List*> parsedSelectors;
       for (size_t i = 0, L = arglist->length(); i < L; ++i) {
         Expression* exp = dynamic_cast<Expression*>(arglist->value_at_index(i));
+        if (exp->concrete_type() == Expression::NULL_VAL) {
+          error("$selectors: null is not a valid selector: it must be a string,\na list of strings, or a list of lists of strings for `selector_append'", pstate);
+        }
         string exp_src = exp->perform(&to_string) + "{";
         Selector_List* sel = Parser::parse_selector(exp_src.c_str(), ctx);
         parsedSelectors.push_back(sel);
```

saper:
We need this(?) in almost every instance of this marvel:

````cpp
      string exp_src = exp->perform(&to_string) + "{";
      /* do something with */ Parser::parse_selector(exp_src.c_str(), ctx);

````

and there are `ARGSEL` macros that might be helpful and the right place to do this...

xzyfer:
Leave the error for now. I've got a separate patch.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/468

PR #1431, opened 2015-08-10
Better mirror Ruby Sass errors for null params to selector functions

xzyfer:
This PR updates the selector functions to throw errors if a `null` argument is supplied.

Originally mentioned in https://github.com/sass/libsass/pull/1430#issuecomment-129283759

Fixes https://github.com/sass/libsass/issues/1432

/cc @saper

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -71,6 +71,12 @@ namespace Sass {
                                     false, true);
   }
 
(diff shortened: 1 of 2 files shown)


xzyfer:
Spec added https://github.com/sass/sass-spec/pull/469

PR #1436, opened 2015-08-12
Check for mingw32 target, not host

saper:


xzyfer:
Thanks as always @saper

Issue #1360, opened 2015-07-20
Move C-API library headers into subfolder


mgreter:
Pretty much all libraries install their headers in a subdirectory. We currently just drop multiple headers into the include folder directly (`/usr/include` on most systems). IMO we should change this and put all of our C-API headers in a `libsass` sub-folder (`/usr/include/libsass`).

Since this is actually a breaking change for implementors, I wanted to discuss this here. The update is very simple, since you really only need to change all include from `#include "sass.h"` to `#include "libsass/sass.h"` (and for all other C API headers).

I would really like to ship this change also with 3.3, since implementors already have to adjust their builds, so both needed changes could be done in one go!

drewwells:
Sounds good, I fear leaving around outdated headers for libsass all the
time. Would be nice to have them all in one place

saper:
Requiring

     #include <libsass/sass.h>

instead of

     #include <sass.h>

is an API change. (I know you can cheat with `-I`, but that's still breaking the API).

mgreter:
Yep, breaking change, that's why I want to include it in 3.3, otherwise we have to wait for 3.4. And we are not following semver for now, but we try to add only breaking changes with bigger version increments!

saper:
Ah, in that case it's fine of course.

mgreter:
I will do this once the other PRs are merged, too much moving right now!

xzyfer:
:+1:

xzyfer:
@mgreter did you want to get this into 3.3.?

mgreter:
After looking a bit more into this task I think we should do this when we remove obsolete `sass_interface.h`. So the question is do we want to remove it for 3.3 or wait to 3.4?

xzyfer:
Lets wait for 3.4. I'm keen to feature freeze 3.3 ASAP

am11:
@mgreter, I totally respect the decision made by @xzyfer. Just an opinion; 

* 3.3 is going to be a **HUGE** release of libsass
* Many major downstreams are already using new API
* `sass_interface` is already deprecated for a while now.

Given this situation, wouldn't it make sense to forgo `sass_interface` for 3.3? We can always offer help to someone who is asking. Last I remember, .NET wrapper was using it and it was also upgraded..

mgreter:
I agree and I would be in favor of dropping it for 3.3 ...

xzyfer:
My concern was purely with regards to getting a feature freeze so we can get 3.3 RC out asap.

[...]

PR #1439, opened 2015-08-12
win: Uses VS2015 for CI builds

am11:
VS2015 is now available for our package: http://www.appveyor.com/blog/2015/08/02/visual-studio-2015-rtm.

@mgreter, also removed unused `snprintf` and `vsnprintf` unused declarations. Besides VS2015 supports C99 `snprintf` and friends (and also supports Windows XP), so IMO there is no point of not daringly using C99 string formatters without fallbacks :grey_exclamation:  :smile:


saper:
Oh it seems that telling MSBuild which Microsoft Visual C++ version we are using is not so easy, so what about `#ifdef _MSC_VER < 1900` instead?

xzyfer:
I was wondering the same as @saper. Is there any reason to not support both somehow? 
Should we expect people to be on the latest hotness to be able to build LibSass?

am11:
@saper, @xzyfer, those are *unused* functions! :)
Otherwise, it is rather easy to version MSVCR version with this condition: `defined(_MSC_VER) && _MSC_VER < 1900`.

QuLogic:
Uh, `snprintf` is used right [here](https://github.com/am11/libsass/blob/master/src/json.cpp#L1326).

am11:
Ah, missed that one hidden under macro! updated the conditional statement.

win/libsass.vcxproj
```diff
@@ -1,17 +1,12 @@
...
     <SASSC_VERSION>[NA]</SASSC_VERSION>
     <LIBSASS_VERSION>[NA]</LIBSASS_VERSION>
   </PropertyGroup>
-
   <Target Name="GitVersion">
-
-    <Exec Command="git -C .. describe --abbrev=4 --dirty --always --tags > libsass-version" />
-    <Exec Command="git -C ../sassc describe --abbrev=4 --dirty --always --tags > sassc-version" />
-
+    <Exec Command="git -C .. describe --abbrev=4 --dirty --always --tags &gt; libsass-version" />
```
xzyfer: `&gt;` is that right?

xzyfer:
Aside from my comment this is mostly whitespace changes so I'm cool with it.

saper:
Need to check if 14.0 tools version breaks lodem msbuild from 2013

am11:
@saper, tested with bother VS2013 and VS2014 and everything was green. :)
Did some digging and found that the ToolsVersion in first line of vcxproj file is irrelevant in native C++ projects when we have `Microsoft.Cpp.*.props` imported (as we have; with that tools version gets overridden and also I think msbuild only considers it for .NET projects only if no further tools version is defined).
One more interesting thing I found during this exercise and commented here (the CLI tip): http://stackoverflow.com/questions/19801072/the-builds-tools-for-v120-platform-toolset-v120-cannot-be-found#comment51867426_19801072 <- the command line argument has the highest precedence. Also we can set environment variable to select tooling for all projects in the given environment.

win/libsass.vcxproj
am11: Yep. It actually prefers escaped strings in the params and values.  :)
I verified it as followed, (please let me know if that is the correct way to verify it):

```powershell
C:\users\adeel\Source\Repos\libsass [master +2 ~0 -0 !]> .\win\bin\Debug\sassc.exe -v
sassc: 3.2.5-14-g5c78
libsass: 3.1.0-603-gc380
sass2scss: 1.0.3
```

xzyfer: You'd know better than I would ;)

am11: Actually, it was automagically escaped by VS when I upgraded the project. Seems this is the legit way for executable commands: http://stackoverflow.com/a/2414096/863980 & https://msdn.microsoft.com/en-us/library/ms228186.aspx although `>` is not one of the special character which requires escaping; going by [their chart](https://msdn.microsoft.com/en-us/library/bb383819.aspx). :D

mgreter: This is just how XML works and IMO has nothing to do with VS.

am11: true! :+1:

xzyfer: Yep, I was just curious about meaningless noise diffs.

saper:
@am11 what do you think about this :) tested only with my VC2013 https://github.com/am11/libsass/pull/1 Can you see if it works with VC2015 (should skip compiling `c99func.c` completely).

am11:
@saper, that's pretty neat! :+1: 
I have left a comment (without the fix, it was complaining when I had VS2013 tooling selected). 

@xzyfer, @mgreter, do you guys like that (separation of concerns) approach by took by @saper?

win/libsass.vcxproj
am11: @xzyfer, I totally agree about the noise. It is just auto-formatting by VS, once we do that it won't get changed in future (just the end of file extra linefeed it omits in some cases, which we should keep for `git`). With that said, should I revert the extra changes?

xzyfer: Nah it's cool. It just stood out.

am11:
@xzyfer, @mgreter, changes by @saper are incorporated via 574b3941559d7759e6b417b607efc67e97047536. PTAL.

@mgreter, on a separate note, when  we build on win, the process leaves behind two files `win/libsass-version` and `win/sassc-version`. Should we delete them if they are unused? I am not sure if they are required afterwards, if they do we can add them those two paths to `.gitignore`. Please advise.

mgreter:
@am11 are they overwritten/updated on consecutive compiles? If so we could leave them and add them to .gitignore. If they stay the same once they got generated, they should be cleaned up. Reason is that they should be updated if you add a commit and re-compile!

am11:
@mgreter, it was updating for all consecutive builds (with/without cleanup), so I have added the rule in `.gitignore` via e0e23e0.

saper:
> @mgreter, it was updating for all consecutive builds (with/without cleanup), so I have added the rule in `.gitignore` via e0e23e0.

One idea: can we easily move them in "bin/{Debug|Release}" folder? I thin that's where they really belong.

This way the source directory can stay read-only all the time (we need this for UNIX as well btw.)

am11:
@saper, I have deleted the files after it sets the variable: 49f5886.

Tested with both VS2013 and VS2015 and `win/bin/Debug/sassc -v` prints correct versions in both cases.

@mgreter, it turned out we don't need to both `.gitignore` for this one.

saper:
We can get rid of those files btw .:) patch pending.

saper:
Here it is https://github.com/am11/libsass/pull/2 MSBuil 12.0 required, but no temporary files! Tested with VS2013

xzyfer:
I'm happy to ship this when the Windows folk (@am11, @saper) give the go ahead

am11:
This is tested and ready.

xzyfer:
Looks like you're still working on this @am11. Let me know when you're finished and the commits are cleaned up :)

am11:
ouch! I thought you are AFK so I can mess around a bit with appveyor.yml :wink:

xzyfer:
Go nuts. just catching up on email! At the very least this release is blocked on https://github.com/sass/libsass/pull/1447 so you've got time. 

I'll be AFK for the next 24hr(ish)

am11:
@xzyfer, this will fail as it depends on https://github.com/sass/sassc/issues/132. Please first merge sassc PR, then I'll rerun the CI build. Once CI gets green again, this is ready for merging.

Latest changes made in this PR:

* `libsass\win\libsass.sln` now produces dynamic lib `libsass.dll` (or static `libsass.lib`) as it should. Currently it is producing sassc.exe and compiling sassc. To make it produce static lib, set environment variable `LIBSASS_STATIC_LIB` and unset to produce dll again.
* `libsass\sassc\win\sassc.sln` now produces `libsass\sassc\bin\sass.exe` as it should and AppVeyor CI will use it.
* Both projects shared include files via `libsass.targets`, so we don't need to add sources in multiple projects anymore (for example, at present (without https://github.com/sass/sassc/issues/132), sassc/win/sassc.sln is not compiling due to the stale include references..)

saper:
One idea: why not use one variable (`SASS_LIBSASS_PATH` - the same one as Travis CI is using) and just use directories relative to that `src` and `include`?

am11:
@saper, I used `LIBSASS_HEADERS_DIR` and `LIBSASS_SRC_DIR` in `.targets`, `.vcxproj.filters` and `.vcxproj` files after reading @mgreter's proposal: https://github.com/sass/libsass/issues/1360. At the moment, they have same values in `[libsass/sassc].vcxproj`, but the point was to make the project future refactor compatible. I also used `LIBSASS_DIR` and `LIBSASS_STATIC_LIB` (user-defined env var which is used in libsass.vcxproj).

It will definitely be a good idea to align these build-time variables naming convention across the board (including those we used in node-sass), but that will probably require major overhaul covering gyp, Makefile and vcxproj files in libsass and node-sass. IMO we should have an RFC-issue tracking that kind of cleanup in libsass issue tracker.

xzyfer:
@am11 can you also please update the building libsass wiki page for windows?

am11:
@xzyfer, sure. I intend to updated the wiki once this is merged. :)

xzyfer:
@am11 go fourth and assume it'll be merged :) 

At this point I'm mostly waiting for docs (https://github.com/sass/sassc/pull/132#issuecomment-131622092) or until I'm back at my computer. Which ever comes first. Hopefully both :)

xzyfer:
@am11 This still after sass/sassc#132 has been merged

saper:
Shouldn't we maybe split VC2015 integration (which is done and works) and sassc/libsass divorce into a separate PR?

xzyfer:
@saper I like the sound of that.

xzyfer:
Frankly if this is working as expected I'm happy to get it shipped. @am11 has put in a lot of great work here and a split although nice, is nitpick a this point IMHO.

am11:
@xzyfer, this is ready as is. The there was a sassc<->sass typo and it actually works: MSVC builds passed. 
Do you still want me to refactor the commits? :)

xzyfer:
@am11 I saw it was building in CI so I assumed you were still working on it. If you think you can clean the commit history it more logically then yes please.

I do see that one of the commits failed to compile, which although I don't use Windows for debugging, is not something we'd generally allow because it makes bisecting more difficult.

saper:
> Frankly if this is working as expected I'm happy to get it shipped.
> @am11 has put in a lot of great work here and a split although nice,
> is nitpick a this point IMHO.

I agree - if we get it working, that it may be combined. Of course.

am11:
@xzyfer, I have rebased it so all appveyor.yml changes are in the first commit. Then the order follows which retain @saper's commits and the final VC solution refactoring goes in bedefc3 which conforms with the changes made in sassc. It can be further squashed down to one multi-authors commit for tidiness. 
Let me know if there is anything. 
Aside: I will update the [wiki page](https://github.com/sass/libsass/wiki/Building-with-Visual-Studio) further and send a separate PR for sassc Readme later.

xzyfer:
4a8b1f3209257c94369ea8f15de8d577

mgreter:
@am11 thanks for this nice addition, looks like more and more people are actually using MSVC since it has become freely available. I have only one minor nitpick with the documentation:

```
- "X:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild" libsass\win\libsass.sln
+ "%programfiles(x86)%\MSBuild\12.0\Bin\MSBuild" libsass\win\libsass.sln
```

I guess this way the line would be better for copy & paste and I hope that people will understand what they need to change if they installed VS in another path (is that even possible?).

am11:
@mgreter, thanks for the suggestion. I have simplified the commands in wiki using environment variables: `%ProgramFiles%` and `%ProgramFiles(x86)%` and their PowerShell variants `${env:ProgramFiles(x86)}` and `$env:ProgramFiles`. Also upgraded the sassc PR accordingly: https://github.com/sass/sassc/pull/133.

> what they need to change if they installed VS in another path (is that even possible?).

During the installation, we can change the install directory in VS, but (I think) MSBuild is always installed at `%ProgramFiles(<process-arch>)%/MSBuild`. Otherwise it needs to be located via registry: http://stackoverflow.com/q/328017 and so forth..

PR #1446, opened 2015-08-15
MemoryManager: Use free instead of delete.

### LINKED ISSUE #1443: crash on invalid input when compiled with Visual Studio
When I compile libsass + sassc with Visual Studio 2013 (using the vcxproj included with libsass), and the input CSS source contains errors, sassc is crashing. 

My input is simply:
`foo { color: rgba(10,10,10,10); }`

First an exception is thrown at error_handling.cpp:36, but immediately after this there is an invalid pointer operation in memory_manager.cpp:70 (Memory_Manager<T>::destroy, deleting np).

	sassc.exe!Sass::Memory_Manager<Sass::AST_Node>::destroy(Sass::AST_Node * np=0x004bac98) Line 70	C++
 	sassc.exe!operator delete<Sass::AST_Node>(void * np=0x004bac98, Sass::Memory_Manager<Sass::AST_Node> & mem={...}) Line 38	C++
 	sassc.exe!_GetSystemTimeAsFileTime@4()	C++
 	sassc.exe!Sass::Eval::operator()(Sass::Function_Call * c=0x0049b348) Line 644	C++
 	sassc.exe!Sass::Function_Call::perform(Sass::Operation<Sass::Expression *> * op=0x0038f128) Line 1084	C++
 	sassc.exe!Sass::Expand::operator()(Sass::Declaration * d=0x0049a9e8) Line 207	C++
 	sassc.exe!Sass::Declaration::perform(Sass::Operation<Sass::Statement *> * op=0x0038f120) Line 497	C++
 	sassc.exe!Sass::Expand::append_block(Sass::Block * b=0x0049a758) Line 587	C++
 	sassc.exe!Sass::Expand::operator()(Sass::Block * b=0x0049a758) Line 79	C++
 	sassc.exe!Sass::Block::perform(Sass::Operation<Sass::Statement *> * op=0x0038f120) Line 362	C++
 	sassc.exe!Sass::Expand::operator()(Sass::Ruleset * r=0x00498458) Line 104	C++
 	sassc.exe!Sass::Ruleset::perform(Sass::Operation<Sass::Statement *> * op=0x0038f120) Line 392	C++
 	sassc.exe!Sass::Expand::append_block(Sass::Block * b=0x00498290) Line 587	C++
 	sassc.exe!Sass::Expand::operator()(Sass::Block * b=0x00498290) Line 79	C++
 	sassc.exe!Sass::Block::perform(Sass::Operation<Sass::Statement *> * op=0x0038f120) Line 362	C++
 	sassc.exe!Sass::Context::parse_file() Line 324	C++
 	sassc.exe!sass_parse_block(Sass_Compiler * compiler=0x00499c70) Line 504	C++
 	sassc.exe!sass_compiler_parse(Sass_Compiler * compiler=0x00499c70) Line 655	C++
 	sassc.exe!sass_compile_context(Sass_Context * c_ctx=0x00497838, Sass::Context::Data cpp_opt={...}) Line 537	C++
 	sassc.exe!sass_compile_file_context(Sass_File_Context * file_ctx=0x00497838) Line 642	C++
 	sassc.exe!compile_file(Sass_Options * options=0x00497428, char * input_path=0x0048b897, char * outfile=0x00000000) Line 108	C
 	sassc.exe!main(int argc=2, char * * argv=0x0048b858) Line 284	C
 	sassc.exe!__tmainCRTStartup() Line 626	C
 	sassc.exe!mainCRTStartup() Line 466	C
 	kernel32.dll!@BaseThreadInitThunk@12()	Unknown
 	ntdll.dll!___RtlUserThreadStart@8()	Unknown
 	ntdll.dll!__RtlUserThreadStart@8()	Unknown

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25708792-crash-on-invalid-input-when-compiled-with-visual-studio?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

am11:
On instantiating `Color` at src/functions.cpp (`BUILT_IN(rgba_4)` body)
, `Memory_Manager`'s `operator new()` is called which after explicitly
calling the underlying `operator new(size_t)`, `static_cast` the
allocated memory for `void*` -> `AST_Node*` conversion.

The `delete` operator in `Memory_Manager` is called when an exception
is thrown during the construction of the object (right after the `new`)
. When calling underlying `delete` from custom `delete` implementation,
it throws access violation probably because it tries to `delete` all
continents of objects which weren't initialised properly, as in case of
invalid rgba value, `error_handling` layer throws.

Therefore, more rawer `free()` is used to avoid `bad_alloc`<->`delete`
panics. :)

Note that this is the quick fix to the problem. The actual fix would be
to implement a custom exception-tolerant `New`.

Read more on this subject:
http://www.scs.stanford.edu/~dm/home/papers/c++-new.html.

Fixes #1443.

### DIFF
--- a/src/memory_manager.cpp
+++ b/src/memory_manager.cpp
@@ -67,7 +67,7 @@ namespace Sass {
     // remove from pool
     remove(np);
     // release memory
-    delete np;
+    free(np);
   }
 
   // compile implementation for AST_Node


xzyfer:
Couldn't we just assign the constructor arguments to locally scoped variables before calling the `Color` constructor? That way the error is thrown before we construct.

As as aside the we shouldn't error in this case at all as mentioned in https://github.com/sass/sass-spec/pull/472#issuecomment-131481610

am11:
This is the fix for regression. The throw behavior can be addressed separately IMO.

> Couldn't we just assign the constructor arguments to locally scoped variables before calling the  Color  constructor? 

Good idea. Another good idea is probably to *always* finish off with `free` rather than `delete` in a custom `delete` operator implementation as we have more control over memory than we should. :) So IMO this fix is necessary in any case. One obvious question is "why it is not throwing elsewhere?". AFAICT, this is the only path to `destroy()` function, but it is unwarranted because I haven't debugged with the entire sass-spec suite but only a couple dozens. ;)

am11:
> Couldn't we just assign the constructor arguments to locally scoped variables before calling the Color constructor? 

On this note again; as you indicated previously that the Ruby Sass behavior is to clamped $alpha>1 to 1, $alpha<0 to 0, $[r/g/b] > 255 to 255 and $[r/g/b]<0 to 0. IMO refactoring those functions in `functions.cpp` does not make any sense when the ultimate aim is to comply with Ruby Sass' no-throw clamping behavior.

xzyfer:
I have refactored the colour functions in #1449.

xzyfer:
@am11 should https://github.com/sass/libsass/pull/1446/files#diff-3f8f7ba253c6298d2c1de8bfdd63c389R53 also be updated to `free`?

saper:
Using `free()` in the normal case causes destructors not to be called. This is not good, given especially given we want to free memory as we go and not only when shutting down. I'd suggest determining object ownership and doing something like #1386 in the mid-term, unless there is a better idea.

am11:
> I have refactored the colour functions in #1449.

:+1: 

> @am11 should https://github.com/sass/libsass/pull/1446/files#diff-3f8f7ba253c6298d2c1de8bfdd63c389R53 also be updated to  free ?

Sort answer: No
Long version: in theory (well and in my understanding):
* If memory was allocated (in C-style) `malloc`'d in ctor, then it should be `free()` in dtor.
* If the memory was `new`'d then we have the C++ reference tracking and `delete` would work.
* In case there is a chance of *ctor-went-wrong-after-successful-memory-allocation-with new* (for instance some exception was thrown including `bad_alloc` (which can be avoided by C++11's new-with-nothrow)), then one should have a custom `operator delete` implementation (like we have) which should finalize with `free()`.

Other than this, I think the `MemoryManager` implementation is pretty solid IMO! (well until some other creepy bug showed up on the radar :smile:)

xzyfer:
Thanks for the clarification @am11. I'll need to school up on memory managers.

saper:
Oh noes, but this patch makes the memory manager to always use `free()` ! Not only on some crazy constructor failure....

xzyfer:
@saper I agree. This is a rare edge case. The one known occurrence has been patched. I've reverted this in https://github.com/sass/libsass/pull/1450.

am11:
> Not only on some crazy constructor failure

@saper, have you tested it before commenting?

saper:
On Mon, 17 Aug 2015, Adeel Mujahid wrote:

> > Not only on some crazy constructor failure
> 
> @saper, have you tested it before commenting?

No? Should I check if destructors get called?

am11:
@saper, I really :heart: your inputs, but please see https://github.com/sass/libsass/pull/1450#issuecomment-131819583.

saper:
Do you know why exactly VS aborted? Did it create the object with "new" before the parameters were evaluated?

xzyfer:
From @am11 research this isn't isolated to VS 
https://github.com/sass/libsass/issues/1443#issuecomment-131463458

saper:
On Mon, 17 Aug 2015, Michael Mifsud wrote:

> >From @am11 research this isn't isolated to VS 
> https://github.com/sass/libsass/issues/1443#issuecomment-131463458

Ok, I wasn't able to reproduce it on clang, only with VS.

am11:
It is not the Windows-only issue. It was one a genuine hidden bug which gets hidden again by https://github.com/sass/libsass/pull/1449 and will emerge in some other form. The only time our/custom delete operator is called is when object construction was choking.
I was able to repro it with clang 3.6 on Ubuntu. YMMV..

saper:
Let me go back to 8127946d3f8a7c140ba705afaf0673d75a3f7186 and try.

From what I see this `destroy()` method gets never called normally.

am11:
> From what I see this  destroy()  method gets never called normally.

Yup I think so too and when it is called, the process is already in some kind of disorient state.

saper:
No, in most cases (this begins to change right now) we just don't free memory explicitly. Let me try to reproduce the bug, because just because we almost never `delete` our objects manually it does not mean we should do it this way.

am11:
@saper, maybe my knowledge about C/C++ memory management is not as good as yours as you seems to be pretty sure about it, but let me know if you find out anything interesting because I for one won't make big claims before being absolutely sure and especially before testing the situation at hand..

saper:
I think that placement new() is one of the most confusing things in C++ and I wish it will go away in libsass one day.

PR #1447, opened 2015-08-16
Color: Limit RGBA values to the boundaries

am11:
This is to conform with ruby-sass behavior.

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -245,20 +245,20 @@ namespace Sass {
     Signature rgb_sig = "rgb($red, $green, $blue)";
     BUILT_IN(rgb)
     {
-      return new (ctx.mem) Color(pstate,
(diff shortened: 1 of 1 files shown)


am11:
@xzyfer, with this change, `Color` ctor does not throw either. :)

xzyfer:
Thanks @am11. I'll wait til I have the spec suit ready. Since you've made the change in the marco and not in the Color ctor itself we need to sure that all colour functions respect this behaviour. It could be that some color functions do infact throw if the arg is out of range.

xzyfer:
I think the right solution here is to leave the marco as is and put the clamping in the `Color` ast node's ctor.

am11:
@xzyfer, I have amended the commit. PTAL.

xzyfer:
Superseded by #1449

PR #1450, opened 2015-08-17
Revert "MemoryManager: Use free instead of delete."

xzyfer:
Reverts sass/libsass#1446

### DIFF
--- a/src/memory_manager.cpp
+++ b/src/memory_manager.cpp
@@ -67,7 +67,7 @@ namespace Sass {
     // remove from pool
     remove(np);
     // release memory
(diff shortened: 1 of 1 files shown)


am11:
@xzyfer, why is this reverted?

xzyfer:
Partly https://github.com/sass/libsass/pull/1446#issuecomment-131817270, and largely because I don't understand the change enough to make the right call here.

am11:
IMO that was the right fix as I spent quite much time debugging sass-spec suite tests with this change. But your call :)

PR #1454, opened 2015-08-18
MSVC: Removes unnecessary NOMINMAX macros

am11:
* `NOMINMAX` macro is only required where we include `<windows>` header
  because it has `min` and `max` defined as macros; and we need it in
  one file: `src/file.cpp`.
* Also adds .user (VS user file) to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ bin/*
 .deps/
 .libs/
 win/bin
+*.user
 
 # Final results
 
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1,6 +1,3 @@
-#if defined(_MSC_VER)
-#define NOMINMAX
-#endif
 #include "ast.hpp"
 #include "context.hpp"
 #include "node.hpp"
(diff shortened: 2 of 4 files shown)


am11:
@mgreter, separately; can we replace `#ifdef DEBUG` with standard C `#ifndef NDEBUG` postprocessor directive, so we can get debug info with MSVC as well?

saper:
@am11 thanks for this, lgtm

am11:
> So `file.cpp` is the only one that needs it?

Yes; `<windows>` header is required in `file.cpp` for path resolution etc.

saper:
> > So `file.cpp` is the only one that needs it?
> 
> Yes; `<windows>` header is required in `file.cpp` for path resolution etc.

Please excuse my overzeal then :)

xzyfer:
Taking your word for it :)

am11:
> Taking your word for it :)

Fingers crossed! :ok_hand: :v:

saper:
@am11 Just filed https://github.com/sass/libsass/issues/1518 to deal with `-DDEBUG`

PR #1458, opened 2015-08-20
rename SIZE to LENGTH to avoid conflict with windows headers

drewwells:
The use of `SIZE` in units.cpp/hpp conflicts with the mingw32 windows header definition of SIZE: http://sourceforge.net/u/earnie/winapi/winapi/ci/f9fa22a4ecce88ad12a8371a40f00ef702ec3e1b/tree/include/windef.h#l322

Rather than messing with this definition, SIZE has been renamed to LENGTH which actually is a more accurate name. https://developer.mozilla.org/en-US/docs/Web/CSS/length Every now and then things work out like that!

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -56,7 +56,7 @@ namespace Sass {
   {
     switch (unit & 0xFF00)
     {
-      case SassUnitType::SIZE:        return SassUnitType::SIZE; break;
+      case SassUnitType::LENGTH:      return SassUnitType::LENGTH; break;
       case SassUnitType::ANGLE:       return SassUnitType::ANGLE; break;
       case SassUnitType::TIME:        return SassUnitType::TIME; break;
       case SassUnitType::FREQUENCY:   return SassUnitType::FREQUENCY; break;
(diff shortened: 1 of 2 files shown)


drewwells:
Appveyor sure is paying a lot to download mingw32 over and over. It's about time they started including 64bit version.

xzyfer:
Nice work @drewwells. Any clues as to why our CI hasn't been failing on this?

drewwells:
I'm joining the cpp files, so I'm finding conflicts in files that don't explicitly include headers. units.cpp is using `SIZE` which is already used by windows. If units needed to include the windows headers, it would get an ambiguous definition error.

xzyfer:
I'm sold

drewwells:
:beers: :beers:

saper:
I just came to conclusion we should rename all constants and enums either a prefix...

drewwells:
This is okay for now.

technoAri:
plz visit https://github.com/technoAri to see my contributions.

PR #1462, opened 2015-08-22
Fix memory manager for MSVC

### LINKED ISSUE #1451: Should throw error we performing binary operations on strings
```scss
@if "foo" > 0 {}
```
Ruby Sass
```
Error: Undefined operation: ""foo" gt 0".
        on line 56 of test.scss
```
Libsass 3.2.5
```
Error: may only compare numbers
        on line 56 of test.scss
>> @if "foo" > 0 {}
```
Libsass master
```
Segfault
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25863521-should-throw-error-we-performing-binary-operations-on-strings?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1460: crash with invalid @warn input in Visual Studio
Similar to #1443 that I reported regarding a crash calling rbga with invalid parameters when compiled with VS, I find that

@warn foo!;

causes a crash in the same place (Memory_Manager<T>::destroy(T* np)).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26024579-crash-with-invalid-warn-input-in-visual-studio?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Fixes #1451
Fixes #1460 and properly addresses the problem described also in #1446

MSVC calls allocator first before evaluating arguments for the actual object constructor. Since the evaluation phase can quite easily throw an error (like in `parse_list`), the memory manager is left in an undefined state. The object is already tracked but not initialized and we get an error if we try to `delete` the object (we should use free instead).

```C++
new (ctx.mem) Warning(pstate, parse_list());
// -> calls overloaded `new` (allocate and manage memory)
// -> calls `parse_list` afterwards (throwing an error)
// -> memory already managed, so we try to delete it ...
```

In gcc `parse_list` is called before the overloaded new operator!
I've added a macro to be used instead of overloaded new operator!

Memory Manager now allocates an additional byte to keep status if the object was properly initialized. The macro makes sure that we first allocate the memory and once `new` was successfull, the status is updated, so we call the destructor once the object goes out of scope. Uninitalized objects, which is the case if `new` throws an error, are deallocated via `free` (without calling the destructor).

Note: we could also "abuse" the actual memory of the object, since `AST_Node` is a virtual class we probably always have the `vfptr` at the first position. Other than that nothing guarantees that the first byte of an object cannot be `0` after initialization, therefore keeping the "status" and the actual object separated in memory seems safer and I guess the overhead is minimal.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -287,11 +287,11 @@ namespace Sass {
     }
     if (!found)
     {
-      Compound_Selector* cpy = new (ctx.mem) Compound_Selector(*rhs);
+      Compound_Selector* cpy = SASS_MEMORY_NEW(ctx.mem, Compound_Selector, *rhs);
       (*cpy) << this;
       return cpy;
     }
-    Compound_Selector* cpy = new (ctx.mem) Compound_Selector(rhs->pstate());
+    Compound_Selector* cpy = SASS_MEMORY_NEW(ctx.mem, Compound_Selector, rhs->pstate());
     for (size_t j = 0; j < i; ++j)
     { (*cpy) << (*rhs)[j]; }
     (*cpy) << this;
@@ -311,7 +311,7 @@ namespace Sass {
       if (!rhs->is_universal_ns())
       {
         // creaty the copy inside (avoid unnecessary copies)
-        Type_Selector* ts = new (ctx.mem) Type_Selector(*this);
+        Type_Selector* ts = SASS_MEMORY_NEW(ctx.mem, Type_Selector, *this);
         // overwrite the name if star is given as name
         if (ts->name() == "*") { ts->name(rhs->name()); }
         // now overwrite the namespace name and flag
@@ -325,7 +325,7 @@ namespace Sass {
     if (name() == "*" && rhs->name() != "*")
     {
       // creaty the copy inside (avoid unnecessary copies)
-      Type_Selector* ts = new (ctx.mem) Type_Selector(*this);
+      Type_Selector* ts = SASS_MEMORY_NEW(ctx.mem, Type_Selector, *this);
       // simply set the new name
       ts->name(rhs->name());
       // return copy
@@ -341,7 +341,7 @@ namespace Sass {
 
     // if the rhs is empty, just return a copy of this
     if (rhs->length() == 0) {
-      Compound_Selector* cpy = new (ctx.mem) Compound_Selector(rhs->pstate());
+      Compound_Selector* cpy = SASS_MEMORY_NEW(ctx.mem, Compound_Selector, rhs->pstate());
       (*cpy) << this;
       return cpy;
     }
@@ -353,14 +353,14 @@ namespace Sass {
       if (typeid(*rhs_0) == typeid(Type_Selector))
       {
         // if rhs is universal, just return this tagname + rhs's qualifiers
-        Compound_Selector* cpy = new (ctx.mem) Compound_Selector(*rhs);
+        Compound_Selector* cpy = SASS_MEMORY_NEW(ctx.mem, Compound_Selector, *rhs);
         Type_Selector* ts = static_cast<Type_Selector*>(rhs_0);
         (*cpy)[0] = this->unify_with(ts, ctx);
         return cpy;
       }
       else if (dynamic_cast<Selector_Qualifier*>(rhs_0)) {
         // qualifier is `.class`, so we can prefix with `ns|*.class`
-        Compound_Selector* cpy = new (ctx.mem) Compound_Selector(rhs->pstate());
+        Compound_Selector* cpy = SASS_MEMORY_NEW(ctx.mem, Compound_Selector, rhs->pstate());
         if (has_ns() && !rhs_0->has_ns()) {
           if (ns() != "*") (*cpy) << this;
         }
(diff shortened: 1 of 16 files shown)


xzyfer:
Not thrilled about the new AST node construction syntax. Is it necessary?

mgreter:
As I explained, there would only be a way if we would "abuse" the same memory as the object occupies, but this is not guaranteed to work by any standards (it would rely on the fact that the first bytes are not zero after the object has been initialized, which should be true for virtual classes). The macro is IMO the safest way, since we need to be invoked as soon as the object is initialized, so we know when it is safe to call delete on it. I cannot think of any other way without performance penalty!

saper:
If we modify the way the allocator is invoked, why not ditch the placement new altogether and just add objects to the MemoryManager via the macro?

Another improvement idea (the next step?) would be to use some kind of smart pointers there...

mgreter:
We need new to initialize the object (ie call the constructor) ... not sure how you want to ditch that? And yes, we may use the additional memory space to add in object memory reference counting at a later point.

saper:
Of course `new` has to stay. Let's ditch `(mgr.allocate(...))` only.

If you plan to do reference counting, why not use smart pointers instead (`shared_ptr` is refcounted afaik). Besides, it would be easier to introduce step-by-step child object ownership (something along the lines of #1386).

mgreter:
We may want to use an actual custom allocator at a later stage, so keeping it that way we only need to change the actual implementation later. This PR fixes the problem at hand, so I'm happy with the current state. We can discuss shared_ptr and what not when we actually think about using them.

saper:
> We may want to use an actual custom allocator at a later stage, so
> keeping it that way we only need to change the actual implementation
> later. This PR fixes the problem at hand, so I'm happy with the
> current state. We can discuss shared_ptr and what not when we actually
> think about using them.

Sorry,
"We may want to use (...) at a later stage" is not fixing the problem at hand.

mgreter:
It fixes the issues seen with MSVC without changing anything else. This is what I see *shrug*. It handles the memory correctly, without leaking if an error occurs after initial memory allocation, before object is really constructed. It releases it via delete correctly afterwards.

saper:
> It fixes the issues seen with MSVC without changing anything
> else. This is what I see *shrug*. 
> It handles the memory correctly,
> without leaking if an error occurs after initial memory allocation,
> before object is really constructed. It releases it via delete
> correctly afterwards.

I have a question:

I'd like to implement object ownership for those AST_Node's which
are tighly bound. For example, Definition instances should
own its Parameter instances.

I believe this should be done by implementing destructors
which delete subordinate objects.

Should I take those subordinate objects out of control of the
memory manager or is there a chance that memory manager
can have weak pointers and figure out itself, that the subordinate
object got purged?

mgreter:
There is no real micro memory management implemented yet. I guess there is now a basic foundation to get the next steps done (for example you can now pass a `Memory_Manager` reference in most places, where memory gets allocated, a few still take a `Context`, which is sometimes not available). So far we only use the `Memory_Manager` on the global `Context` object. So every `AST_Node` has a guaranteed life span as long as the `Context` is available.

I can see that for normal `AST_Nodes` a tree may be well suited, but a `Value` may be assigned to the global variable scope, so it may should life longer than it's actual parent. Long story short, there is still a lot of heavy lifting to be done to properly support something like this. And btw. the real paint point are variable assignments, as a loop can easily chew up all the memory. I had a working experimental branch that released memory from assignments correctly, but that has become unmergable with latest master.

saper:
On Sun, 23 Aug 2015, Marcel Greter wrote:

> It fixes the issues seen with MSVC without changing anything
> else. This is what I see *shrug*. It handles the memory correctly,
> without leaking if an error occurs after initial memory allocation,
> before object is really constructed. It releases it via delete
> correctly afterwards.

It changes the invocation. It adds more spaghetti code (sorry).
Why not just use something like this sketch:

    #define SASS_MEMORY_NEW(mgr, Class, ...)                         \
       mgr.add(new Class(__VA_ARGS__))

where `mgr.add()` does only:

    nodes.push_back(parameter);

Do we really need more? I think playing with our own overallocated
memory "just in case" is going in the wrong direction. It's already
way too complicated.

mgreter:
We will use custom allocators in the future! Having it there already means that the basics are already tested. Once we add micro memory management we can do various experiments to figure out what kind of allocator fits bests. A very easy optimization will be to keep a certain amount of memory container around for re-use. So I don't see the point to remove it now just to add it back later. I don't think it adds much overhead (if any at all).

saper:
> We will use custom allocators in the future! Having it there already
> means that the basics are already tested. Once we add micro memory
> management we can do various experiments to figure out what kind of
> allocator fits bests. A very easy optimization will be to keep a
> certain amount of memory container around for re-use. So I don't see
> the point to remove it now just to add it back later. I don't think it
> adds much overhead (if any at all).

You will certainly disagree, but I think that playing with
an optimized memory allocator in a situation where even basic
memory management is non-existing is a premature optimization.

Plus I don't see the reason of keeping 80+ lines doing nothing.

Current code is obsure enough!

EOT from my side.

xzyfer:
I'm happy to move forward with. We all agree there is room for improvement but it fixes the critical issue at hand.

@mgreter appears to have plans for memory management moving forward, as well as the relevant context.

>Let's ditch (mgr.allocate(...)) only.
>Plus I don't see the reason of keeping 80+ lines doing nothing.

@saper makes good points, but I agree with @mgreter with leaving these for now. However I do think we should provide a user error in this code path. It'll help us track any missed `new`s and prevent contributors incorrectly using this method.

mgreter:
rebased to master

xzyfer:
This will conflict with #1475

mgreter:
BTW. it's very simple to update to the new macro:

```c++
new (ctx.mem) Class(x, y, ...);
SASS_MEMORY_NEW(ctx.mem, Class, x, y, ...);
```

I will keep rebasing ...

xzyfer:
Shipping this since #1475 is still not quite correct.

PR #1463, opened 2015-08-22
Fix source map include paths

mgreter:
Fixes a few issues discovered while implementing [sourcemap inspector] [1].

- Mixed up arguments for `add_source`
- Recognize absolute URLs for file handling

[1]: http://libsass.ocbnet.ch/srcmap/


mgreter:
This change can be seen in action at http://libsass.ocbnet.ch/srcmap/
screen-srcmap-inspect

PR #1469, opened 2015-08-23
Add error for non existing function with keyword arguments

### LINKED ISSUE #1418: Should be an error for named arguments in missing functions
Named arguments are not allowed in missing functions and an error should be raised.  This includes functions invoked via call() when the function is not found.

Test Case 1:
```
p {
	color: missing($a: b);
}
```

Ruby SASS Output:
```
/* Error: Function missing doesn't support keyword arguments ...
```

Libsass Output:
```
p {
  color: missing($a: b);
}
```

Test Case 2:
```
p {
	color: call(missing, $a: b);
}
```

Ruby SASS Output:
```
/* Error: Function missing doesn't support keyword arguments ...
```

Libsass Output:
```
p {
  color: missing(b);
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25107006-should-be-an-error-for-named-arguments-in-missing-functions?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Fixes #1418

### DIFF
--- a/src/debugger.hpp
+++ b/src/debugger.hpp
@@ -424,6 +424,9 @@ inline void debug_ast(AST_Node* node, std::string ind = "", Env* env = 0)
     std::cerr << ind << "Arguments " << expression;
     if (expression->is_delayed()) std::cerr << " [delayed]";
     std::cerr << " (" << pstate_source_position(node) << ")";
+    if (expression->has_named_arguments()) std::cerr << " [has_named_arguments]";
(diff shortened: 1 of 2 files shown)


xzyfer:
:+1:

PR #1470, opened 2015-08-23
Implement error when multiple imports are found

mgreter:
When multiple errors are found, ruby sass gives this message:

```
Error: It's not clear which file to import for '@import "imp"'.
       Candidates:
         imp.sass
         imp.scss
       Please delete or rename all but one of these files.
        on line 1 of foo.sass
  Use --trace for backtrace.
```

With this patch, libsass gives this error message:

```
Error: It's not clear which file to import for '@import "imp"'.
       Candidates:
         imp.scss
         imp.sass
       Please delete or rename all but one of these files.
        on line 1 of foo.sass
>> @import "imp";
   ^
```

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -248,17 +248,29 @@ namespace Sass {
 
   // Add a new import file to the context
   // This has some previous directory context
-  std::string Context::add_file(const std::string& base, const std::string& file)
+  std::string Context::add_file(const std::string& base, const std::string& file, ParserState pstate)
   {
     using namespace File;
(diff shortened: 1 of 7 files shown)


xzyfer:
:+1:

martinf55:
This breaks on files which import an .scss for which there is also a .css file in the same folder.

custom.scss:
@import "base"

base.scss:
...

base.css:
...

rubysass and libsass prior to this commit correctly select the base.scss file for import.

Issue #1397, opened 2015-07-27
Difference in interpolated function calls


xzyfer:
```scss
div {
  color: #{foo}(a, 1+2, c);
}
```
Ruby Sass
```scss
div {
  color: fooa, 3, c; }
```
LibSass
```scss
div {
  color: foo(a, 3, c); }
```

I'm not sure what the rationale behind this behaviour is. I've asked for an explanation upstream. However for the time being this is an inconsistency with an existing spec.

Spec https://github.com/sass/sass-spec/tree/master/spec/libsass-todo-tests/libsass/interpolated-function-call

andrew-skybound:
I've run some additional tests and it looks like there are quite a few failing cases which are all likely related to the originally reported issue.

Here are six test cases, showing interpolation around a function call, and around a parenthesized expression.  The first six declarations are the test cases themselves, the next six show the `type-of()`, and the last six show the `length()`.

Input:
```
p {
	color: #{A}round(0.5);
	color: round(0.5)#{B};
	color: #{A}round(0.5)#{B};
	color: #{A}(B C);
	color: (B C)#{D};
	color: #{A}(B C)#{D};
	/* typeof */
	color: type-of(#{A}round(0.5));
	color: type-of(round(0.5)#{B});
	color: type-of(#{A}round(0.5)#{B});
	color: type-of(#{A}(B C));
	color: type-of((B C)#{D});
	color: type-of(#{A}(B C)#{D});
	/* lengths */
	color: length(#{A}round(0.5));
	color: length(round(0.5)#{B});
	color: length(#{A}round(0.5)#{B});
	color: length(#{A}(B C));
	color: length((B C)#{D});
	color: length(#{A}(B C)#{D});
}
```

Ruby Sass
```
p {
  color: A1;
  color: 1B;
  color: A1B;
  color: AB C;
  color: B CD;
  color: AB CD;
  /* typeof */
  color: string;
  color: string;
  color: string;
  color: string;
  color: string;
  color: string;
  /* lengths */
  color: 1;
  color: 1;
  color: 1;
  color: 1;
  color: 1;
  color: 1;
}
```

Libsass
```
p {
  color: Around(0.5);
  color: round0.5B;
  color: Around0.5B;
  color: A(B C);
  color: B CD;
  color: AB CD;
  /* typeof */
  color: string;
  color: list;
  color: list;
  color: string;
  color: list;
  color: list;
  /* lengths */
  color: 1;
  color: 2;
  color: 2;
[...]

PR #1474, opened 2015-08-23
Fix `functional_schema` parsing ending with interpolation

mgreter:
Addresses https://github.com/sass/libsass/issues/1397

The issue with additional space added still exists.
This PR only addresses the function schema parsing.


xzyfer:
:+1:

PR #1481, opened 2015-08-24
source mingw32 from bintray

drewwells:
Sourceforge often fails to download or gets severly capped when appveyor
attempts to download.


xzyfer:
I was about to do this. Thanks @drewwells

drewwells:
Figured I'd put my code where my mouth was :)

drewwells:
Also added the 4.9.2 mingw to bintray. The versioning of bintray is confusing, I'll add proper versioning to the 5.1.0 on there if we want to upgrade at some point.

xzyfer:
Probably going to steal this for node-sass as well if you have no objections.

drewwells:
Works for me, I believe it's unlimited for open source software.

Issue #1482, opened 2015-08-24
Parent reference in selector with @extends produce wrong output


xzyfer:
Originally reported on https://github.com/sass/sassc/issues/137
```scss
.mango {
  color: red;
}

.something {
  &__else {
    @extend .mango;
    color: blue;
  }
}
```
Ruby Sass
```scss
.mango, .something__else {
  color: red; }

.something__else {
  color: blue; }
```
LibSass 3.2.5
```scss
.mango, .something__else {
  color: red; }

.something__else {
  color: blue; }
```
LibSass master
```scss
mango, __else.something {
  color: red; }

.something__else {
  color: blue; }
```
Spec added https://github.com/sass/sass-spec/pull/478

PR #1483, opened 2015-08-24
Implement `Compound_Selector` merging in `append` function

mgreter:
Fixes https://github.com/sass/libsass/issues/1482

Code is a bit dirty but should do the job ...


xzyfer:
:+1:

PR #1495, opened 2015-08-29
Code: Fixes include guard condition

am11:
With this change, we can compile `src/c99func.c` without any [extra condition](https://github.com/sass/libsass/blob/bedefc/win/libsass.targets#L59) in down-streams' build process.

### DIFF
--- a/src/c99func.c
+++ b/src/c99func.c
@@ -21,18 +21,11 @@
   THE SOFTWARE.
 */
 
-#ifdef _MSC_VER
-#define _CRT_SECURE_NO_WARNINGS
-#define _CRT_NONSTDC_NO_DEPRECATE
-#endif
+#if defined(_MSC_VER) && _MSC_VER < 1900
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
-
-#ifdef snprintf
-#undef snprintf
-#endif
 
 static int c99_vsnprintf(char* str, size_t size, const char* format, va_list ap)
 {
@@ -57,3 +50,5 @@ int snprintf(char* str, size_t size, const char* format, ...)
 
     return count;
 }
+
+#endif


saper:
Well, the whole point of the c99func.c is not to compile that file when not needed... you can leave ifdefs if you Iike but I'd leave msbuild conditional rules in place.

am11:
Yeah I remember the point of c99func.c. However, the point of this PR is to save every downstream from checking the condition to be able to build with both VS2013 and VS2015. See https://github.com/am11/node-sass/blob/request-ci-artifacts/src/libsass.gyp#L82-L84, there is no decent condition to check MSC_VER there and this approach saved me sometime and hopefully will save others' as well.

saper:
That is why I think we can keep MSBuild rule and leave conditionals inside. I will deal with libsass.gyp once our gyp fully supports vs2015.
I want to make sure we don't get any linker errors when not linking c99func.obj

am11:
I found this guy trying to figure out a way to detect toolset version in gyp: http://stackoverflow.com/q/26311665/863980 (received no answer). 
Then tested with VS2013 (on AppVeyor) and VS2015 (both locally and on AppVeyor) with this change, there is no linker error so far; compilation succeeds (only some [tests are failing](https://github.com/sass/libsass/issues/1500) atm).

saper:
Here it goes: https://github.com/saper/node-sass/commit/edf71270750fa6d6538359409edd935f8c4beb35 - can you check that with VS2015 if `c99func.c` gets excluded from the build?

am11:
Yup c99func gets excluded from VS2015, when `ps: $env:GYP_MSVS_VERSION=2015`. Nicely done! 

BTW, AppVeyor is also building with VS2015, but the file pangyp was emitting VS2013 proj files, with c99func. I have sent a PR on your fork.

Issue #1500, opened 2015-08-30
Source-map path resolution is broken on Windows


am11:
While requesting latest artifacts from AppVeyor [yesterday](https://github.com/sass/node-sass/issues/970#issuecomment-136054161), I realized that some source-map path related tests were failing on Windows which are passing on Linux (and even Windows when switched back to libsass v3.2.0 stable, i.e. node-sass current master).

TravisCI build: https://travis-ci.org/am11/node-sass/jobs/77901212
AppVeyor build: https://ci.appveyor.com/project/am11/node-sass/build/702

Step to repro:

In PowerShell:

```powershell
git clone -b request-ci-artifacts https://github.com/am11/node-sass
cd node-sass

git submodule update --init --recursive
# currently pointing to tip of libsass master

npm install
node scripts/build -f

mkdir /temp
notepad /temp/blah.scss

# enter:
# a{b:c}
# save and close

node
# in node REPL
r = require('./').renderSync({file:'/temp/blah.scss', outFile: '/temp/blah.css', sourceMap: true})

r.css.toString()
r.map.toString()
```

Yields:

> 'a {\n  b: c; }\n\n/*# sourceMappingURL=C:/temp/blah.css.map */'

(sourceMappURL is absolute path, which is incorrect)

and

> '{\n\t"version": 3,\n\t"file": "C:/temp/blah.css",\n\t"sources": [\n\t\t"C:/temp/blah.scss"\n\t],\n\t"mappings": "AAAA,C
AAC,CAAA;EAAC,CAAC,EAAC,CAAE,GAAE",\n\t"names": []\n}'

(file path and those in sources array are absolute, which is again incorrect)

Repeated ditto steps on Ubuntu (with `sudo -i` and switch `notepad` with `nano`); received the desired output:

> 'a {\n  b: c; }\n\n/*# sourceMappingURL=blah.css.map */'

and

> '{\n\t"version": 3,\n\t"file": "blah.css",\n\t"sources": [\n\t\t"blah.scss"\n\t],\n\t"mappings": "AAAA,CAAC,CAAA;EAAC,CA
AC,EAAC,CAAE,GAAE",\n\t"names": []\n}'

PR #1502, opened 2015-08-30
Fix `resolve_relative_path` for absolute windows paths

mgreter:
They wrongly were seen as urls (protocol vs drive letter).

Fixes https://github.com/sass/libsass/issues/1500


am11:
I can confirm that it fixes the node-sass tests issue. Thank you @mgreter! :+1:

Issue #439, opened 2014-07-11
Placeholder selector issues


errorx666:
I have the following mixins designed to emulate `:nth-child(even)` and `:nth-child(odd)` in IE8.  I am having two problems that do not occur in the default Sass implementation:

```scss
@mixin odd( $selector: '*', $n: 50 ) {
	$local-sel: '&';
	$local-uniq: unique_id();

	#{$selector}:nth-child(odd) { @content }

	%#{$local-uniq} { @content; }

	#{$selector}:first-child {
		@for $i from 1 through $n {
			#{$local-sel} { @extend %#{$local-uniq}; }

			$local-sel: $local-sel + ' + ' + $selector + ' + ' + $selector;

		}

	}

}

@mixin even( $selector: '*', $n: 50 ) {
	$local-sel: '&';
	$local-uniq: unique_id();

	#{$selector}:nth-child(even) { @content }

	%#{$local-uniq} { @content; }

	#{$selector}:first-child + #{$selector} {
		@for $i from 1 through $n {
			#{$local-sel} { @extend %#{$local-uniq}; }

			$local-sel: $local-sel + ' + ' + $selector + ' + ' + $selector;

		}

	}

}
```

An example use-case is:
```scss
ul > {
  @include odd( li, 5 ) { background: #ccc;  }
}
```
Which in Sass 3.3.9 produces:
```css
ul > li:nth-child(odd) {
  background: #ccc;
}
ul > li:first-child, ul > li:first-child + li + li, ul > li:first-child + li + li + li + li, ul > li:first-child + li + li + li + li + li + li, ul > li:first-child + li + li + li + li + li + li + li + li {
  background: #ccc;
}
```

In LibSass I get: `source string:7: error: invalid selector after %unique-id`

---
The unique ID is needed to stop successive invocations from using the same placeholder selector.

If I replace the unique ID, I get a successful compile, but the selectors do not match the Sass implementation:

```scss
@mixin odd( $selector: '*', $n: 50 ) {
	$local-sel: '&';

	#{$selector}:nth-child(odd) { @content }

	%foo { @content; }

	#{$selector}:first-child {
		@for $i from 1 through $n {
			#{$local-sel} { @extend %foo; }

			$local-sel: $local-sel + ' + ' + $selector + ' + ' + $selector;

		}

	}

}

ul > {
  @include odd( li, 5 ) { background: #ccc;  }
}
```

Expected:
```css
ul > li:nth-child(odd) {
  background: #ccc;
}
[...]

PR #1503, opened 2015-08-31
Fix copy paste error in `mergeFinalOps`

mgreter:
And the sequel: Mission impossible 2 accomplished -> Fixes https://github.com/sass/libsass/issues/439
Original ruby sass code: https://github.com/sass/sass/blob/stable/lib/sass/selector/sequence.rb#L400


xzyfer:
Nice find!

Issue #1505, opened 2015-08-31
Mixins applied via a header no longer work


drewwells:
Bisected to this change: https://github.com/sass/libsass/commit/463465c1d474e270bb1792053bcd58427a0ce6f7

See a test (in Go) here: https://github.com/wellington/go-libsass/blob/master/header_test.go#L8-L35

Adding a mixin via a header, then calling that mixin results in an error.

Header
```
@mixin mix() {
  width: 50px;
}
```
Main sass file
```
div {
  @include mix();
}
```

**Expected**:
```
div {
  width: 50px; }
```

**Got**:
```
Error > stdin:2
		no mixin named mix

		Backtrace:
			stdin:2
		div {
		  @include mix();
		}
```

mgreter:
Please checkout #1509 ... I think this should fix this issue!

PR #1509, opened 2015-08-31
Improve importer C-API and path handling

mgreter:
This is a breaking C-API change, altough only minor and in an experimental feature!

Renamed two API functions to be more close to their actual use:
`sass_import_get_path` ->  `sass_import_get_imp_path`
`sass_import_get_base` -> `sass_import_get_abs_path`

This PR also reworks and should make path handling better understandable. Still far from perfect under the hood, but IMO the API is slowly getting there. It _should_ now support urls correctly.

This also fixes another bug introduced by a fix commited a few days ago. Headers were all stored under the same virtual file-name. Therefore the "cache" always returned the same header. We now add an index to the virtual file-name if multiple headers are included in the same file (Fixes https://github.com/sass/libsass/issues/1505).

### DIFF
--- a/include/sass_functions.h
+++ b/include/sass_functions.h
@@ -57,7 +57,7 @@ ADDAPI void ADDCALL sass_delete_importer (Sass_Importer_Entry cb);
 ADDAPI Sass_Import_List ADDCALL sass_make_import_list (size_t length);
 // Creator for a single import entry returned by the custom importer inside the list
 ADDAPI Sass_Import_Entry ADDCALL sass_make_import_entry (const char* path, char* source, char* srcmap);
-ADDAPI Sass_Import_Entry ADDCALL sass_make_import (const char* path, const char* base, char* source, char* srcmap);
+ADDAPI Sass_Import_Entry ADDCALL sass_make_import (const char* imp_path, const char* abs_base, char* source, char* srcmap);
 // set error message to abort import and to print out a message (path from existing object is used in output)
 ADDAPI Sass_Import_Entry ADDCALL sass_import_set_error(Sass_Import_Entry import, const char* message, size_t line, size_t col);
 
@@ -67,8 +67,8 @@ ADDAPI void ADDCALL sass_import_set_list_entry (Sass_Import_List list, size_t id
 ADDAPI Sass_Import_Entry ADDCALL sass_import_get_list_entry (Sass_Import_List list, size_t idx);
 
 // Getters for import entry
-ADDAPI const char* ADDCALL sass_import_get_path (Sass_Import_Entry);
-ADDAPI const char* ADDCALL sass_import_get_base (Sass_Import_Entry);
+ADDAPI const char* ADDCALL sass_import_get_imp_path (Sass_Import_Entry);
+ADDAPI const char* ADDCALL sass_import_get_abs_path (Sass_Import_Entry);
 ADDAPI const char* ADDCALL sass_import_get_source (Sass_Import_Entry);
 ADDAPI const char* ADDCALL sass_import_get_srcmap (Sass_Import_Entry);
 // Explicit functions to take ownership of these items
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1924,6 +1924,7 @@ namespace Sass {
     // some final cosmetics
     if (res == "-0.0") res.erase(0, 1);
     else if (res == "-0") res.erase(0, 1);
+    else if (res == "") res = "0";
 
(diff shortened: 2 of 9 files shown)


drewwells:
Does this break any existing functions or have just these two functions
moved to new names?

mgreter:
There were some changes under the hood to support urls better. Other than that everything should behave as before (perl-libsass unit tests also triggered and are now working again).

saper:
(related: https://github.com/sass/libsass/issues/1160)

xzyfer:
No objections from me.

drewwells:
@mgreter most of my header tests are passing now.

I found one last oddity. If you include a header then do a file compile, the only file that shows up in resolved imports is the header not the file being compiled. Previously, the resolved imports included only the file being compiled and not the header. If I remove the header and do a file compile, the resolved imports includes an absolute path to the file being compiled.

Input
```
import_entry_....("/path/to/header");
sass_make_file_context("/path/to");
```

Previous:
```
resolved_imports -> "/path/to";
```

Now:
```
resolved_imports -> "header"
```

Imo, correct:
```
resolved_imports -> ["/abs/path/to/header", "/abs/path/to/file"]
```

mgreter:
What is `resolved_imports`? That's not from libsass directly!? Are you sure you use `abs_path` and not `imp_path`? Can you reproduce with a plain C sample!?

drewwells:
sass_context_get_included_files. Yeah I switched to abs_path/imp_path, it
didn't compile until I did that :).

drewwells:
This shows what I'm talking about

```
#include <stdio.h>
#include <string.h>
#include "sass_context.h"

Sass_Import_List sass_importer(const char* path, Sass_Importer_Entry cb, struct Sass_Compiler* comp)
{
  // get the cookie from importer descriptor
  void* cookie = sass_importer_get_cookie(cb);
  Sass_Import_List list = sass_make_import_list(2);
  const char* local = "local { color: green; }";
  const char* remote = "remote { color: red; }";
  list[0] = sass_make_import_entry("/tmp/styles.scss", strdup(local), 0);
  list[1] = sass_make_import_entry("http://www.example.com", strdup(remote), 0);
  return list;
}

int main( int argc, const char* argv[] )
{

  // get the input file from first argument or use default
  const char* input = argc > 1 ? argv[1] : "../test/scss/main.scss";

  // create the file context and get all related structs
  struct Sass_File_Context* file_ctx = sass_make_file_context(input);
  struct Sass_Context* ctx = sass_file_context_get_context(file_ctx);
  struct Sass_Options* ctx_opt = sass_context_get_options(ctx);

  // allocate custom importer
  Sass_Importer_Entry c_imp =
    sass_make_importer(sass_importer, 0, 0);
  // create list for all custom importers
  Sass_Importer_List imp_list = sass_make_importer_list(1);
  // put the only importer on to the list
  sass_importer_set_list_entry(imp_list, 0, c_imp);
  // register list on to the context options
  sass_option_set_c_headers(ctx_opt, imp_list);
  // context is set up, call the compile step now
  int status = sass_compile_file_context(file_ctx);

  // print the result or the error to the stdout
  if (status == 0) puts(sass_context_get_output_string(ctx));
  else puts(sass_context_get_error_message(ctx));

  char** imps = sass_context_get_included_files(ctx);
  size_t size = sass_context_get_included_files_size(ctx);
  for (int i=0; i<size; i++) {
    printf("%s\n", imps[i]);
  }

  // release allocated memory
  sass_delete_file_context(file_ctx);

  // exit status
  return status;

}
```

Output:
`//www.example.com`

Should be:
```
//www.example.com
/abs/test/scss/main.scss
```

mgreter:
You're case should be covered now. Please open new bugs if I broke anything else!

drewwells:
So is it considered a bug that the headers are still left out of imports on the file compiler?

mgreter:
This is by design and I would argue that it doesn't make much sense to include them there (it is included in the source-maps though, as you can see when you enable "Custom Header Test" in srcmap inspector). I guess `@import` statements in the header should be included but not the headers themselve. Most implementors use IMO that array to get info which files they need to watch and headers should be generated by code, so they cannot be watched. If they actually can, you should probably just use an import (either in the headers or in the main source file). Does that make sense?

drewwells:
That is fine, I don't have a use case for watching headers just noticed that the unit tests are failing. It might be worth documenting somewhere that headers are not included in the resolved imports list. I did notice these before so that it would be a breaking change from 3.2.5 behavior.

Issue #1508, opened 2015-08-31
extern "C" copy_strings() should not throw a C++ exception


saper:
I was experimenting with `/EHs` vs `/EHsc` [options to the Visual Studio](https://msdn.microsoft.com/en-us/library/1deeycx5.aspx), since the next version of `gyp` we are using for node-sass will be able to automagically add `/EHsc`. But with that option Visual Studio complains, that `copy_strings` is `extern "C"` but throws unexpected exception. 

My suggestion: take `static` functions out of the `extern "C"` block or just roll the function into the caller (it's one invocation only).

Thoughts?

mgreter:
I guess we could just move it to utils?

mgreter:
Is this still a thing?

xzyfer:
I believe this was closed by #1521. Can you please confirm @saper.

saper:
Should be. Merging now changes into `node-sass` and then I can see it very clearly with our gyp setup.

PR #1511, opened 2015-09-01
Move `copy_strings` and `free_string_array` to `util`

mgreter:
Addresses https://github.com/sass/libsass/issues/1508

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -187,35 +187,6 @@ extern "C" {
     return str == NULL ? "" : str;
   }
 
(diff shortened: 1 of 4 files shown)


saper:
Good idea to move those functions - actually they both are supposed to do the same, just `copy_strings` diverged quickly.

I have proposed my take in https://github.com/sass/libsass/commit/e9bffb2bfc5d46a0049cb25d8dc0f159e451d546 (part of https://github.com/sass/libsass/pull/1521 now).

Issue #1513, opened 2015-09-01
Regression in 'inline source comments' and 'source map mappings'


am11:
With the latest libsass: 702ac5aa7079e8b8ab905a307f99840b16cda0d6, two of the node-sass tests are failing
https://travis-ci.org/am11/node-sass/jobs/78170861#L1888-L1937.

The first one is about source comment and is referring to [this input](https://github.com/sass/node-sass/blob/v3.0.0/test/fixtures/source-comments/index.scss) and the corresponding [expected output](https://github.com/sass/node-sass/blob/v3.0.0/test/fixtures/source-comments/expected.css). The line number is off by 2 in the actual output.

The second failure is related to source-map mappings, here is the diff of decoding: https://www.diffnow.com/?report=bohl1 (left: old, right: new). Though, I have not verified yet whether old behavior was correct in this test case or the new one.

mgreter:
https://github.com/sass/libsass/pull/1510

am11:
Apparently it is happening with those changes inclusive as my libsass submodule is pointing to https://github.com/sass/libsass/commit/702ac5aa7079e8b8ab905a307f99840b16cda0d6:
https://github.com/am11/node-sass/tree/request-ci-artifacts/src.

am11:
@mgreter, also tested with stable release v3.2.5 of libsass and sassc:

* Extracted https://github.com/sass/libsass/archive/3.2.5.zip
* Then extracted: https://github.com/sass/sassc/archive/3.2.5.zip inside libsass.
* Built libsass/sassc/win/sassc.sln.
* Saved a file \temp\test.scss with the following code:

  ```scss
    .a {
        color: #000;
    }

    .b {
        color: #001;
    }

    .c {
        color: #002;
    }
  ```
  (tested with both line endings \r and \r\n, there is no difference in the output)

* Ran `libsas/sassc/bin/sassc /temp/test.scss`, on stdout I got:

    ```css
    /* line 1, /temp/test.scss */
    .a {
      color: #000; }

    /* line 5, /temp/test.scss */
    .b {
      color: #001; }

    /* line 9, /temp/test.scss */
    .c {
      color: #002; }
    ```

With current master, it is producing:

```css
/* line 1, C:/temp/test.scss */
.a {
  color: #000; }

/* line 3, C:/temp/test.scss */
.b {
  color: #001; }

/* line 7, C:/temp/test.scss */
.c {
  color: #002; }
```

[...]

PR #1515, opened 2015-09-01
Add option to lexer to force parserState update

mgreter:
Addresses https://github.com/sass/libsass/issues/1513


am11:
I can confirm that this fixes the line-number issue described in #1513. :+1: 

Thank you @mgreter! :shipit:

Issue #885, opened 2015-02-12
API: source-map related options enabling criteria


am11:
With growing source-map related options, there has been some confusion among node-sass users. 

node-sass takes the user options and send to LibSass as is (at the risk of default values `undefined` type casted to `null`).

I believe there is a need of a well-defined policy. to make the "require options to set" criteria for each case. I propose to make this kind of policy as relaxed as possible, i.e. let users set minimum options to enable the intended feature. To devise such a policy, some changes might be required.

Copying from https://github.com/sass/node-sass/issues/622#issuecomment-70050834, here is an extract of what we have found:

- `source_map_contents == true` should emit source-map without needing the presence of `source_map`.
- `source_map_contents == true` should emit source-map without needing the presence of `output_path`.
- `source_map_embed == true` should emit source-map without needing the presence of `source_map`.
- `source_map_embed == true` should emit blank string on `sass_context_get_source_map_string(ctx)`. (**NEW** based on https://github.com/joliss/broccoli-sass/pull/43#issuecomment-73139839).
- `source_map_embed == true` should emit source-map without needing the presence of `input_path`, when `data` option is set.
- `source_map_embed == true && source_map_contents == true` should emit source-map without needing the presence of `output_path`. (hint: emit `file: ''` in JSON)
- `source_map_embed == true && source_map_contents == true` should emit source-map without needing the presence of `input_path` when `data` option is set.

*(note: there might be more considerations here that I am missing ..)*

Once agreed upon and LibSass API defines the policy to set which options to enable what, we will use the same in node-sass README (or hopefully wiki) for node-sass API implementers.

aexmachina:
I can confirm that specifying `sourceMapEmbed: true` does _not_ produce an embedded source map unless `sourceMap: true` is also specified.

mgreter:
IMO this would mostly just mean to implicitly enable `source_map` when either `source_map_contents` or `source_map_embed` are enabled. IMO this can easily been done by implementors, but I don't have personaly anything against such a change. What's your opinion @xzyfer?

xzyfer:
I agree with @am11, I believe I may have originally instigated this report. IMHO it's poor UX to have to enable `source_map` if you've explicitly declared which type of sourcemap you want.

Do we even need `source_map`? Is there a use case for declaring `source_map` that we cannot infer from other options?

am11:
[...]

PR #1516, opened 2015-09-02
Make `source_map_embed` independent of `source_map_file`

mgreter:
This should IMO address https://github.com/sass/libsass/issues/885

The `source_map_embed` option no longer needs `source_map_file` to be present. We currently always return a source-map on the C-API level (AFAIR), but from the semantics we do not guarantee that, when neither of two main options are set. `source_map_embed` has higher priority and `source_map_contents` doesn't do anything on its own.

So we have these options in regard of source-maps:

`source_map_file`: libsass only adds the sourceMappingURL to the output
`source_map_embed`: generate sourceMappingURL with embedded source-map
`source_map_contents`: adds the `sourceContents` array to the source-map
`omit_source_map_url`: suppress any sourceMappingURL output

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -289,8 +289,13 @@ namespace Sass {
     emitter.finalize();
     OutputBuffer emitted = emitter.get_buffer();
     std::string output = emitted.buffer;
-    if (source_map_file != "" && !omit_source_map_url) {
-      output += linefeed + format_source_mapping_url(source_map_file);
+    if (!omit_source_map_url) {
+      if (source_map_embed) {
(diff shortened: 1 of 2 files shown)


am11:
I have tested with your branch. It seems like even prior to this delta, some parts of #885 were addressed with previous improvements in source-map and file related PRs.

With this PR, AFAICT, all the concerns are addressed and the feature is much resilient than it used to be.

Thanks @mgreter for doing this. Your [sourcemap library](https://github.com/mgreter/sourcemap.cpp) is precious! :sunglasses:

Issue #1514, opened 2015-09-01
Floating point + the random() function


mknadler:
Bug exists in Ruby Sass as well; see https://github.com/sass/sass/pull/1819.

Given two integers, `x` and `y`, normal behavior is:
`x / y * y => x`

With certain values for x, however:
`x / y * y => x.0`

Under normal circumstances, this is not much of a problem; e.g., for font size, it's nbd. However, this behavior causes some calls to the `random` function to fail unexpectedly, as that function expects an integer for the `$limit` parameter.

Example here: http://sassmeister.com/gist/247d0a8155bed38bbca2.

PR #1517, opened 2015-09-02
Improve integer test for `random` function

mgreter:
This should fix https://github.com/sass/libsass/issues/1514 ... maybe it is worth a further look as I expected a smaller rounding error to compensate. Also the checks should probably be implemented more generic so others can re-use them.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1584,9 +1584,9 @@ namespace Sass {
   bool Number::operator== (const Expression& rhs) const
   {
     if (const Number* r = dynamic_cast<const Number*>(&rhs)) {
-      return (value() == r->value()) &&
-             (numerator_units_ == r->numerator_units_) &&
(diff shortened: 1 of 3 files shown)


am11:
There is probably one related edge case as @chriseppstein described in https://github.com/sass/sass/pull/1819:

```scss
a {
  b: (29 / 7 * 7) == 29;
}
```

should produce:

```css
a {
  b: true; }  /* at present, it generates b: false */
```

Reminds me of: https://isocpp.org/wiki/faq/newbie#floating-point-arith2  :)

mgreter:
Also added the epsilon test for number values, so the other sample should now also work. As I said, there are probably more edge cases like this, but this should at least fix the most important ones.

am11:
Thanks! a45d5c0 fixed the aforementioned edge case for equality. :+1:

PR #1426, opened 2015-08-04
Refactor feature query support to better match the reference impl

xzyfer:
The aim of this refactor is simple to bring the code and AST more inline with the Ruby Sass implementation to help us better address difference between how we handle feature queries. As an intended side affect this fixes some issues.

Fixes #1425 
Spec https://github.com/sass/sass-spec/pull/461

PR #1520, opened 2015-09-02
bool needs_parens(Supports_Condition* cond)

saper:
Fixes:

````
  clang++ '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' '-DLIBSASS_VERSION="3.2.5"' '-DDEBUG' '-D_DEBUG' -I/home/saper/.node-gyp/3.0.0/src -I/home/saper/.node-gyp/3.0.0/deps/uv/include -I/home/saper/.node-gyp/3.0.0/deps/v8/include -I../src/libsass/include  -fPIC -pthread -Wall -Wextra -Wno-unused-parameter -m64 -g -O0 -fno-rtti -std=gnu++0x -std=c++0x -fexceptions -frtti -MMD -MF ./Debug/.deps/Debug/obj.target/libsass/src/libsass/src/ast.o.d.raw -g -c -o Debug/obj.target/libsass/src/libsass/src/ast.o ../src/libsass/src/ast.cpp
In file included from ../src/libsass/src/ast.cpp:1:
../src/libsass/src/ast.hpp:1489:13: warning: 'const' type qualifier on return type has
      no effect [-Wignored-qualifiers]
    virtual const bool needs_parens(Supports_Condition* cond) { return false; }
            ^~~~~~
````

https://github.com/sass/libsass/pull/1426#issuecomment-137068114

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -13,13 +13,13 @@ namespace Sass {
 
   static Null sass_null(Sass::Null(ParserState("null")));
 
-  const bool Supports_Operator::needs_parens(Supports_Condition* cond) {
+  bool Supports_Operator::needs_parens(Supports_Condition* cond) const {
     return dynamic_cast<Supports_Negation*>(cond) ||
           (dynamic_cast<Supports_Operator*>(cond) &&
            dynamic_cast<Supports_Operator*>(cond)->operand() != operand());
   }
 
-  const bool Supports_Negation::needs_parens(Supports_Condition* cond) {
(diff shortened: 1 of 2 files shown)


xzyfer:
Interesting that I don't see this is clang. We must have different versions.

saper:
> Interesting that I don't see this is clang. We must have different versions.

node-gyp sets C++ compiler 
flags to "-Wall -Wextra -Wno-unused-parameter"
(at least with my debugging io.js 3.0 configuration), 
so if you override CXXFLAGS when building libsass you
should get them as well.

PR #1521, opened 2015-09-02
VS Exception handling only in C++ code

saper:
Catch exceptions only in C++ code,
extern "C" functions should (ideally) be exception-free.

While here, move functions shared between `sass_interface.cpp` and `sass_context.cpp` into `util.cpp`

Fixes: https://github.com/sass/libsass/issues/1508

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -187,22 +187,6 @@ extern "C" {
     return str == NULL ? "" : str;
   }
 
-  static void copy_strings(const std::vector<std::string>& strings, char*** array) {
-    int num = static_cast<int>(strings.size());
-    char** arr = (char**) malloc(sizeof(char*) * (num + 1));
-    if (arr == 0) throw(std::bad_alloc());
-
-    for(int i = 0; i < num; i++) {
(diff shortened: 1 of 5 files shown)


src/sass_context.cpp
```diff
@@ -189,12 +189,18 @@ extern "C" {
 
   static void copy_strings(const std::vector<std::string>& strings, char*** array) {
     int num = static_cast<int>(strings.size());
-    char** arr = (char**) malloc(sizeof(char*) * (num + 1));
-    if (arr == 0) throw(std::bad_alloc());
+    char** arr = (char**) calloc(num + 1, sizeof(char*));
+    if (arr == 0)
+      return;
```
QuLogic: Maybe set `*array` to `NULL` as an equivalent to the exception?

saper: Sure, my mistake. Even more, let's make it return the `(char **)` pointer for easier error checking.

mgreter:
Can you resolve the conflicts? Otherwise LGTM!

saper:
Solved in a private branch, will push in the morning CEST

Issue #1519, opened 2015-09-02
Extend statement with & selector causes crash


bassplayer7:
I have this situation in one of my SCSS stylesheets which I'm compiling with libsass (with CodeKit 2):
![_controls_scss_-_river_valley_woodworks_-____sites_river_valley_woodworks_](https://cloud.githubusercontent.com/assets/1171147/9639511/0a84acb0-5172-11e5-8ac5-b49f9b732ef8.png)

At the time of writing, I didn't put it together that `@extend &` *will always extend the current selector* and as such doesn't make sense. However, CodeKit crashed with that and I'm guessing that it threw an unrecoverable exception. I suppose it is conceivable that others would make the same mistake that I did, and it would be nice to have an error saying that the "current select cannot be extended". Obviously, had it compiled or told me, I would have immediately realized the error of my thought at that moment.

mgreter:
I can confirm that this segfaults. Fix seems simple as we just need to error when we extend '&'

```
Can't extend &: can't extend parent selectors
```

@bassplayer7 thanks for the report! I just don't understand why quite a few people actually like to post pictures of code instead of just pasting the code (it keeps us from transcribing the test case) :confounded: 

```scss
foo {
  @extend &;
}
```

bassplayer7:
@mgreter, I'm sorry about that. I absolutely should have included the code: I'm not sure why it didn't cross my mind.

------

Here's where it happened to me, in case it affects your error message:

    .column {
      & {
        width: 31%;
        margin: 1rem 1.5%;
        display: inline-block;
        vertical-align: top;
        
        &-half {
          @extend &;
        }
      }
    }

That was perhaps what made me miss the error of my way originally is that it was not extending a parent selector. I was trying to extend `.column`, but would have extend `.column-half` unintentionally. I'm not sure if that would impact the error message. 

Thanks for working on this! I appreciate libsass very much.

mgreter:
Not sure if you expect this to compile, but ruby sass gives this error (and libsass segfaults):

```
Error: Can't extend &: can't extend parent selectors
        on line 9 of foo.scss
  Use --trace for backtrace.
```

So we need add this check to libsass, but the given code is not valid sass ...

bassplayer7:
Ok. Great. Right, I didn't expect the SCSS to compile, I was just commenting on the error message. And, `&` could, and often would be a parent selector, so that message would make sense. Again, I really appreciate your work on that.

mgreter:
[...]

PR #1524, opened 2015-09-03
Don't segfault on & {}

saper:
Eases https://github.com/sass/libsass/issues/1519

Now "& {}" just returns an empty CSS - whereas Ruby
reports:

    Error: Base-level rules cannot contain the parent-selector-referencing character '&'.
            on line 1 of /home/saper/sw/nodes-sass-problems/1120/test.scss
      Use --trace for backtrace.
    
Another example:

    foo {
      @extend &;
    }

generates this error message:

    Error: nested selectors may not be extended

whereas Ruby Sass gives:

    Error: Can't extend &: can't extend parent selectors
            on line 2 of /home/saper/sw/nodes-sass-problems/1120/test1.scss
      Use --trace for backtrace.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -911,10 +911,9 @@ namespace Sass {
                 *retval << s;
               }
             }
-          }
+          } else {
           // have no tails but parents
           // loop above is inside out
-          else {
             for (size_t i = 0, iL = parents->length(); i < iL; ++i) {
               Complex_Selector* parent = (*parents)[i];
               Complex_Selector* s = parent->cloneFully(ctx);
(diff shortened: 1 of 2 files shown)


saper:
I can't say I fully understand what shall [Complex_Selector::parentize](https://github.com/sass/libsass/blob/e0effd9ba020ed913e0a266e1d1ce74ac28958f4/src/ast.cpp#L883-L971)()  return - I sense this method should fail in some cases right away.

This change just covers up the problem, probably.

mgreter:
This should be covered by https://github.com/sass/libsass/pull/1527

saper:
Superseded by #1527

PR #1527, opened 2015-09-03
Fix two segfaults with parent selectors

mgreter:
Slightly tweaks the way we parse selectors in regard to
the parent selector. We only add one if we are not parsing
inside some root context (passed by boolean argument).

Fixes https://github.com/sass/libsass/issues/1519

[test-1] [1]
```css
foo {
  @extend &;
}
```

[test-2] [2]
```css
& {}
```

[test-3] [3]
```css
& foo {
  bar: baz;
}
```
[test-4] [4]
```css
foo & {
  bar: baz;
}
```

[1]: http://libsass.ocbnet.ch/srcmap/#Zm9vIHsKICBAZXh0ZW5kICY7Cn0=
[2]: http://libsass.ocbnet.ch/srcmap/#JiB7fQ==
[3]: http://libsass.ocbnet.ch/srcmap/#JiBmb28gewogIGJhcjogYmF6Owp9
[4]: http://libsass.ocbnet.ch/srcmap/#Zm9vICYgewogIGJhcjogYmF6Owp9

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1125,13 +1125,18 @@ namespace Sass {
     // Check every rhs selector against left hand list
     for(size_t i = 0, L = length(); i < L; ++i) {
       if (!(*this)[i]->head()) continue;
-      if ((*this)[i]->combinator() != Complex_Selector::ANCESTOR_OF) continue;
       if ((*this)[i]->head()->is_empty_reference()) {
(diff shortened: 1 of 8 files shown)


saper:
Yes, thank you!

xzyfer:
Specs added and activated in https://github.com/sass/sass-spec/pull/689

Issue #1262, opened 2015-06-05
@import's processed breadth-first instead of depth-first


pkmiec:
I'm looking to speed up my asset compilation by switching from http://github.com/rails/sass-rails to http://github.com/bolandrm/sassc-rails. I have created http://github.com/appfolio/sassc-import_once to provide "import once" functionality that's normally provided by http://github.com/Compass/compass/tree/master/import-once for Sass.

The problem is that ```@import``` seems be processed by sassc / libsass breadth-first instead of depth-first. Here is how why that matters,
```scss
/* component.scss */
@import "component/button";
@import "component/color";
``` 

```scss
/* component/button.scss */
@import "component/color";
$button-default-text-color: $text-color;
``` 

```scss
/* component/color.scss */
$text-color: #303030;
``` 

Using sassc (with my sassc-import_once) produces,
```
SassC::SyntaxError: Error: Undefined variable: "$text-color".
        on line 3 of test/fixtures/component/_button.scss
>> $button-default-text-color: $text-color;
```

The reason is that files are imported in the following order,
```
imports   component/button from component.scss
imports   component/color  from component.scss
imports   component/color  from component/_button.scss
```

Notice the ```component/color``` from ```component.scss``` is imported first ... this causes the "import once" from re-importing ```component/color``` next it is seen. 

The depth-first approach would import files in the following order,
```
imports   component/button from component.scss
imports   component/color  from component/_button.scss
imports   component/color  from component.scss
```

I have a failing test that replicates this case,
1. clone sassc-import_once
2. bundle
3. rake

edit: fixed typos

pkmiec:
This is related to https://github.com/sass/libsass/issues/872

xzyfer:
I've done some experimenting here and as far as I can tell Libsass acts the same as Ruby Sass. I've compared SassC, Ruby Sass, and node-sass async render (with and without a custom importer). The behaviour is identical.

Please confirm this is not a logic bug in your plugin, otherwise provide a test using SassC directly that show this error.

saper:
I think this import-once approach may have some trouble due to #1160. Would love to see a self-contained test.

pkmiec:
Thanks. I'll work on getting a self-contained example today.

xzyfer:
Ok I've been able to confirm that although the imports are being resolved depth-first (i.e. the output is correct), the custom importers are being execute breadth-first. This looks like an issue with custom importers and not Libsass.

saper:
[...]

PR #1528, opened 2015-09-03
Parse imports right away after discovery

mgreter:
Makes import resolve depth-first instead of breadth-first
Fixes https://github.com/sass/libsass/issues/1262


xzyfer:
:+1:

Issue #1171, opened 2015-05-07
Compilation and length issues for appended variable-length arglist


davidkpiano:
Compilation runs forever with the below code:

```scss
@function foo($initial, $args...) {
    $args: append($args, 3);

    @return nth($args, 1);
}

@debug foo(1, 2);

// EXPECTED: 2

// ACTUAL (Ruby Sass v.3.4.13): 2
// ACTUAL (Node Sass v.3.0.0 and lower): << never finishes compiling, no output >>
```

Also, potentially related - an appended variable-length arglist should have an accurate length given by `length()`

```scss
@function foo($initial, $args...) {
    $args: append($args, 3);

    @return bar($initial, $args...);
}

@function bar($args...) {
    @debug $args; // DEBUG: 1, 2, 3 
                  // (in both node-sass and Ruby Sass)
    
    @return length($args);
}

@debug foo(1, 2);

// EXPECTED: 3

// ACTUAL (Ruby Sass v.3.4.13): 3
// ACTUAL (Node Sass v.3.0.0 and lower): 2
```

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/364

xzyfer:
@davidkpiano is this part of SassDash?

davidkpiano:
@xzyfer Yep, `_defaults()` and `_assign()` (as well as other functions) rely on this behavior. :smile:

xzyfer:
Ok cool, prioritizing this to 3.2.3.

mgreter:
First issue is covered by the PR. Maybe we should create a new one for the second, since it is related to how we handle lists. IMO the second sample can probably be shortened down to:
```css
foo {
  bar: inspect(( (1, 2), (3, 4) ));
}
```

ruby sass:
```css
foo {
  test-01: (1, 2), (3, 4); }
```


libsass:
```css
foo {
  test-01: 1, 2, 3, 4; }
```

davidkpiano:
@mgreter Sure, want me to make a new issue with the second example? I feel like it might be an issue with arglists from variable-length arguments, rather than just lists (correct me if I'm wrong), hence the specific example.

mgreter:
If you can confirm the first one is gone, sure! Close this one then and create a separate issue for the list append bug (which is caused by the sample I posted above, I'm very sure about that).

Ruby sass appends `3` to this list `(1, 2)`, which results in `(1, 2), 3`, which has a length of 2.
[...]

PR #1529, opened 2015-09-04
Fix sass function bug with restargs and nested arglists

mgreter:
Fixes https://github.com/sass/libsass/issues/1171

### DIFF
--- a/src/bind.cpp
+++ b/src/bind.cpp
@@ -128,13 +128,37 @@ namespace Sass {
             List* ls = dynamic_cast<List*>(a->value());
             // skip any list completely if empty
             if (ls && ls->empty()) continue;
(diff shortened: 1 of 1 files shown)


am11:
I can confirm that with this change, the output is consistent with Ruby-Sass output for the issue described in #1171.

However, the out-of-range error message from R-Sass reads like:

> List index is -4 but list is only 3 items long for `nth'

While LibSass says:

> Error: index out of bounds for `nth($list, $n)`

mgreter:
Errors have never been exactly like ruby sass. I only lately added a similar error reporting than ruby sass, so I guess this error just hasn't been upgraded, or did it say anything else before!?

Issue #1530, opened 2015-09-04
regression in importing behavior


drewwells:
https://github.com/sass/libsass/pull/1528 caused a regressions in custom importer functions. Libsass attempts to resolve every import; failing to find the import on the file system, it throws an error. Previously, libsass would call registered imports. Registered importers set via `option_set_c_importers` are not called.

Reverting the above PR fixes the tests that are failing.

mgreter:
I cannot reproduce with perl-libsass, my spec tests with custom importer didn't show any regression!

Can you provide a sampe that shows the behavior?

drewwells:
It only affects the file compiler

mgreter:
Not sure when I get around to look into it. If you could provide another standalone C example it would greatly help me to get a fix faster, as I will have to produce a test case for myself too. IMHO it is not a release-blocker since the feature in question is still marked experimental.

drewwells:
This is definitely a blocker for me as it makes impossible to handle virtual imports that don't point to real files. It also worked just fine before this refactor.

Is there an example of using a custom importer. It would take me a long time to rewrite this behavior in c.

mgreter:
https://github.com/sass/libsass/wiki/API-Sass-Importer-Example

drewwells:
Possibly related, using the following as styles.scss in that example:

```
@import "/tmp/styles.scss";
```

Results in
```
remote {
  color: red; }

remote {
  color: red; }
```

drewwells:
Running this example, you can see the custom importer is never executed. It will work if you switch to data compiler (untested in C).

```
#include <stdio.h>
#include <string.h>
#include "sass_context.h"

Sass_Import_List sass_importer(const char* path, Sass_Importer_Entry cb, struct Sass_Compiler* comp)
{
  // get the cookie from importer descriptor
  void* cookie = sass_importer_get_cookie(cb);
  Sass_Import_List list = sass_make_import_list(1);
  printf("loooking up %s\n", path);
  char* nope = "nope{ color: blue; }";
  if (strcmp(path,"noop")!=0) {
    list[0] = sass_make_import_entry("nope.scss", strdup(nope), 0);
  } else {
    printf("no match\n");
    list[0] = sass_make_import_entry("", strdup(""), 0);
  }
  printf("exit\n");
  return list;
}

int main( int argc, const char* argv[] )
{

  // get the input file from first argument or use default
  const char* input = argc > 1 ? argv[1] : "styles.scss";

[...]

PR #1533, opened 2015-09-05
Fix too early parser execution for initial file

mgreter:
Fixes https://github.com/sass/libsass/issues/1530


drewwells:
:+1: fixed my regressions

Issue #1271, opened 2015-06-11
Char escaping doesn't follow Sass 3.3 or 3.4 ways


kaelig:
See http://sassmeister.com/gist/cd6d8d263d555bca6cf9

Neither of these things return the correct value in LibSass:
```scss
$character-code: f102;

test {

  /* Expected: "\f102" */

  /* Sass 3.4 */
  content: unquote("\"\\#{$character-code}\"");

  /* Sass 3.3 */
  content: str-slice("\x", 1, 1) + $character-code;

}
```

kaelig:
If a workaround exists I'd be interested to know about it.

HugoGiraudel:
On top:

```scss
$value: str-slice(unquote('\\'), 1);
```

`/` in Ruby Sass, `//` in LibSass.

HugoGiraudel:
@kaelig: This should do the trick on LibSass:

```scss
.foo {
  content: #{'\"\\' + $character-code + "\""};
}
```

kaelig:
Thanks Hugo, although LibSass renders the actual unicode char (`content: ;`). Is that what is supposed to happen? Does it work everywhere?

HugoGiraudel:
Reported in #1231.

JLHwung:
Another situation might be relevant to this issue, see http://sassmeister.com/gist/6b58d3d5b01bb2a932c6

This snippet is from bootstrap-sass-official and it breaks the apperance of breadcrumb component.

xzyfer:
Fixed via #1534

Issue #1231, opened 2015-05-21
PUA code literal rendering


indrekpaas:
Using `compressed` mode
```scss
div::before {
  content: #{"\""+\e600+"\""};
}
```
renders a literal character
```css
div::before{content:}
```
with LibSass [(demo)](https://gist.github.com/indrekpaas/7e125d0873e3e8c2bce7), as opposed to
```css
div::before{content:"\e600"}
```
with Ruby Sass [(demo)](https://gist.github.com/indrekpaas/c37322e7bb873e4edaf1).

xzyfer:
Fixed via #1534

PR #1534, opened 2015-09-05
Hotfix for interpolation adding additional escapes

### LINKED ISSUE #1115: variable interpolation causes additional string escaping

mgreter:
This fixes one of the last urgent issues with interpolations. I have tried to solve it multiple times but I fail to see any other logic than implemented here (look ahead if we expect the final string to be a quoted string or not). With this patch we pass these additional spec tests:

- libsass-todo-issues\issue_1115 (Fixes #1115)
- parser\interpolate\29_binary_operation\todo\06_escape_interpolation
- parser\interpolate\30_base_test\todo\06_escape_interpolation

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -1424,10 +1424,10 @@ namespace Sass {
   ////////////////////////////////////////////////////////
   class String_Quoted : public String_Constant {
   public:
-    String_Quoted(ParserState pstate, std::string val, char q = 0)
+    String_Quoted(ParserState pstate, std::string val, char q = 0, bool keep_utf8_escapes = false)
     : String_Constant(pstate, val)
     {
-      value_ = unquote(value_, &quote_mark_);
+      value_ = unquote(value_, &quote_mark_, keep_utf8_escapes);
       if (q && quote_mark_) quote_mark_ = q;
     }
     virtual bool operator==(const Expression& rhs) const;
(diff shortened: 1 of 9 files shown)


mgreter:
Added another small commit that fixes https://github.com/sass/libsass/issues/1271

mgreter:
And another commit added that fixes #1231

xzyfer:
Nice on @mgreter

PR #1551, opened 2015-09-16
Implement static parameters for built-in functions

mgreter:
Brings about 15% performance gain for the spec test suite!

I recently made some more performance assessments. One of the more obvious performance drawbacks is that we currenlty parse the function prototypes everytime we create a new context. This overhead is "noticable" (in term of overall compile time) for small inputs and we also do it over and over again for long running processes. 

The easiest way to fix it is to implement the prototypes (`Parameters`) directly as AST_Nodes. While it is not overly complex, it does add some additional complexity when adding new internal functions. This PR adds static prototypes for all internal functions, but I keept the code for parsing the `Parameters` node if the Parameters are not statically implemented (so it's actually optional).

It may add some complexity but IMO it's worth for the performance we get! I tried to research some alternative methods to achieve this, but doing it by hand was IMO the most portable option. Most functions are straight forward and there only a few cases with `rest` arguments.

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -711,16 +711,15 @@ namespace Sass {
       return includes;
   }
 
-  void register_function(Context& ctx, Signature sig, Native_Function f, Env* env)
+  void register_function(Context& ctx, std::string name, Signature sig, Parameters* params, Native_Function f, Env* env)
   {
-    Definition* def = make_native_function(sig, f, ctx);
+    Definition* def = make_native_function(name, sig, params, f, ctx);
     def->environment(env);
     (*env)[def->name() + "[f]"] = def;
   }
-
-  void register_function(Context& ctx, Signature sig, Native_Function f, size_t arity, Env* env)
+  void register_function(Context& ctx, std::string name, Signature sig, Parameters* params, Native_Function f, size_t arity, Env* env)
   {
-    Definition* def = make_native_function(sig, f, ctx);
+    Definition* def = make_native_function(name, sig, params, f, ctx);
     std::stringstream ss;
     ss << def->name() << "[f]" << arity;
     def->environment(env);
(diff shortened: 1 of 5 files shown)


src/functions.cpp
```diff
@@ -29,20 +29,22 @@
...
   using std::endl;
 
-  Definition* make_native_function(Signature sig, Native_Function func, Context& ctx)
+  Definition* make_native_function(std::string name, Signature sig, Parameters* params, Native_Function func, Context& ctx)
   {
-    Parser sig_parser = Parser::from_c_str(sig, ctx, ParserState("[built-in function]"));
-    sig_parser.lex<Prelexer::identifier>();
-    std::string name(Util::normalize_underscores(sig_parser.lexed));
-    Parameters* params = sig_parser.parse_parameters();
+    if (params == 0) {
```
mgreter: Skip the prototype parsing if `Parameters` are passed to us to gain performance.

src/context.cpp
```diff
@@ -477,101 +477,101 @@ namespace Sass {
   {
     using namespace Functions;
     // RGB Functions
-    register_function(ctx, rgb_sig, rgb, env);
+    register_function(ctx, "rgb", rgb_sig, &rgb_params, rgb, env);
```
mgreter: We now pass the static `Parameters` for each function (`&rgb_params`)

src/functions.cpp
```diff
@@ -1125,6 +1362,10 @@ namespace Sass {
     }
 
     Signature min_sig = "min($numbers...)";
+    Parameter min_numbers(ParserState("[min-numbers]"), "$numbers", 0, true);
+    Parameters min_params(ParserState("[min]"), {
+      &min_numbers
+    }, false, true);
```
mgreter: Here is a function with `rest` arguments

akhleung:
Gains would be significant for the test suite because it's a lot of tiny inputs -- but I doubt this is representative of most real-world inputs. Just wondering if the added complexity is worth it; the old way is (IMO) very readable, maintainable, and provides a bit of implicit documentation for the built-ins!

mgreter:
@akhleung I see the point, but the same also applies for long running processes that repeatedly create new contexts. I vote to bite this bullet since we do not add new functions very often, it is still not overly complex and this now really is the lowest hanging fruit to further improve performance. We still need to "old" prototype string for error reporting. We might could improve readability by using macros to declare the parameters, but I really would like to move this from runtime to compile time!

saper:
Too much redunancy, will be very error-prone. What about parsing the signature during construction and storing the parsed arguments in `Signature`?

mgreter:
Another compelling reason for me to do this: currently it's harder than needed to debug the parser. One can not just add print statements to the parser functions, since it will not only be invoked for the code you try to debug, but also to parse the prototypes of these functions. So you end up with way too much output and you either have to implement another guard (not hard, but I already did this about 5 times the last month, so I keep doing this over and over). So this would actually have helped me to speed up my development, despite adding a little bit of complexity. The code also didn't change a bit since then.

xzyfer:
I agree with @akhleung and @saper. I don't think this is worth it.

mgreter:
I voice strongly for including this. I have not seen any hard evidence for the claimed problems. I can work easily with it, it's optional, doesn't really change anyway and has lots of cases, so it should be easy to setup now functions once needed (and to repeat, old way is still there, completely optional).

saper:
The problem is this commit is adding 900 lines of redundant code, which previously fit in 100.

mgreter:
I'm going to make this call here and will merge this after the next release is made. @xzyfer I guess you may want to make 3.3.7 before releasing node-sass? I also want to split functions.cpp into multiple compile units, therefore I want to base that on the changes from here. Sorry that I could not convince you guys of the benefits, but I don't feel that the "redundancy" is a problem or lines of code per se.

xzyfer:
As stated by everyone else in this issue this is adding 900 virtually pointless LOC to, as you stated yourself, avoid occasionally adding a guard when debugging.

@saper @akhleung  and myself agree this change makes the code unnecessarily significantly harder to reason about, without adding any real benefit.

> I have not seen any hard evidence for the claimed problems

You have also not backed up your own claims. As @akhleung the spec runner is not indicative of real work use cases.

> old way is still there, so completely optional

Having the old way being completely optional is not a good a thing. Having two ways to do something is a bad thing.

> but I don't feel that the "redundancy" is a problem 

Redundant code is a problem.

xzyfer:
This current node-sass release will be addressing a Node 6 compatibility issue. We won't be conflating that with LibSass update also. Node Sass will land 3.3.7 in an upcoming release.

mgreter:
I will put this simple. If you guys would contribute more actual code to libsass it would be easier for me to accept your opinions here. But since I'm pretty much the only guy doing the heavy lifting on the code base for 3 years now (ok, I need to except @xzyfer here, since he does occasionally contribute bigger amounts of code).

The parsing of prototypes with internal parser on runtime hasn't tripped my workflow off for the first time. This would solve the issue I as a developer have from time to time. I agree it's not a very elegant solution, but it does work and nobody has to worry about it, as it just works (its dead simple). This is frankly one of the topics I sometimes wonder why I'm even putting in the hours I do!

PR #1555, opened 2015-09-19
Ruby Sass uses 9 spaces to indent stack frame

saper:


xzyfer:
Do you have an example. Locally it looks like Libsass and Ruby Sass but use 8 space. Could be something about my maching

```
// Ruby Sass
Error: Invalid CSS after "...nt 'Comic Sans'": expected "{", was ";"
        on line 30 of test.scss
  Use --trace for backtrace.

----
// Libsass
Error: property "font" must be followed by a ':'
        on line 30 of test.scss
>>   font 'Comic Sans';
   --^
```

saper:
Yes it is 8 spaces per stack frame and one space coming from somewhere. You can easily see this with sass/sass-spec#494

xzyfer:
>You can easily see this with

I could not. I did however see that for warning there is 9 spaces, which appears to be purely aesthetic for alignment.  However this change is not just for warnings.

saper:
Good point. New version of my error testing suite takes only god first line for comparison, so let's ditch this for now.

xzyfer:
> takes only god first line for comparison

:+1: this has been my approach in the past

PR #1582, opened 2015-10-05
Added the formatted error message to Sass_Context.error_json

xzyfer:
This was originally requested in https://github.com/sass/libsass/issues/938, and base work was done by @mgreter in https://github.com/sass/libsass/pull/958.

This is super useful for us in node-sass, as well as our downsteam implementors like grunt-sass (https://github.com/sindresorhus/grunt-sass/issues/201) and gulp-sass. 

Yes, implementors can easily implement something similar themselves. However I believe there is value both in having some level of consistency, and easing the integration process.

/cc @am11 @mgreter @sindresorhus

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -186,12 +186,6 @@ extern "C" {
     catch (Error_Invalid& e) {
       std::stringstream msg_stream;
       std::string cwd(Sass::File::get_cwd());
-      JsonNode* json_err = json_mkobject();
-      json_append_member(json_err, "status", json_mknumber(1));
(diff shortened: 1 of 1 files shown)


sindresorhus:
:+1: Yes, every consumer *could* do this, but they would probably do it badly and it would be sad to have so much duplicated code.

am11:
Thanks @xzyfer! 
:+1: to the idea of bringing consistency among the down-streams.

mgreter:
Just a note: not sure if we are guarenteed to always have something usefull in `e.pstate`.

PR #1594, opened 2015-10-11
Fix sizeof argument

unmole:


saper:
Wouldn't `sizeof(union Sass_Value *)` work as well?

unmole:
@saper It would. The `void*` just makes it really obvious that you only need enough memory to store a pointer.

xzyfer:
Thanks @unmole. I'm happy with either approach.

mgreter:
I cherry picked this in https://github.com/sass/libsass/pull/1609. Thanks for your contribution @unmole!

Issue #1567, opened 2015-09-21
Error with comments between media queries


tiago:
**test.scss**:
```scss
@media
screen,  /* something */
print {  /* whatever */
    body { line-height: 1.2 }
}
```
---
**Ruby Sass (3.4.18)**:
```css
@media screen, print {
  /* whatever */
  body {
    line-height: 1.2; } }
```
**SassC (3.3.0-beta1) with LibSass (3.3.0-beta2-2-g997c)**:
```bash
Error: media query expression must begin with '('
        on line 2 of test.scss
>> screen,  /* something */
   ------^
```

saper:
Test written https://github.com/sass/sass-spec/pull/510

nkovacs:
Is this related? 

```scss
a.asc:after {
    content: /*"\e113"*/ "\e151";
}

a.desc:after {
    content: /*"\e114"*/ "\e152";
}
```

Works fine with sass 3.4.14, breaks with libsass 3.2.5

xzyfer:
@nkovacs no that is not related. Your issue is fixed in the 3.4.0-beta.

mgreter:
More complete test case:
```scss
/* any */@media/* first */
/* screen */screen /*something */ , /* else */
/* not */not/* print */print /* final */ {  /* whatever */
    body { line-height: 1.2 }
}
```

PR #1598, opened 2015-10-13
Fix media-query parsing with block comments

mgreter:
Fixes https://github.com/sass/libsass/issues/1567
Also fixes a minor parsing bug with `not/and` keywords.
Lexer would have interpret `notscreen` as `not screen`.


xzyfer:
LGTM, please update the test case as mentioned in https://github.com/sass/libsass/issues/1567#issuecomment-147875058 before merging.

xzyfer:
Update specs in https://github.com/sass/sass-spec/pull/540

xzyfer:
Spec https://github.com/sass/sass-spec/pull/542

Issue #1590, opened 2015-10-08
If comparison inconsistency between ruby and lib sass


peterwilsoncc:
It's very minor but a valid comparison in rubySass required refactoring for libSass.

rubySass: `@if ( & == $sel ) {`

libSass: `@if ( "&" == $sel ) {`

Full mixin follows:

```sass
@mixin last-child($sel: null){
	// for use with the miniselectivzr code in the polyfill JS
	// standard file uses :last-child psuedo class
	// IE file uses the u-last-child utility class
	@if ( "&" == $sel ) {
		@if ( true == $is_ie8_file ) {
			&.u-last-child {
				@content;
			}
		} @else {
			&:last-child {
				@content;
			}
		}
	}
	@else {
		@if ( true == $is_ie8_file ) {
			#{$sel}.u-last-child {
				@content;
			}
		} @else {
			#{$sel}:last-child {
				@content;
			}
		}
	}
}
```

xzyfer:
Thanks for the report. Could you please provide a minimal snippet of Sass
code that exhibits this problem so we can debug it?

saper:
Simple test written https://github.com/sass/sass-spec/pull/536
Can I specify literal selectors somehow as the argument to the `@where` mixin?

saper:
@peterwilsoncc I think that "&" in libsass does not expose the parent selector, it just compares strings. Can you confirm this is your case as well? See [the test case](https://github.com/saper/sass-spec/blob/95053d9b3ed77304e0a071f1b624e7a6c27a3575/spec/libsass-todo-issues/issue_1590/quoted-ampersand-does-not-select/input.scss) and [the expected output from Ruby Sass](https://github.com/saper/sass-spec/blob/95053d9b3ed77304e0a071f1b624e7a6c27a3575/spec/libsass-todo-issues/issue_1590/quoted-ampersand-does-not-select/expected_output.css) taken from https://github.com/saper/sass-spec/commit/95053d9b3ed77304e0a071f1b624e7a6c27a3575

peterwilsoncc:
Thanks @saper, very much appreciated.

PR #1599, opened 2015-10-14
Improve parent selector handling

mgreter:
Fixes https://github.com/sass/libsass/issues/1590
- Improve compatibility when inspecting
- Improve handling with functions (append/nth)


xzyfer:
Spec activated - https://github.com/sass/sass-spec/pull/546

Issue #1584, opened 2015-10-05
`@at-root` fails to place `@content` block at root in mixin


lunelson:
Ran across this today. The previous issue where dynamic classes where not working with `@at-root` seems to be fixed, but normal usage of `@at-root` in mixins seems to be broken—or not yet fixed:

```scss
@mixin foo($out: false) {
  @if $out {
    @at-root { @content; }
  }
}

@mixin bar() {
  @at-root { @content; }
}

@mixin baz($string) {
  @at-root .#{$string} { @content; }
}

.test {
  @include foo(true) {
    .nest1 {
      color: red;
    }
  }
  @include bar() {
    .nest2 {
      color: green;
    }
  }
  @include baz('nest3') {
    color: blue;
  }
  @at-root {
    .nest4 {
      color: yellow;
    }
  }
}
```
expected:
```css
.nest1 {
  color: red;
}
.nest2 {
  color: green;
}
.nest3 {
  color: blue;
}
.nest4 {
  color: yellow;
}
```
actual:
```css
.test .nest1 {
  color: red; }

.test .nest2 {
  color: green; }

.nest3 {
  color: blue; }

.nest4 {
  color: yellow; }
```

PR #1608, opened 2015-10-19
Improve at-root handling with mixins

mgreter:
Fixes https://github.com/sass/libsass/issues/1584


xzyfer:
Spec activated - https://github.com/sass/sass-spec/pull/545

Issue #1557, opened 2015-09-19
Capitalization with media query rules breaks compilation


scniro:
@media ALL and (max-width: 100px) {
    }

will compile fine, but

    @media ALL AND (max-width: 100px) {
    }

breaks with the following...

>expected '{' in media query

saper:
This is what  Ruby Sass is doing:

````
Error: Invalid CSS after "...-width: 100px) ": expected media query list, was "{"
        on line 1 of test
  Use --trace for backtrace.
````

so this is intentional.

scniro:
Definitely not intentional, and definitely not a sass issue. See [sassmeister gist](http://sassmeister.com/gist/13a2edd814c12c4ec92d). This will work fine on the current version of sass (3.4.14), but selecting libsass 3.2.5 will reproduce the error (control panel setting)

saper:
Sorry, but the example you have originally given was incomplete. Now it is obvious, what's wrong. Thanks.

scniro:
Sorry for the incomplete first example. I did have a whole SO question/thread on this but I deleted it for it was not totally appropriate there. I wanted to fork the gist with the error but that site will not allow you to fork if there is an error (so I included instructions to repro). Thanks for understanding!

saper:
Ok, now it's 404. Can you post a full test here? I'd like to add it to the test suite.

saper:
Ok, got test ready for the test suite.

xzyfer:
This is due to our lexer being case sensitive for keywords like and, or and new.

PR #1613, opened 2015-10-21
Implement case insensitive keyword matcher

mgreter:
Fixes https://github.com/sass/libsass/issues/1557


xzyfer:
Spec activated https://github.com/sass/sass-spec/pull/571

PR #1617, opened 2015-10-22
Change C-API to pass compiler to custom functions

mgreter:
In preparation for import stack access (to be implemented) => https://github.com/sass/node-sass/issues/895


xzyfer:
:+1:

PR #1621, opened 2015-10-24
Add `Sass::round` to work around compiler issues

mgreter:
Needs sass/sass-spec#555 to be merged to pass!

It seems with some compilers (cygwin) round is not defined
within the `std` namespace, but in the root namespace.
Working around this issue by `using namespace std` and
then using `::round` function from the root namespace.

### DIFF
--- a/script/ci-build-libsass
+++ b/script/ci-build-libsass
@@ -93,6 +93,13 @@ then
 
   JSON=$(curl -L -sS https://api.github.com/repos/sass/libsass/pulls/$TRAVIS_PULL_REQUEST)
 
+  if [[ $JSON =~ "API rate limit exceeded" ]];
+  then
+    echo "Travis rate limit on github exceeded"
+    echo "Retrying via 'special purpose proxy'"
+    JSON=$(curl -L -sS http://libsass.ocbnet.ch/libsass-spec-pr.psgi/$TRAVIS_PULL_REQUEST)
+  fi
+
   RE_SPEC_PR="sass\/sass-spec(#|\/pull\/)([0-9]+)"
 
   if [[ $JSON =~ $RE_SPEC_PR ]];
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1864,17 +1864,17 @@ namespace Sass {
     // resolved color
     std::string res_name = name;
 
-    double r = round(cap_channel<0xff>(r_));
-    double g = round(cap_channel<0xff>(g_));
-    double b = round(cap_channel<0xff>(b_));
+    double r = Sass::round(cap_channel<0xff>(r_), precision);
+    double g = Sass::round(cap_channel<0xff>(g_), precision);
+    double b = Sass::round(cap_channel<0xff>(b_), precision);
     double a = cap_channel<1>   (a_);
 
     // get color from given name (if one was given at all)
     if (name != "" && name_to_color(name)) {
       const Color* n = name_to_color(name);
-      r = round(cap_channel<0xff>(n->r()));
-      g = round(cap_channel<0xff>(n->g()));
-      b = round(cap_channel<0xff>(n->b()));
+      r = Sass::round(cap_channel<0xff>(n->r()), precision);
+      g = Sass::round(cap_channel<0xff>(n->g()), precision);
+      b = Sass::round(cap_channel<0xff>(n->b()), precision);
       a = cap_channel<1>   (n->a());
     }
     // otherwise get the possible resolved color name
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -11,6 +11,7 @@
 #include "util.hpp"
 #include "expand.hpp"
 #include "utf8_string.hpp"
+#include "sass/base.h"
 #include "utf8.h"
 
 #include <cstdlib>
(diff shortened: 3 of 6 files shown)


saper:
will test on the old Solaris, too.

mgreter:
This now needs https://github.com/sass/sass-spec/pull/555 to pass!
Libsass now passes the specs for x86 and x64 builds (it failed on x86)!

saper:
I frankly doubt one can get stable results there when working with floating point

mgreter:
It seems to fix the discrepancy between latest ruby sass and libsass ...

This is what I was observing on mingw32 and cygwin32:
```
double r = ...
std::cerr << r << "\n"; => 25.5
std::cerr << round(r) << "\n"; => 25
```

So using the double did not work correctly anyway ...

mgreter:
Btw. I'm currently debugging libsass through the perl test matrix:
http://matrix.cpantesters.org/?dist=CSS-Sass%203.3.0_02
This clearly indicates we currently have a problem on 32bit.

saper:
One cannot check for floating point results by comparing numbers for equality, it should be `| actual_value - expected_value | < ε`, where ε is a very small floating point number...

mgreter:
The problem is about rounding, not comparing numbers !? :confused:

saper:
The integer color values come from floating point calculations, don't they?

mgreter:
Nope, the go through the hsla conversion, which uses double

saper:
double is floating point too (exponent + mantissa)

mgreter:
Sorry, thought you meant `float` explicitly ...

saper:
double has more precision, but that does not help really when converting to integers.

mgreter:
Results should gradually come in over the next few days:
http://matrix.cpantesters.org/?dist=CSS-Sass%203.3.0_03

xzyfer:
The cause of the failing colour specs is that Ruby Sass added epsilon rounding (as @saper suggested) in 3.4.15. We should do the same thing. We added epsilon round for the `random` function in 3.3.

xzyfer:
Wen shouldn't use `std::round` for rounding in colour conversions. It won't predictably pass tests.

https://github.com/sass/sass/commit/2c537b14a290215c6e869cf4a8a86f5f67102da8
https://github.com/sass/sass/commit/14ec0d077bcc57f480f482c4fab6b9042153c047

mgreter:
Actually the real fix on the ruby side seems to be https://github.com/sass/sass/commit/4e3e1d5684cc29073a507578fc977434ff488c93.
So feel free to add to this PR to implement it the same way.
Seems like truncating it to a float was enough to pass the specs.
Looks like we could just add/subtract 0.00001 before passing it through round.

mgreter:
Made some more tests and indeed ruby sass uses some IMHO strange rounding magic:

```css
foo {
  pos-1: round(0.500001);
  pos-2: round(0.50001);
  pos-3: round(0.49999);
  pos-4: round(0.499999);
  neg-1: round(-0.500001);
  neg-2: round(-0.50001);
  neg-3: round(-0.49999);
  neg-4: round(-0.499999);
}
```

Sass 3.4.19 yields:
```css
foo {
  pos-1: 1;
  pos-2: 1;
  pos-3: 0;
  pos-4: 1;
  neg-1: 0;
  neg-2: 0;
  neg-3: 0;
  neg-4: 0;
}
```

This PR yields:
```
out: foo {
  pos-1: 1;
  pos-2: 1;
  pos-3: 0;
  pos-4: 0;
  neg-1: -1;
  neg-2: -1;
  neg-3: 0;
  neg-4: 0;
}
```

I guess with 3.4.20 (https://github.com/sass/sass/commit/4e3e1d5684cc29073a507578fc977434ff488c93) the negative side will be fixed. Anyway, it feels like this epsilon should at least consider the precision set in the options?

mgreter:
OK, seems easy enough to get what ruby sass does:
```
if (val < 0) val -= std::pow(0.1, 6);
else if (val) val += std::pow(0.1, 6);
return ::round(val);
```

Not sure why ruby sass does it the way as seen in https://github.com/sass/sass/commit/4e3e1d5684cc29073a507578fc977434ff488c93, but this should IMO give the same results:

```css
out: foo {
  pos-1: 1;
  pos-2: 1;
  pos-3: 0;
  pos-4: 1;
  neg-1: -1;
  neg-2: -1;
  neg-3: 0;
  neg-4: -1;
}
```

mgreter:
@xzyfer I have simplified the handling of custom spec test runs with CI. Also added the same logic to appveyor. I also made the epsilon respect the precision option. IMO this is ready to be merged!

xzyfer:
https://github.com/sass/sass-spec/pull/555 is problematic. It _cannot_ be merged until we bump sass-spec to at least 3.4.15. Otherwise sass-spec will fail on every PR.

This _could_ really blow out the scope of something should in 3.3.2. We need to look into many active specs break under 3.4.15, before moving forward here. I'll look into this tonight.

xzyfer:
@mgreter please rebase this off master and squash the remaining two commits. I'm evaluating the feasibility of this landing in 3.3.2.

xzyfer:
Thanks @mgreter. Bumping to 3.4.19 doesn't fail many tests based on output. However many now fail on  due to added deprecation warnings. This seems like a big job. Might be better scoped to 3.4.

mgreter:
IMO this is quite important since we currently fail the specs for 32bit builds

xzyfer:
These specs don't fail the 32bit builds in CI. Merging this would fail all our CI builds for either LibSass or Sass Spec. We can't merge this yet.

mgreter:
It does fail [reliably for me] [1] on 32bit and [Travis] [2] / Appveyor only provide 64bit envs as far as I know.

[1]: http://matrix.cpantesters.org/?dist=CSS-Sass%203.3.0_02;os=gnukfreebsd;perl=5.16.2;reports=1
[2]: https://github.com/travis-ci/travis-ci/issues/986#issuecomment-124141683

xzyfer:
My mistake. Presumably the failures are just rounding errors. If so they affect is imperceivable to users. The point still stands that merging this would break CI for us. This has been this way for a long time, it can wait until we're ready.

mgreter:
This should now pass with specs set at 4.9.16! Didn't yet try to go higher, can be done later.

PR #1623, opened 2015-10-25
[RFC] Add back gcc 4.4 compatibility

mgreter:
Since I get quite a few build fails from [cpan test matrix] [1] for gcc 4.4, I tried how difficult it is to bring back gcc 4.4 compatibility based upon some previous work. It actually is not too hard to get libsass compiled with gcc 4.4 (and passing the spec tests). I want to put this PR into consideration, since I would like to apply the `replace-range-for-loops` automatically when gcc 4.4 is detected by perl-libsass. Falls under "nice to have", but changes seem rather minimal.

Feedback welcome

[1]: http://matrix.cpantesters.org/?dist=CSS-Sass%203.3.0_02

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -242,7 +242,7 @@ namespace Sass {
     virtual void adjust_after_pushing(std::pair<Expression*, Expression*> p) { }
   public:
     Hashed(size_t s = 0) : elements_(std::unordered_map<Expression*, Expression*>(s)), list_(std::vector<Expression*>())
-    { elements_.reserve(s); list_.reserve(s); reset_duplicate_key(); }
+    { /* elements_.reserve(s); */ list_.reserve(s); reset_duplicate_key(); }
     virtual ~Hashed();
     size_t length() const                  { return list_.size(); }
     bool empty() const                     { return list_.empty(); }
@@ -1096,7 +1096,7 @@ namespace Sass {
       if (hash_ == 0) {
         hash_ = std::hash<std::string>()(name());
         for (auto argument : arguments()->elements())
-          hash_combine(hash_, argument->hash());
+        { hash_combine(hash_, argument->hash()); }
       }
       return hash_;
     }
@@ -1363,7 +1363,7 @@ namespace Sass {
     {
       if (hash_ == 0) {
         for (auto string : elements())
-          hash_combine(hash_, string->hash());
+        { hash_combine(hash_, string->hash()); }
       }
       return hash_;
     }
--- a/script/replace-range-for-loops.pl
+++ b/script/replace-range-for-loops.pl
@@ -1,0 +1,42 @@
+use strict;
+use warnings;
+
+use File::Slurp;
+use File::Basename;
+use File::Spec::Functions;
+
+my $root = $ENV{'SASS_LIBSASS_PATH'} || catfile(dirname($0), '..');
+
+sub process($)
+{
+
+	my $count = 0;
+	my ($file) = @_;
+
+	my $cpp = read_file($file, { binmode => ':raw' });
+
+	my $org = $cpp;
+
+	my $re_decl = qr/(?:const\s*)?\w+(?:\:\:\w+)*(?:\s*[\*\&])?/;
+	my $re_val = qr/\w+(?:\(\))?(?:(?:->|\.)\w+(?:\(\))?)*/;
+
+	$cpp =~ s/for\s*\(\s*($re_decl)\s*(\w+)\s*:\s*(\(\*?$re_val\)|$re_val)\s*\)\s*{/
+		$count ++;
+		"for (auto __$2 = $3.begin(); __$2 != $3.end(); ++__$2) { $1 $2 = *(__$2); ";
+	/gex;
+
+	return if $org eq $cpp || $count == 0;
+
+	warn sprintf "made %02d replacements in %s\n", $count, $file;
+
+	write_file($file, { binmode => ':raw' }, $cpp);
+
+}
+
+my $rv = opendir(my $dh, catfile($root, "src"));
+die "not found ", catfile($root, "src") unless $rv;
+while (my $entry = readdir($dh)) {
+	next if $entry eq "." || $entry eq "..";
+	next unless $entry =~ m/\.[hc]pp$/;
+	process(catfile($root, "src", $entry));
+}
--- a/src/debugger.hpp
+++ b/src/debugger.hpp
@@ -325,7 +325,7 @@ inline void debug_ast(AST_Node* node, std::string ind, Env* env)
     std::cerr << " " << block->tabs() << std::endl;
     debug_ast(block->media_queries(), ind + " @ ");
     // std::vector<std::string>         files_;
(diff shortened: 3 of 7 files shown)


saper:
I have #995 and #996 to for this with autoconf. 4.4 has no issues with variadic templates?

mgreter:
I was able to get it to compile with gcc 4.4.2, so it seems it had proper support for variadic templates. Tested it via various old [strawberry perl portable installs for windows] [1]. Might be that I use different cflags on the perl-libsass install, because I think this posed a problem when I tried some time ago.

[1]: http://strawberryperl.com/releases.html

ArcticSnowman:
This is essential to use this module on Centos6 based hosts, in corporate environments

xzyfer:
How so? I routinely build LibSass on centos 5. Updating GCC is trivial
enough.

xzyfer:
IMHO we cannot do this without also running GCC 4.4 in CI as well

ArcticSnowman:
In our corporate environment, we are required to stick to the officially supported Centos6 releases of GCC, which is stuck at 4.4.7.

src/ast.hpp
```diff
@@ -309,7 +309,7 @@ namespace Sass {
     virtual void adjust_after_pushing(std::pair<Expression*, Expression*> p) { }
   public:
     Hashed(size_t s = 0) : elements_(ExpressionMap(s)), list_(std::vector<Expression*>())
-    { elements_.reserve(s); list_.reserve(s); reset_duplicate_key(); }
+    { /* elements_.reserve(s); */ list_.reserve(s); reset_duplicate_key(); }
```
saper: #995 adds an autoconf macro here:

+#ifdef HAVE_CXX_UNORDERED_MAP_RESERVE
+      elements_.reserve(s); list_.reserve(s);
+#endif

For non-autoconf builds it stays undefined, so no harm here.

mgreter:
I have rebased this to latest master and fixed the remaining issue with mac clang. I pretty much would like to get this merged, as it would allow perl-libsass to autodetect this and apply the last "patch" in form of the script in the second commit to the source tree. To get the git tree 4.4 compatible, we would need to get rid of all `for(auto el : container)` loops, and I actually like that feature a lot. So for me an automated way to "unroll" these loops is good enough. 

@ArcticSnowman would that be good enough to support your use case? It does add the dependency on the perl script, but perl should be widely available or you can always generate the gcc 4.4 compatible source tree somewhere else and transfer it (or generate a patch to apply).

Since the proof is in the pudding (libsass via strawberry perl and gcc 4.4.3 on windows):

```cmd
mkdir libsass-gcc-4.4
cd libsass-gcc-4.4
wget -c http://strawberryperl.com/download/5.12.3.0/strawberry-perl-5.12.3.0-portable.zip
mkdir perl
cd perl
unzip ..\strawberry-perl-5.12.3.0-portable.zip
portableshell.bat

git clone https://github.com/sass/perl-libsass.git
cd perl-libsass
git submodule update --init
cd libsass
git remote add mgreter https://github.com/mgreter/libsass.git
git fetch mgreter
git checkout -b compat/gcc-4.4 mgreter/compat/gcc-4.4
perl script\replace-range-for-loops.pl
cd ..
cpan ExtUtils::CppGuess
cpan Test::Differences
cpan Encode::Locale
gcc -v
cpan .
```

saper:
I also like the short `for` loops, but I think we should unroll them to avoid issues like somebody adding or changing some loop to the form that cannot be automatically unrolled with a script (like the ones patched above).

mgreter:
IMO any code change may add something gcc 4.4 incompatible (I did so with ltrim/rtrim). I guess there are always chances we need to fix up these incompatibilites at a later stage, and I don't see the short for loops as an exclusion here. And as a developer I really would like to keep the convenience.

xzyfer:
Please a documentation page for compiling with gcc 4.4 which at the very
least documents the Python dependency, the purpose of the Python script,
and it's caveats i.e. naive parsing.
On 2 May 2016 7:11 AM, "Marcel Greter" <notifications@github.com> wrote:

IMO any code change may add something gcc 4.4 incompatible (I did so with
ltrim/rtrim). I guess there are always chances we need to fix up these
incompatibilites at a later stage, and I don't see the short for loops as
an exclusion here. And as a developer I really would like to keep the
convenience.

—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub
<https://github.com/sass/libsass/pull/1623#issuecomment-216072995>

src/ast.hpp
mgreter: Ordered Map doesn't seem to support it anyway, so I guess this has resolved itself.

k-funk:
My shop only supports GCC 4.4.7 on our production machines. I'd love to see this get merged in, and the python library updated when it's ready.

Roguelazer:
+1 most production environments aren't going to support gcc5 for a while, it would be nice to keep Saas working without requiring crazy-bleeding-edge versions of things. Enterprise life-cycles are measured in years, not months...

xzyfer:
I'm open to a PR for cpplint style check in CI. With cinema level
guarantees I don't feel comfortable attempting gcc 4.4 support.

k-funk:
Any progress on this? Still seems pretty unreasonable that libsass can't be used on CentOS 6 (LOTS of shared-hosting boxes).

mgreter:
I'm regularly releasing [perl-libsass][1] with [gcc 4.4 support][4]. The [archive][2] contains gcc 4.4 compatible [libsass sources][3] (this branch with unrolled for loops). On compatible systems it should also automatically compile and install sassc plus libsass plugins. Unfortunately I wasn't able to convince other libsass maintainers to get this merged into master (see their comments/reasoning above), but you could try to apply the patches in this branch on any implementation (i.e. node-sass). Then you only need to execute `perl script\replace-range-for-loops.pl` (see example in one of my earlier comment) and compile/install the given implementation.

[1]: http://search.cpan.org/~ocbnet/CSS-Sass/
[2]: http://search.cpan.org/CPAN/authors/id/O/OC/OCBNET/CSS-Sass-3.4.4.tar.gz
[3]: http://cpansearch.perl.org/src/OCBNET/CSS-Sass-3.4.4/libsass/
[4]: http://matrix.cpantesters.org/?dist=CSS-Sass%203.3.6

glebm:
Do we still want to deal with non C++ 11 compilers in 2019?

mgreter:
I will and have updated this branch for perl-libsass and will do so in the future.
As long as the changes needed are manageable of course ...

glebm:
I'm curious, why does Perl require such an old GCC version?

mgreter:
> I'm curious, why does Perl require such an old GCC version?

There are quite a few CPAN testers that still use 4.4, there are even some that are still on gcc 4.2.
It might not be that many in real life, but it's still nice to have as many green as possible.
Btw. the patches are [included in perl-libsass](https://github.com/sass/perl-libsass/tree/develop/patches) itself and applied automatically if needed.

http://matrix.cpantesters.org/?dist=CSS-Sass

PR #1626, opened 2015-10-26
Refactor C++ context to expose call stack

mgreter:
This is a big refactor of the C++ context handling side. Basically we now directly use the C structs instead of converting them back and forth. Also made it more similar to the C-API by adding `Data_Context` and `File_Context`. We now also take care of initializing the options accordingly on the C++ side, so I was able to remove some stuff from the C-API.

I tried to streamline the import handling by adding specific classes for all different but similar representations of import, resources, etc. I think this will make the internas much clearer once it is used consistently throughout the code (I already got rid of `Sass_Queues` struct).

```c++
// requested import
class Importer {
  public:
    // requested import path
    std::string imp_path;
    // parent context path
    std::string ctx_path;
    // base derived from context path
    // this really just acts as a cache
    std::string base_path;
};

// a resolved include (final import)
class Include : public Importer {
  public:
    // resolved absolute path
    std::string abs_path;
};

// a loaded resource
class Resources {
  public:
    // the file contents
    char* contents;
    // conected sourcemap
    char* srcmap;
};

// parsed stylesheet from loaded resource
class StyleSheet : public Resources {
  public:
    // parsed root block
    Block* root;
};
```

Fixes https://github.com/sass/libsass/issues/1638

### DIFF
--- a/docs/api-function-example.md
+++ b/docs/api-function-example.md
@@ -5,8 +5,17 @@
 #include <stdint.h>
 #include "sass/context.h"
 
-union Sass_Value* call_fn_foo(const union Sass_Value* s_args, void* cookie)
+union Sass_Value* call_fn_foo(const union Sass_Value* s_args, Sass_Function_Entry cb, struct Sass_Compiler* comp)
 {
+  // get context/option struct associated with this compiler
+  struct Sass_Context* ctx = sass_compiler_get_context(comp);
+  struct Sass_Options* opts = sass_compiler_get_options(comp);
+  // get information about previous importer entry from the stack
+  struct Sass_Import* import = sass_compiler_get_last_import(comp);
+  const char* prev_abs_path = sass_import_get_abs_path(import);
+  const char* prev_imp_path = sass_import_get_imp_path(import);
+  // get the cookie from function descriptor
+  void* cookie = sass_function_get_cookie(cb);
   // we actually abuse the void* to store an "int"
   return sass_make_number((intptr_t)cookie, "px");
 }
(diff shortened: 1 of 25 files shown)


mgreter:
This is pretty much ready, but could use some more testing. My poor memory leak tests seem to indicate that everything is freed as needed. There is room for more improvements though.

mgreter:
Just to repeat how to access the import stack during custom functions:
```
union Sass_Value* call_sass_function(const union Sass_Value* s_args, Sass_Function_Entry cb, struct Sass_Compiler* compiler)
  struct Sass_Import* previous = sass_compiler_get_last_import(compiler);
  const char* prev_abs_path = sass_import_get_abs_path(previous);
  const char* prev_imp_path = sass_import_get_imp_path(previous);
```

mgreter:
//CC @xzyfer to fix eyeglass import once issue

xzyfer:
@mgreter I pulled out a bunch of the code that wasn't relevant larger refactor into #1662, #1663, #1664, #1665, #1666

This also appears to include the rounding refactor from #1621.

Could you please rebase off master, and also remove those sass rounding commits? 
Then I'll dig into this refactor. The scope now look small enough that this could land in 3.3.2.

mgreter:
Will do, actually the main commit is https://github.com/mgreter/libsass/commit/1a68364ea703b1152f5303efa64ac6e541bab46d which is still rather big.

xzyfer:
I'll work through it over the next couple days. Nothing has jumped out at me yet. This is probably one of the last things before 3.3.2. Could you also update the markdown docs as well?

mgreter:
There is nothing to update in docs for this PR??

mgreter:
Disabled the wiki, since info is outdated and master is now in github ... maybe we could use git 2.6 subtree feature to get wiki visibility on github back ...

xzyfer:
Sorry https://github.com/sass/libsass/pull/1626#issuecomment-151022461 made me think there was

xzyfer:
Can you also rebase out the rounding commits. They don't belong here.

mgreter:
Rebased, but will probably result in merge conflicts with the other "rounding" PR ...

xzyfer:
That's ok. The rounding PR wont land in 3.3.2 because updating sass-spec to a new Ruby Sass breaks too many specs. (mostly on error messages / deprecation warnings)

xzyfer:
LGTM. I'm not sold on the names of the new classes, purely because without the comments their purpose is unclear.

I would like to see https://github.com/sass/libsass/pull/1626#issuecomment-151022461 added to documentation somewhere as part of this PR. Then feel free to :ship:

mgreter:
@xzyfer IMO rouding PR is quite important since we currently fail the spec tests when compiled for 32bit.

PR #1631, opened 2015-10-26
Fix regression with evalution binary operatons with strings

### LINKED ISSUE #1629: Regression in outputting property list values
```scss
foo {
  background: url(...) 2rem 3rem / auto 2rem;
}
```
LibSass 3.2.5
```scss
foo {
  background: url(...) 2rem 3rem/auto 2rem; }
```
LibSass 3.3.0
```scss
foo {
  background: url(...) 2rem 3rem/"auto" 2rem; }
```
Spec added https://github.com/sass/sass-spec/pull/558
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27693281-regression-in-outputting-property-list-values?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xzyfer:
Ruby Sass does does not evaluate binary operations between numbers and string. Instead is simply outputs the expression.

Fixes #1629 
Spec sass/sass-spec#560

### DIFF
--- a/src/eval.cpp
+++ b/src/eval.cpp
@@ -1325,6 +1325,12 @@ namespace Sass {
     if (ltype == Expression::NULL_VAL) error("Invalid null operation: \"null plus "+quote(unquote(rstr), '"')+"\".", lhs.pstate());
     if (rtype == Expression::NULL_VAL) error("Invalid null operation: \""+quote(unquote(lstr), '"')+" plus null\".", rhs.pstate());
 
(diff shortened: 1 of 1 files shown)


xzyfer:
I don't think we're ever going to get this 100% right until we move resolving of operations to be the AST nodes themselves.

PR #1636, opened 2015-10-26
add "File input path" to formatted error json

cfebs:
Adds `File input path: path/to/sass/input` line to `formatted` json error key.

As a consumer of grunt sass (and therefore libsass), it would be very valuable to know the input file that was being processed at the time of error.

Here is an example of where it would be useful in a large code base.

#### Example

_error.scss_
```scss
@import './lib.scss';

.test {
    color: $map;
}
```
_lib.scss_
```scss
$map: (
    1: '#000'
);
```

This is an error, but there is the output given currently:

```
Error: (1: "#000") isn't a valid CSS value.
        on line 2 of lib.scss
>>     1: '#000'
   ----^
```

If `lib.scss` is very commonly imported in a large codebase it can be cumbersome to track down which file is actually the cause of error.

With this change the output would be:

```
Error: (1: "#000") isn't a valid CSS value.
        on line 2 of lib.scss
>>     1: '#000'
   ----^
File input path: error.scss
```

Which points you (at least) to the file being processed at time of error - hopefully saving some headaches for people trying to track down their errors.

#### Notes

Only added in the file context, processing from `stdin` via `sassc` does not apply this line.

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -228,6 +228,10 @@ extern "C" {
         msg_stream << "   " << std::string(e.pstate.column - move_in, '-') << "^\n";
       }
 
+      std::string msg_input_file_prefix("Input path: ");
+      std::string rel_input_path(Sass::File::resolve_relative_path(c_ctx->input_path, cwd, cwd));
+      msg_stream << msg_input_file_prefix << rel_input_path << "\n";
+
       JsonNode* json_err = json_mkobject();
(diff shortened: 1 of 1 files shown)


xzyfer:
We're consistent with Ruby Sass. We don't intend to diverge on this. Please open a feature request with Ruby Sass https://github.com/sass/sass/issues

Issue #1640, opened 2015-10-27
Regression in `@content` inside `@else`


xzyfer:
Originally reported [via Twitter](https://twitter.com/iainspad/status/658795627032289280)

```scss
@mixin foo() {
    @if false {
      a { b: c }
    } @else {
      @content;
    }
}

@include foo() {
  .foo {
    bar: baz;
  }
}
```
LibSass 3.2.5
```css
.foo {
  bar: baz; }
```
LibSass 3.3.1
```
Error: Mixin "foo" does not accept a content block.

       Backtrace:
       	test.scss:65
        on line 65 of test.scss
>> @include foo() {
   ---------^
```

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/569

PR #1641, opened 2015-10-27
Fix regression with `has_content` detection

mgreter:
Fixes https://github.com/sass/libsass/issues/1640


xzyfer:
Spec activated https://github.com/sass/sass-spec/pull/570

Issue #1650, opened 2015-10-27
:nth-of-type(2n - 1) space in argument results in error


kkopachev:
Updated to new libsass/node-sass recently and noticed regression (or intended?) with the code:
```
tr:nth-of-type(2n - 1) { color: red; }
```
outputs an error
> "Invalid CSS after "": expected 1 selector or at-rule, was "th-of-type(4n -...""

While code with no spaces works:
```
tr:nth-of-type(2n-1) { color: red; } // works!
tr:nth-of-type(2n + 1) { color: red; } // works too!
tr:nth-of-type(2n+1) { color: red; } // and this one is fine
```

spacing around minus sign makes libsass complain.
> $ node-sass -v
node-sass	3.4.1	(Wrapper)	[JavaScript]
libsass  	3.3.1	(Sass Compiler)	[C/C++]

mgreter:
OK, this is pretty odd.

```
:nth-of-type(2n-1),
:nth-of-type(2n-  1),
:nth-of-type(2n  -1),
:nth-of-type(2n  -  1),
:nth-of-type( 2n  -  1 )
{ color: red; }
:nth-of-type(2n+1),
:nth-of-type(2n+  1),
:nth-of-type(2n  +1),
:nth-of-type(2n  +  1),
:nth-of-type( 2n  +  1 )
{ color: red; }
```

In compressed mode this produces:
```
:nth-of-type(2n-1),
:nth-of-type(2n-  1),
:nth-of-type(2n  -1),
:nth-of-type(2n  -  1),
:nth-of-type(2n  -  1)
{color:red}

:nth-of-type(2n+1),
:nth-of-type(2n+1),
:nth-of-type(2n+1),
:nth-of-type(2n+1),
:nth-of-type(2n+1)
{color:red}  
```

No idea why ruby sass does not apply the same logic here. Looks like a bug to me ...

xzyfer:
The white space being preserved looks like the work of a static value parser.

xzyfer:
Looks like Ruby Sass' `almost_any_value_token` method is the cause.

chriseppstein:
That's a ruby bug, not a high priority one, but definitely a bug.

mgreter:
@chriseppstein thanks for the confirmation, I guess it's a one byte fix on the ruby side ...

mgreter:
This should be fixed on master for release 3.3.3.

PR #1652, opened 2015-10-27
Improve selector lookahead for "negative" binomials

mgreter:
Fixes https://github.com/sass/libsass/issues/1650
Spec: https://github.com/sass/sass-spec/pull/574


xzyfer:
Looks like in compressed more Ruby Sass does some minor selector rewriting which accounts for the difference noted in https://github.com/sass/libsass/issues/1650#issuecomment-151645209.

https://github.com/sass/sass/blob/stable/lib/sass/tree/visitors/to_css.rb#L300

We should be able to emulate this behaviour.

mgreter:
https://github.com/sass/sass/issues/1874

Issue #1651, opened 2015-10-27
@extend inside @at-root (without: media) doesn't work


mgol:
``` scss
.a {
  display: block;
}

.b {
  @at-root (without: media) {
    @extend .a;
  }
}
```

Ruby Sass

``` css
.a, .b {
  display: block; }
```

LibSass

``` json
Error: Extend directives may only be used within rules.
        on line 51 of test.scss
>>     @extend .a;
   ----^
```

Spec added sass/sass-spec#620
Depends on #1661

mgreter:
Looks like we never really supported `without` syntax correctly. I personaly see it the first time ...

xzyfer:
This depends on https://github.com/sass/libsass/issues/1661

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/620

xzyfer:
Heads up @mgol we've reverted the patch for this issue because it was simply a placebo. We currently do not support with/without with at-root.

xzyfer:
Scratch that, we do support with/without. As previously state this issue is due to the parser attempting to assert things it cannot know. The nesting checking for `@extends` must be done in eval or later.

xzyfer:
Specs deactivated in https://github.com/sass/sass-spec/pull/741

mgol:
This still has the `3.3.3` milestone and libsass 3.3.4 has already been released. Why is that?

xzyfer:
There was a patch slated for 3.3.3 but it wasn't sufficient. Changed the milestone to 3.4.

mgreter:
Related to https://github.com/sass/libsass/issues/2444 and https://github.com/sass/libsass/issues/1890

PR #1655, opened 2015-10-28
Fix parsing of At_Root_Blocks with expressions

mgreter:
Fixes https://github.com/sass/libsass/issues/1651
Spec: https://github.com/sass/sass-spec/pull/575

Does not pass the error spec. Couldn't come up with anything better, but it should solve the reported regression. We could need some more spec tests for this ...

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -2121,7 +2121,11 @@ namespace Sass {
       expr = parse_at_root_expression();
     }
     if (peek < exactly<'{'> >()) {
-      body = parse_block(true);
+      std::string feat("");
+      if (expr && expr->feature())
+        feat = expr->feature()->to_string();
+      // allow declarations in certain cases
+      body = parse_block(feat != "without");
     }
     else if ((lookahead_result = lookahead_for_selector(position)).found) {
       Ruleset* r = parse_ruleset(lookahead_result, false);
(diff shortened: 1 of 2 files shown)


xzyfer:
Ok so this is more complicated than it seems.

We do *fully* support `without` and there is good test coverage for it. The at-root logic happens in the cssize step. This is how Ruby Sass does it.

We should not throw the `@extend` error in the parser. We should throw during the extend process. Only then can we know for sure that the `@extend` is valid.

src/parser.cpp
```diff
@@ -2121,7 +2121,11 @@ namespace Sass {
       expr = parse_at_root_expression();
     }
     if (peek < exactly<'{'> >()) {
-      body = parse_block(true);
+      std::string feat("");
+      if (expr && expr->feature())
+        feat = expr->feature()->to_string();
+      // allow declarations in certain cases
+      body = parse_block(feat != "without");
```
xzyfer: Unfortunately the `without` directive is too complex for this. For example should error.

```
a { b: c }

.d {
  @supports (foo: bar) {
    @media all {
      @at-root (without: media) {
        @extend a;
      }
    }
  }
}
```

mgol:
This still has the `3.3.3` milestone and libsass `3.3.4` seems to already have been released. Why is that?

xzyfer:
This patch slated for 3.3.3 but it isn't sufficient. Closing as this approach doesn't address the underlying issue.

PR #1665, opened 2015-10-28
Clean up some internal method names

xzyfer:
Extracted from #1626

Clean up some internal method names for accuracy in preparation for a refactor `context.cpp`.


saper:
Just a remark: I think render is a bit too generic for a good name, one could imagine having two steps: one to compile the code into some intermediate form and another one "executing", "performing" or "running" it.

Issue #1667, opened 2015-10-28
map-get not returning correct value


camflan:
I'm still trying to track this bug down, but I'd like to start some discussion this. The official Ruby Sass compiler does not have this bug, it returns the correct results. 

Things I through/debugged before finally trying sass and seeing that it worked as expected:
- Somehow I was modifying the map during execution and that's why it was returning the value for get-size(3). Nope, that's not it.
- Somehow get-size was getting passed only the original value without my scaling factor. Nope, that's not it.

I have a set of functions used to return a size from a scale map. I've left my debugging output here so the console output below makes sense.

```
$type-scale: (
	-15:0.066667rem,
	-10:0.186rem,
	-9:0.211rem,
	-8:0.26rem,
	-7:0.295rem,
	-6:0.364rem,
	-5:0.413rem,
	-4:0.51rem,
	-3:0.578rem,
	-2:0.714rem,
	-1:0.809rem,
	0:1rem,
	1:1.133rem,
	2:1.4rem,
	3:1.586rem,
	4:1.96rem,
	5:2.221rem,
	6:2.744rem,
	7:3.109rem,
	8:3.842rem,
	9:4.353rem,
	10:5.378rem,
	11:6.094rem,
	12:7.53rem,
	13:8.531rem,
	14:10.541rem,
	15:11.943rem,
	16:14.758rem
);

@function get-size($size) {
	@if map-has-key($type-scale, $size) {
		@debug $type-scale;
		@debug $size + " => " + map-get($type-scale, $size);
		@return map-get($type-scale, $size);
	}

	@warn "Not a valid size.";
	@return null;
}

@function scale-size($rem-size, $steps) {
	$size-key: get-key-for-value($type-scale, $rem-size);
	
	@if $size-key {	
		@debug "HERE";
		$new-size: $size-key + $steps;
		@return get-size($new-size);
	}
	
	@warn "Not able to find size for " + $rem-size;
	@return null;
}

@function get-key-for-value($map, $value) {
	@each $map-key, $map-value in $map {
		@if $map-value == $value {
			@return $map-key
		}
	}
	
	@warn $value + " not found in " + $map;
	@return null;
}
```

SASSC
----------
And here is the output from sassc, the pair of output is from get-size(7);
The second output with "HERE" is when I call through scale-size(3, 4);

```
[...]

PR #1676, opened 2015-11-01
Make sure hash is recalculated after value update

mgreter:
Fixes https://github.com/sass/libsass/issues/1667 and also saw and fixed another bug:

```scss
$map {
  1: 1,
  1px: 1px
}
```

Resulted in duplicate key for map error.

Sass-Spec PR: https://github.com/sass/sass-spec/pull/594


xzyfer:
:+1:

Issue #1561, opened 2015-09-20
Error message mismatch for duplicate map key


saper:
Difference between Ruby Sass and libsass from https://github.com/sass/sass-spec/pull/373

```
$a: 'foo';
$b: 'foo';
$map: (
  $a: 1,
  $b: 2
);
```
```
-"Duplicate key "foo" in map ($a: 1, $b: 2)."
+"Duplicate key "foo" in map ("foo": 2)."
```
Specs added https://github.com/sass/sass-spec/pull/373/files

PR #1678, opened 2015-11-01
Improve error message for duplicate keys

mgreter:
Fixes https://github.com/sass/libsass/issues/1561


xzyfer:
Is there a test to assert this behaviour doesn't regress?

saper:
I think I have pushed somewhere into "maps" tests but It is not there...

xzyfer:
It exists but it hasn't been activated.

PR #1681, opened 2015-11-02
More consistent deprecation warnings

xzyfer:
The deprecation warning format in Ruby Sass is inconsistent which makes it difficult to
match. These are the two common cases. This PR addresses the two most common cases.

Spec https://github.com/sass/sass-spec/pull/581

### DIFF
--- a/src/error_handling.cpp
+++ b/src/error_handling.cpp
@@ -22,7 +22,7 @@ namespace Sass {
     warn(msg, pstate);
   }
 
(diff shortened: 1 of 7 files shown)


xzyfer:
The algorithm I have for determining which path to show isn't perfect, but it's closer than it was and it makes sass-spec happy. I've opened https://github.com/sass/sass/issues/1880 to help get us on the same page moving forward.

PR #1687, opened 2015-11-03
[WIP] Update to ruby sass 3.4.19 specs

mgreter:
Last commit should pass the spec tests, but needs better implementation and test coverage for two or three specfic cases (like delayed evaluation of binary operations could need a refactor).

Updated Spec Tests: https://github.com/sass/sass-spec/pull/589


xzyfer:
@mgreter I'm pretty much done with this work. Leave it to me. We won't bump to 3.4.19 in this release since that conflicts with the static parser refactor I'm working on. We'll aim for 3.4.16 so we can get the sass colour rounding in place.

PR #1688, opened 2015-11-03
Implement `Sass::round` function

xzyfer:
Cherry picked from https://github.com/sass/libsass/pull/1621

This PR implements `Sass::round` which emulates Ruby Sass' elipson rounding. This allows us to bump sass-spec to 3.4.15. One step closer to 3.4.19 which will come after my static value parsing refactor is complete.

Spec https://github.com/sass/sass-spec/pull/590
Closes #1621

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1864,17 +1864,17 @@ namespace Sass {
     // resolved color
     std::string res_name = name;
 
-    double r = round(cap_channel<0xff>(r_));
-    double g = round(cap_channel<0xff>(g_));
-    double b = round(cap_channel<0xff>(b_));
+    double r = Sass::round(cap_channel<0xff>(r_));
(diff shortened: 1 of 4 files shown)


xzyfer:
Interestingly this is failing on OSX builds.

xzyfer:
@mgreter AFAIK `Sass::Util::round` doesn't care about precession. This is for the output stage. I've removed it from this PR. Feel free to add it back if there is a specific need for it. I'm not aware of one.

mgreter:
@xzyfer IMHO `0.00001` means that it has an epsilon for precision of 5 and should be different for other precision options. IMHO this is a ruby sass bug ...

xzyfer:
That's not our decision to make. Ruby Sass bugs are LibSass features.

mgreter:
https://github.com/sass/sass/issues/1883

PR #1691, opened 2015-11-03
Use an access token to increase the GH API limit

xzyfer:
The proxy approach is unreliable, and I'm wasting too much time with this. If this hack works I'm shipping it.

```
Travis rate limit on github exceeded
Retrying via 'special purpose proxy'
curl: (7) Failed connect to libsass.ocbnet.ch:80; Connection refused
```

The key needs to obfuscated like this because github as a bot that scans for committed keys and revokes them for security. This is a read only key so it's safe to be out in the public. Alternatively there are TravsiCI's [encrypted env vars](http://docs.travis-ci.com/user/encryption-keys/) but they don't work for [PRs from forks](http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests) which makes them useless to us.

### DIFF
--- a/script/ci-build-libsass
+++ b/script/ci-build-libsass
@@ -91,14 +91,8 @@ then
 
   echo "Fetching PR $TRAVIS_PULL_REQUEST"
 
-  JSON=$(curl -L -sS https://api.github.com/repos/sass/libsass/pulls/$TRAVIS_PULL_REQUEST)
 
-  if [[ $JSON =~ "API rate limit exceeded" ]];
-  then
-    echo "Travis rate limit on github exceeded"
-    echo "Retrying via 'special purpose proxy'"
-    JSON=$(curl -L -sS http://libsass.ocbnet.ch/libsass-spec-pr.psgi/$TRAVIS_PULL_REQUEST)
-  fi
+  JSON=$(curl -L -sS https://api.github.com/repos/sass/libsass/pulls/$TRAVIS_PULL_REQUEST)
 
   RE_SPEC_PR="sass\/sass-spec(#|\/pull\/)([0-9]+)"
 


mgreter:
@xzyfer you are aware that this token gives write access to the repo!?

xzyfer:
Does it? It's meant to be read only.

mgreter:
https://github.com/afragen/github-updater/issues/109

xzyfer:
I don't follow. He claims it give write access, with no evidence even though GH allows for read-only tokens. Are you able to use this to write to the repo?

xzyfer:
You appear to be correct. *rage*

>Grants read/write access to code, commit statuses, collaborators, and deployment statuses for public repositories and organizations. Also required for starring public repositories.

mgreter:
I didn't try, and I don't know how you set it up exactly. But I was under the impression they are pretty much all read/write ... https://help.github.com/articles/creating-an-access-token-for-command-line-use/: ```Remember to keep your tokens secret; treat them just like passwords! They act on your behalf when interacting with the API. Don't hardcode them into your programs. Instead, opt to use them as environment variables.```

xzyfer:
I am significantly frustrated by having been blocked in CI issue for the last couple hours. We should go back to just running the sass-spec PR code on the minimal set of combinations.

mgreter:
I'm restarting my server, so the fallback proxy should work in a few ticks ... unfortunately embedded libsass process brings it down from time to time, trying to track the reason for that for quite some time.

xzyfer:
If we're going to stay down the proxy approach we should find a 3rd party proxy service. I have no issues giving someone money to make this problem go away. We need something myself or @saper can also manage without needing to get a hold of you. Timezones and all that.

mgreter:
Feel free to set something up! My proxy is a perl psgi application and I can share that code if you need. For now it is at least another possibility before failing anyway.

QuLogic:
So if I understand this correctly, it's trying to read the PR text and see if there's a sass/sass-spec reference to pull? How about moving that reference into the git commit message? Then you can just check locally and not hit the API at all.

xzyfer:
@QuLogic this becomes problematic when you have multiple commits in a PR. It's also practically impossible to get contributors to follow any kind of guidelines when it comes to git.

QuLogic:
I'm not so sure; just pick the latest one: `git log --grep 'sass/sass-spec' master..origin/pulls/$TRAVIS_PULL_REQUEST | head -1` (Sorry, I forgot the exact branch name Travis uses of the top of my head.)

If contributors don't mention that they updated (or plan to) sass-spec in their message, then just have the build fail. Actually, I think there's a CI that checks that...

PR #1703, opened 2015-11-05
Testing coveralls

xzyfer:


mgreter:
Works again as expected IMHO: https://coveralls.io/jobs/9510294. Guess you did fix it!

xzyfer:
Ish.. It does create build but doesn't push to GH which is annoying me.

PR #1704, opened 2015-11-06
Implement rounding precision

mgreter:
Should pass https://github.com/sass/sass-spec/pull/696

Update rounding behavior for upcoming [ruby sass 3.4.20] [1].

My implementation should return the same results but avoids the fmod division operation.
We may also want to calculate the epsilon only once at startup and store it in options.

[1]: https://github.com/sass/sass/issues/1883

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -2116,17 +2116,17 @@ namespace Sass {
     // resolved color
     std::string res_name = name;
 
-    double r = Sass::round(cap_channel<0xff>(r_));
-    double g = Sass::round(cap_channel<0xff>(g_));
-    double b = Sass::round(cap_channel<0xff>(b_));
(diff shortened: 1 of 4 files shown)


xzyfer:
Let's leave this until sass spec is 3.4.20/ready. I assume there's a test
case for this new behaviour?

mgreter:
As confirmed by ruby sass dev team, this was a bug. Not sure why we should withold a bugfix that doesn't affect our spec tests! We have a lot of other edges that are not 100% compatible with ruby sass output. It does not break existing spec tests, so for me this is ready to be merged!

saper:
Is precision a user-defined option in ruby sass as well?

mgreter:
Yes it is.

PR #1713, opened 2015-11-10
Fix regressions in context refactor

### LINKED ISSUE #1711: Regression in `sass_context_get_included_files_size`
This bug is a high priority and will back ported to 3.3.2.

Prior to 3.3.2 `sass_context_get_included_files_size` absolute paths, in 3.3.2 it now returns relative paths.

This regression occurred in 610acbfba44764180cf8677ef6e6752c37e8d5a8

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28138897-regression-in-sass_context_get_included_files_size?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1712: Regression with custom importers
This bug is a high priority and will back ported to 3.3.2.

A custom importer should be able to import the same path twice. Below is a common pattern in node-sass custom importers.

```
sass.render({
  data: "@import foo; @import bar;',
  importer: function(url, prev, done) {
    done({
      file: '/some/other/path.scss',
      contents: 'div {color: yellow;}'
    });
  }
});
```

The path here serves no real purpose for the import itself but is useful for paths in `@debug`, `@error` and backtraces. 

In 3.3.2 a regression was introduced that causes an error to be throw.

```
duplicate resource registered
```

This regression occurred in 610acbfba44764180cf8677ef6e6752c37e8d5a8

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28139507-regression-with-custom-importers?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1714: Regression with line comments show absolute paths
Line comments should be absolute. This was incorrectly changed in 610acbfba44764180cf8677ef6e6752c37e8d5a8 to show relative paths.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28140256-regression-with-line-comments-show-absolute-paths?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xzyfer:
This PR fixes two major regressions from the context refactor.

#### Don't throw an error when importing the same path

> Importing the same path is completely valid in Sass. We don't error when `@import`ing the same path, this should not when using custom importers.

Fixes #1712 

#### The included_file vector should be absolute paths not relative

>This behaviour was incorrectly changed in 610acbf which caused some serious regressions.

Fixes #1711 

#### Line comments should be absolute

>This was incorrectly changed in 610acbf to show relative paths.

Fixes #1714

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -276,7 +276,7 @@ namespace Sass {
     resources.push_back(res);
 
     // add a relative link to the working directory
-    included_files.push_back(abs2rel(inc.abs_path, CWD));
+    included_files.push_back(inc.abs_path);
     // add a relative link  to the source map output file
     srcmap_links.push_back(abs2rel(inc.abs_path, source_map_file, CWD));
 
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -259,11 +259,11 @@ namespace Sass {
 
     // do not parse same resource twice
     // maybe raise an error in this case
-    if (sheets.count(inc.abs_path)) {
-      free(res.contents); free(res.srcmap);
-      throw std::runtime_error("duplicate resource registered");
-      return;
-    }
+    // if (sheets.count(inc.abs_path)) {
+    //   free(res.contents); free(res.srcmap);
+    //   throw std::runtime_error("duplicate resource registered");
+    //   return;
+    // }
 
     // get index for this resource
     size_t idx = resources.size();
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -135,8 +135,7 @@ namespace Sass {
       if (ctx && ctx->c_options->source_comments) {
         std::stringstream ss;
         append_indentation();
-        std::string path = Sass::File::abs2rel(r->pstate().path, ctx->cwd());
-        ss << "/* line " << r->pstate().line + 1 << ", " << path << " */";
+        ss << "/* line " << r->pstate().line + 1 << ", " << r->pstate().path << " */";
         append_string(ss.str());
         append_optional_linefeed();
       }


src/context.cpp
```diff
@@ -259,11 +259,11 @@ namespace Sass {
 
     // do not parse same resource twice
     // maybe raise an error in this case
-    if (sheets.count(inc.abs_path)) {
-      free(res.contents); free(res.srcmap);
-      throw std::runtime_error("duplicate resource registered");
-      return;
-    }
+    // if (sheets.count(inc.abs_path)) {
```
xzyfer: @mgreter I'm not sure if this has other consequences. Please confirm that this is safe to do.

xzyfer: Reading through the code I don't see an obvious issue with this.

drewwells:
Data compiler looks good, Current dir messages are gone

Before
```
        Error > stdin:1
        file to import not found or unreadable: branch
        Current dir:
        @import "branch";
        div.branch {
          @extend %branch;
          div.leaf {
            @extend %leaf;
          }
        }
```

After
```
Error > stdin:1
        File to import not found or unreadable: branch
        Parent style sheet: stdin
        @import "branch";
        div.branch {
          @extend %branch;
          div.leaf {
            @extend %leaf;
          }
        }

```

drewwells:
File compiler looks good too.

src/context.cpp
mgreter: It might lead to memory leaks and it's also not really defined what should happen since importer could return different content for the same url this way.

xzyfer: > It might lead to memory leaks

This was my concern. It wasn't clear from the code that it was the case though.

> since importer could return different content for the same url this way

This is feature people rely on in custom importers so we need to support it in some fashion.

PR #1721, opened 2015-11-11
Fixes for Windows version

Taritsyn:


xzyfer:
Please provide more information.

xzyfer:
Thoughts @saper @am11?

Taritsyn:
`functions.cpp` file contains fix to the following error:

	Error	7	error C2061: syntax error : identifier 'uint_fast32_t'	C:\tmp\LibSassHost-master\LibSassHost-master\src\libsass\functions.cpp	1179	1	libsass-ia32
	
This error occurs when you compile in Visual Studio 2013.

xzyfer:
Thanks for the clarification. This is a known issue. Follow the discussion in https://github.com/sass/libsass/pull/1719

src/functions.cpp
```diff
@@ -6,6 +6,9 @@
 #include "constants.hpp"
 #include "to_string.hpp"
 #include "inspect.hpp"
+#ifdef _WIN32
```
am11: `#if defined(_MSC_VER) && _MSC_VER < 1900` (because it compiles with MinGW and VS2015+ on Windows), and I will move this block before or after `#ifdef __MINGW32__` block (see few lines below in this file).

am11:
One nit, otherwise LGTM. Thanks for updating the targets file! :)

Taritsyn:
@am11 Thanks for tip. I made changes to `functions.cpp` file.

xzyfer:
Thanks @Taritsyn, you rock! Thanks @am11 for your input.

PR #1736, opened 2015-11-14
Defer creation of hash maps into eval stage

### LINKED ISSUE #1169: Quoted and unquoted colors in maps produces 'Duplicate key "transparent" in map'

mgreter:
This finally fixes #1169, which was caused by a severe bug with how we handle maps!
Additional Spec-Tests: https://github.com/sass/sass-spec/pull/623

The current behavior seems rather random and I didn't care to check when exactly it wouldn't work as expected. The bug is rather obvious, as can be seen with the [following example] [1]:
```
$map: (
  random(): foo,
  random(): bar
);
```

This will currently result in a duplicate key error. The parser actually already creates a map for the given sass code above, which will never work correctly, since we only know the actual keys after the evaluation phase (which is not available to the parser). It feels rather natural to use an even-sized list to carry it over to the evaluation phase, and create the actual map there.

A lot of languages allow the creation of maps from even-sized lists, so it felt natural to extend the existing List class to support this. But we only use this internally (from an external point, we only have space and comma separated lists). Unsure if sass can convert hash maps to lists and vice versa?

The error reporting is once again rather funky. It will always report the hex representation for duplicate color keys. This is the first time I see this happen and our codebase didn't support that in anyway without adding a new function just to get that representation (it's not `to_string` nor `inspect`).

ToDo:

- [x] Implement `to_string` for missing expression types
- [x] Re-use new `to_string` functions in CRTP visitors

[1]: http://libsass.ocbnet.ch/srcmap/#JG1hcDogKAogIHJhbmRvbSgpOiBmb28sCiAgcmFuZG9tKCk6IGJhcgopOw==

### DIFF
--- a/include/sass/values.h
+++ b/include/sass/values.h
@@ -29,7 +29,8 @@ enum Sass_Tag {
 // Tags for denoting Sass list separators
 enum Sass_Separator {
   SASS_COMMA,
-  SASS_SPACE
+  SASS_SPACE,
+  SASS_HASH
 };
 
 // Value Operators
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1880,17 +1880,112 @@ namespace Sass {
     if (empty()) return res;
     if (is_invisible()) return res;
     bool items_output = false;
-    std::string sep = separator() == SASS_COMMA ? "," : " ";
+    std::string sep = separator() == SASS_SPACE ? " " : ",";
     if (!compressed && sep == ",") sep += " ";
     for (size_t i = 0, L = size(); i < L; ++i) {
+      if (separator_ == SASS_HASH)
+      { sep[0] = i % 2 ? ':' : ','; }
       Expression* item = (*this)[i];
       if (item->is_invisible()) continue;
       if (items_output) res += sep;
-      if (Value* v_val = dynamic_cast<Value*>(item))
-      { res += v_val->to_string(compressed, precision); }
+      if (Expression* ex = dynamic_cast<Expression*>(item))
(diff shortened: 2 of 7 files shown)


mgreter:
I've implemented the missing details in https://github.com/sass/libsass/pull/1754. The commit in here already fixes the most urgent bug, so I leave this one as is. With the interpolation refactoring and the one I've linked we're getting IMHO pretty close to a more sane and easier to use internal API.

Issue #1746, opened 2015-11-20
Variables first declared in control flow rules should not be accessible in outer scopes


andrew-skybound:
Variables assigned in block scopes should not be accessible outside of the scope, unless they have been previously assigned in an outer scope (or the outer scope is global and the `!global` flag is specified).

Libsass gets these semantics correct for style, `@at-root`, `@media`, and other CSS @-rules, as well as mixins but not for control flow rules (`@if`, `@while`, `@for`, and `@each`).  Notice in the tests below that the `_local` variables being first declared within these rule types are leaking outside.

It is also notable that assignments that happen in non-control-flow rules do not overwrite variables that have been previously assigned in an outer scope.  These assignments are denoted by variables named `_global` in the tests.  Libsass gets this behavior correct, but I have included tests anyway to help prevent a regression.

Input:
```CSS
$continue: true;

$while_global: initial;
@while $continue {
	$while_local: assigned;
	$while_global: assigned;
	$continue: false;
}

$each_global: initial;
@each $b in 1 {
	$each_local: assigned;
	$each_global: assigned;
}

$for_global: initial;
@for $b from 1 to 2 {
	$for_local: assigned;
	$for_global: assigned;
}

$if_global: initial;
@if true {
	$if_local: assigned;
	$if_global: assigned;
}

$else_global: initial;
@if false {
	/* nothing */
}
@else {
	$else_local: assigned;
	$else_global: assigned;
}

$at_root_global: initial;
@at-root {
	$at_root_local: assigned;
	$at_root_global: assigned;
}

$stylerule_global: initial;
stylerule {
	$stylerule_local: assigned;
	$stylerule_global: assigned;
}

$mixin_global: initial;
@mixin set_variable() {
	$mixin_local: assigned;
	$mixin_global: assigned;
}

@include set_variable();

locals {
	while: variable-exists(while_local);
	each: variable-exists(each_local);
	for: variable-exists(for_local);
	if: variable-exists(if_local);
	else: variable-exists(else_local);
	at_root: variable-exists(at_root_local);
	stylerule: variable-exists(stylerule_local);
	mixin: variable-exists(mixin_local);
}

globals {
	while: $while_global;
	each: $each_global;
	for: $for_global;
	if: $if_global;
	else: $else_global;
	at_root: $at_root_global;
	stylerule: $stylerule_global;
	mixin: $mixin_global;
}
```

Ruby SASS Output:
```CSS
locals {
  while: false;
  each: false;
  for: false;
[...]

PR #1748, opened 2015-11-21
Implement variable shadow scoping

mgreter:
Fixes https://github.com/sass/libsass/issues/1746 (spec tests: https://github.com/sass/sass-spec/pull/625)
Shadow scopes overwrite variables in global scopes if
they already exist there. Normal scopes do not allow
this without setting the `!global` flag on assignment.

### DIFF
--- a/src/environment.cpp
+++ b/src/environment.cpp
@@ -4,11 +4,20 @@
 namespace Sass {
 
   template <typename T>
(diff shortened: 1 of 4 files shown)


xzyfer:
Spec added sass/sass-spec#625

Issue #1747, opened 2015-11-21
nested styles refer to top-level parent in sourcemap line number


tehOPEologist:
### i'm creating this issue here to get more visibility on an issue i originally opened in the node-sass repo.  link here: https://github.com/sass/node-sass/issues/1206

i'm having an issue where my nested styles' sourcemap line number in chrome is referencing the top-most parent selector line and not the line that the nested selector starts.

```
body {
  h1 {
    background: #f00;
  }
}
```

will show file.scss:1 instead of file.scss:2 for the h1

i've tried outputting the map file to a separate file and embedding it, all the same.

***
***

just ran a test with the latest ruby sass and it's sourcemapping nested children on the child's line and not the top-most parent like it should be.  this is definitely a bug and isn't working as intended.

### Ruby Sass 3.4.19:
```scss
body {
  h1 {
    background: red;
  }
}
```
#### compiles to:
```css
body h1 {
  background: red; }

/*# sourceMappingURL=core.css.map */
```
#### core.css.map:
```json
{
"version": 3,
"mappings": "AACE,OAAG;EACD,UAAU,EAAE,GAAG",
"sources": ["core.scss"],
"names": [],
"file": "core.css"
}
```
---
### node-sass 3.4.1:
```scss
body {
  h1 {
    background: red;
  }
}
```
#### compiles to:
```css
body h1 {
  background: red; }

/*# sourceMappingURL=core.css.map */
```
#### core.css.map:
```json
{
	"version": 3,
	"file": "core.css",
	"sources": [
		"core.scss"
	],
	"mappings": "AAAA,IAAI,CACF,EAAE,CAAC;EACD,UAAU,EAAE,GAAI,GACjB",
	"names": []
}
```

nek4life:
Having the same issue with python-libsass, it seems it's the underlying libsass implementation of sourcemaps

mgreter:
I do see your point. But unfortunately libsass technically does the correct thing and in some way chrome also reports the correct thing. It actually shows where the first selector in the sequence is declared. And this is basically not what one would probably expect, specially given that chrome doesn't use much of the information that would be available via the source-map. BTW. with less you also get that line reported in chrome.

[...]

PR #1759, opened 2015-11-25
Add crutch to help browsers interpreting sourcemaps

mgreter:
They should report where the scope gets opened and not
the position of the first simple selector in the sequence.

Browsers actually report the mapping found most right just before the first selector in the sequence. I first tried to simply wrap each entry in a selector list with a mapping from their last complex selector, but as it turned out browsers wouldn't pick that up, so I had to delay the emmitting of that specific source map entry just before the token for the selector is emitted.

Should fix https://github.com/sass/libsass/issues/1747

### DIFF
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -13,6 +13,7 @@ namespace Sass {
     scheduled_space(0),
     scheduled_linefeed(0),
     scheduled_delimiter(false),
+    scheduled_mapping(0),
     in_comment(false),
     in_wrapped(false),
     in_media_block(false),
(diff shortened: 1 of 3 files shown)


xzyfer:
:+1:

x-yuri:
@mgreter I understand the kludge is still in effect? It may be that it's not needed anymore. I didn't test it thoroughly, but from my couple of attempts it seems like at least Chromium 63 looks for the next mapping after [the opening curly bracket][1].

And there's a [related issue][2], where a guy from Firefox team wants to change libsass behavior. But it may be an opportunity to change how Firefox handles things.

[1]: https://github.com/webpack-contrib/css-loader/issues/381#issuecomment-360687110
[2]: https://github.com/sass/node-sass/issues/2092

PR #1767, opened 2015-11-30
removing duplicated content from c api example

rodneyrehm:


mgreter:
Thx @rodneyrehm!

xzyfer:
Thanks!

PR #1774, opened 2015-12-05
win: Converts wchar_t* to UTF-8 std::string

am11:
`_getcwd(_, __)` returns an ANSI string on Windows. The wide-character
variant is [`_wgetcwd(_, __)`](https://msdn.microsoft.com/en-us/library/sf98bd4y).

The result is then converted using C++11's `std::wstring_convert` to
`std::string`.

Fixes sass/node-sass#716.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2015
+﻿os: Visual Studio 2015
 
 environment:
   CTEST_OUTPUT_ON_FAILURE: 1
@@ -27,8 +27,7 @@ install:
       if ($env:Compiler -eq "mingw" -AND -Not (Test-Path "C:\mingw64")) {
         # Install MinGW.
         $file = "x86_64-4.9.2-release-win32-seh-rt_v4-rev3.7z"
-        $url  = "https://bintray.com/artifact/download/drewwells/generic/"
-        $url += $file
+        $url  = "https://bintray.com/artifact/download/drewwells/generic/$file"
         Invoke-WebRequest -UserAgent wget -Uri $url -OutFile $file
         &7z x -oC:\ $file > $null
       }
@@ -53,7 +52,7 @@ build_script:
 
 test_script:
   - ps: |
-      $PRNR = [System.Environment]::ExpandEnvironmentVariables("%APPVEYOR_PULL_REQUEST_NUMBER%")
+      $PRNR = $env:APPVEYOR_PULL_REQUEST_NUMBER
       if ($PRNR -ne "") {
         echo "Fetching info for PR $PRNR"
         $request = (New-Object System.Net.WebClient)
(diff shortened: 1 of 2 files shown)


am11:
@mgreter, unfortunately [`<codecvt>`](http://www.cplusplus.com/reference/locale/codecvt/) header is [not supported by GCC](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.200x), thus getting it right with VS and Clang are the only options on Windows with this code. I have kept the old (broken) behavior for MinGW. In future, we can add fallback for `std::wstring` to `std::string` loss-less conversion to correct this behavior for MinGW.

am11:
Added the fallback for mingw, now it is fixed for both compilers.

xzyfer:
CI is happy and I trust your judgment @am11. Thanks again!

xzyfer:
@am11 is it possible create a test case for this?
I'm guessing you'll need to import a file in a directory with cyrilic characters, maybe?

am11:
Thanks @xzyfer, sure I will prepare a test case for spec repo tonight. =)

am11:
@xzyfer, to test this scenario, once https://github.com/sass/sass-spec/pull/640 is merged, we can add the following lines after https://github.com/sass/libsass/blob/73d5143/appveyor.yml#L70:

```yml
      cd sass-spec/spec/libsass/Sáss-UŢF8/
      $env:TargetPath ./input.scss
```
(scenario is: when cwd path has Cyrillic characters)


Without this, https://github.com/sass/sass-spec/pull/640 will only be testing https://github.com/sass/sassc/pull/156 (that scenario is: when `argv[]` / input path has Cyrillic characters).

am11:
As for the coverage, sine the code paths are mutually exclusive (`#ifdef` out for mingw/msvc), this change has no effect as it will run every time with or without `appveyor.yml` change.

xzyfer:
@am11 can you please update the appveyor.yml as you've suggested?

xzyfer:
Merged sass/sass-spec#640 and sass/sassc#156

am11:
@xzyfer, done! :shipit: :dart:

xzyfer:
:ok_hand: perfect! Thank you

saper:
@am11 is setting `$env:TargetPath` done with intention to override the setting from the `sassc.vcxproj` file? If yes, then it seems that the linker complains with 

````warning MSB8012: TargetPath(sassc/bin/sassc) does not match the Linker's OutputFile property value (C:\projects\libsass\sassc\bin
\sassc.exe). 
This may cause your project to build incorrectly. 
To correct this, please make sure that $(OutDir), $(TargetName) and $(TargetExt) property values match the value specified in %(Link.OutputFile)
````

Shall we modify `$env:OutDir` as well?

am11:
I think `TargetPath` was just a random name which happens to collide with some MSBuild reserved variable name.
Since Sassc's vcxproj has the `OutDir` set correctly to the desired path: https://github.com/sass/sassc/blob/363090a/win/sassc.vcxproj#L58-L64, perhaps we can just rename `TargetPath` to `SasscPath` throughout `libsass/appveyor.yml` to avoid this warning.

PR #1778, opened 2015-12-08
add base64-encode sass function to libsass

jonasjonas:
I don't know if you would merge this, because the base64-encode is not an original sass function and has to be imported through ruby when using normal sass.

But in most of my workflows this is one of the essential functions missing from libsass when e.g. generating svg background or icons from other sass variables, that are used in layouts.

Fortunately libsass already contains a base64 encoder that that this merge would be minimal.


xzyfer:
Thanks for your initiative. Although we cannot accept this pull request we
greatly appreciate your contribution.

As you state this is not a native Sass function. As such it's not something
we cannot have in LibSass.

I suggestion opening a feature request issue on the Ruby Sass issue tracker
for this function. IMHO it could be useful function and a Sass
implementation would be slow.

mgreter:
@jonasjonas you could create a native plugin for this. Please check https://github.com/mgreter/libsass-math to get some ideas how this can be done. You probably need to extract and copy the base64 base parts though.

jonasjonas:
Thanks for the feedback!
I'll look into it.

PR #1780, opened 2015-12-09
Fixup some docs on importers (undefined variables)

asottile:
Noticed a tiny error in the docs while doing my research on importers for dahlia/libsass-python and figured I'd fix it :)


xzyfer:
Thanks @asottile!

Issue #1433, opened 2015-08-10
Makefile `mkdir` needs `-p` flag


toastal:
I tried to build libsass this morning on Debian and `mkdir` was bawking on creating directory because it's exists. If it contained the `-p` flag will prevent that error.

xzyfer:
This has come up before. The concern has been that `-p` is not portable. See discussion in https://github.com/sass/libsass/pull/795

We're open to workarounds.

saper:
We can  use `-$(MKDIR)/lib`

xzyfer:
@toastal can you be more specific on where it failed? I was not able to produce this on a fresh checkout and `make` of libsass master.

saper:
> @toastal can you be more specific on where it failed? I was not able
> to produce this on a fresh checkout and `make` of libsass master.

It might be related to the fact that previous release didn't clean "lib" folder.
So it might happen only once after upgrade or something like that.

toastal:
The `-p` flag is for after files are created so of course it didn't fire on a fresh checkout. The steps I'd done were `git pull` that was behind a month or two. When I ran `make` (without `make clean` so that I'm not having to rebuild what's not needed), make bawked on creating one of the folders because it existed and so nothing was built. I didn't realize that this is an unfortunate incompatibility with Windows's `mkdir`. I manually added `-p` to the Makefile on line 5 and it has worked ever since.

mgreter:
IMO this was fixed by adding the directory as a target, so it gets created on demand without needing the `-p` flag, which is how this should be handled. Please re-open if you can reproduce with latest master.

PR #1782, opened 2015-12-10
Fix libsass build for SmartOS

jaredmorrow:
* Use ginstall rather than install for INSTALL variable
* unset `SEC` because it was being used elsewhere
* Use namespacing where there were complaints

Addresses: #1308 

Things I still needed to do to get npm to install node-sass

in libsass:

```
make
make install
make install-headers
```

For npm:

```
export npm_config_libsass_ext=auto
npm install node-sass
```

I don't know where this should be documented, but I thought I'd throw it in there for anyone else who might happen to hit this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,14 @@
+OS       ?= $(shell uname -s)
 CC       ?= gcc
 CXX      ?= g++
 RM       ?= rm -f
 CP       ?= cp -a
-MKDIR    ?= mkdir
+MKDIR    ?= mkdir -p
 RMDIR    ?= rmdir
 WINDRES  ?= windres
+ifeq ($(OS),SunOS)  # Solaris/Illumos flavors
+INSTALL  = ginstall
+endif
 INSTALL  ?= install
 CFLAGS   ?= -Wall
 CXXFLAGS ?= -Wall
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,12 @@ CC       ?= gcc
 CXX      ?= g++
 RM       ?= rm -f
 CP       ?= cp -a
-MKDIR    ?= mkdir -p
+MKDIR    ?= mkdir
 RMDIR    ?= rmdir
 WINDRES  ?= windres
 ifeq ($(OS),SunOS)  # Solaris/Illumos flavors
 INSTALL  = ginstall
+PREFIX   = /opt/local
 endif
 INSTALL  ?= install
 CFLAGS   ?= -Wall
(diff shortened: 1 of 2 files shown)


jaredmorrow:
Thinking about this a bit more, the proper way to install it would be to set PREFIX to be `/opt/local`, and then I wouldn't have to use the CFLAGS, but then the PREFIX setting in the makefile would have to be changed to be `?=` instead of `=` and that might start down the slippery slope of multi-platform support littering the makefile.

Makefile
```diff
@@ -1,10 +1,14 @@
+OS       ?= $(shell uname -s)
 CC       ?= gcc
 CXX      ?= g++
 RM       ?= rm -f
 CP       ?= cp -a
-MKDIR    ?= mkdir
+MKDIR    ?= mkdir -p
```
xzyfer: I believe there have previously been discussion regard the portability of `-p`.

https://github.com/sass/libsass/issues/1433

xzyfer:
@QuLogic and @saper are probably best suited to make the call on this.

jaredmorrow:
Ah, okay.  Well we can fix it by specifying a proper PREFIX for SmartOS then.

jaredmorrow:
I'll stay subscribed to this PR, so just holler and I'm willing to make any changes and re-push.

Makefile
Vagabond: I would remove this, I added it by accident when tinkering.

jaredmorrow:
Fixed in latest commit, and properly set PREFIX, which was the original need for the `-p`.  Lastly, I removed a conditional that wasn't needed when you can just use the `?=` to get the same logic.

xzyfer:
Great thanks. Give the scope of the changes I'm happy to merge this when CI is happy.

QuLogic:
I guess the Makefile stuff depends on how much OS-specific stuff you want; as you say you can use `make INSTALL=ginstall PREFIX=/opt/local`.

clord:
Ah crap, I went and basically wrote this exact same PR :) patiently waiting for a pull.

saper:
Two questions:

1) Which "install" is now default on Solaris? Or more precisely, which part breaks when using traditional SysV "install"?

2) I think we should just rename all constants and prefix them with something like `LIBSASS_` to avoid issues like the `SEC` one. Otherwise it'll be the endless chase...

xzyfer:
> I think we should just rename all constants and prefix them with something like LIBSASS_ to avoid issues like the SEC one. Otherwise it'll be the endless chase...

Makes sense to me. Should be its own PR though.

jaredmorrow:
Should I remove the Makefile changes so this PR can merge?  I'm fine either
way, I can easily document setting the proper variables on the command
line.

xzyfer:
@jaredmorrow can you answer @saper's question?

>Which "install" is now default on Solaris? Or more precisely, which part breaks when using traditional SysV "install"?

We want to get Makefile compatible with SmartOS. If manual intervention is required then we're not really supporting SmartOS.

xzyfer:
I'm ok with merging this the community can start compiling LibSass from master.

I've created #1802 to track prefixing constants.

PR #1789, opened 2015-12-13
Be sure to install interface.h

SamWhited:
Currently `interface.h` is not being installed by `make install` and is therefore missing from some Linux distros (Arch, in my case).


SamWhited:
Sorry for the noise, didn't notice this had been deprecated. I'm assuming leaving it out was deliberate.

PR #1810, opened 2015-12-28
repo: Add exit code for explicit test

am11:
Also fixes regression form #1774:

In #1774, I switched `ExpandEnvironmentVariables` to `$env`, which
breaks the subsequent empty string checks.
`$env:APPVEYOR_PULL_REQUEST_NUMBER`, returns `NULL`, when commit is not
yet associated with PR.

This PR adds null or empty check along with
minor improvement for `Get-Content` (instead of .NET IO).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ libsass/*
 *.a
 *.suo
 *.sdf
+*.opendb
 *.opensdf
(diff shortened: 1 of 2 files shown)


xzyfer:
I'll take your word for it :)

PR #1817, opened 2015-12-30
util: Adds CRLF awareness to quote()

am11:
Input:

```scss
@import url("a
b");
```

Output on Windows:

```
"@import url(\"a\r\a b\");"
```

Output on Unix (alinged with Ruby Sass, hence the expected output):

```
"@import url(\"a\a b\");"
```

See: https://github.com/sass/libsass/issues/1096#issuecomment-167531680
for details.

Sepc: https://github.com/sass/sass-spec/pull/664.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -488,6 +488,12 @@ namespace Sass {
 
       int cp = utf8::next(it, end);
 
+      // in case of \r, check if the next in sequence
+      // is \n and then advance the iterator.
+      if (cp == '\r' && it < end && utf8::peek_next(it, end) == '\n') {
+        cp = utf8::next(it, end);
+      }
+
       if (cp == '\n') {
         quoted.push_back('\\');
         quoted.push_back('a');


am11:
//cc @xzyfer =)

src/util.cpp
```diff
@@ -488,6 +488,12 @@ namespace Sass {
 
       int cp = utf8::next(it, end);
 
+      // in case of \r, check if the next in sequence
+      // is \n and then advance the iterator.
+      if (cp == '\r' && it < end && utf8::peek_next(it, end) == '\n') {
```
xzyfer: Looking through the rest of this file it looks like we generally treat `\r` as a noop. I think you can drop the peek here.

xzyfer:
Nice find @am11

xzyfer:
I wonder how we can test this since the spec for #1096 includes this test case and passes on Windows.

src/util.cpp
am11: Great! Fixed by fb1fcd7.

am11:
Good point. Since we don't have `.gitattributes` in sass-spec (which usually enforces the repo-wide linefeed rules), we can probably save https://github.com/sass/sass-spec/blob/65d2f5f/spec/libsass-closed-issues/issue_1096/input.scss in notepad (with CRLF ending), then run it once without this commit to make the error visible on AppVeyor.

xzyfer:
Sounds unreliable. Especially give that we periodically regenerate all the specs (usually on osx or linux).

src/util.cpp
am11: @xzyfer, seems like in some cases, we do need to `push_back()` the carriage return to quoted string: https://travis-ci.org/sass/libsass/jobs/99452071#L1108.

am11: On second though, `\r` is a legal character in Unix and Windows. Now I wonder if Ruby Sass is doing the right thing in first place. :bulb:

xzyfer: Interesting.

am11:
@xzyfer, if you commit from Linux and OSX, it will not affect CRLF endings. I have just committed https://github.com/sass/sass-spec/pull/664 from OSX.

src/util.cpp
am11: If CSS3 import only accepts network paths, then `\r` is legal string of length 2. :)
I have reverted it back to this original code. Now it aligns with Ruby both on Windows and OSX. CI seems to be happy too. :palm_tree:

PR #1821, opened 2016-01-01
API: Provides a function to clone string

am11:
From the pure API usage (as in FFI), there is a limitation in case of
`sass_make_data_context`, where API requires the caller to allocate the
memory and transfer the ownership to libsass, so libsass will
destroy it afterwards. In case of FFI, despite the provision of
interoperability, the memory allocated by language runtime may not be
deallocatable by CRT, and it corrupts the heap on `free()`ing the
resources.

This utility method equips the caller to get the cloned string ptr back
from CRT and use it to make data context.

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -375,6 +375,16 @@ extern "C" {
     return options;
   }
 
+  char* ADDCALL sass_copy_c_string (const char* input_string)
+  {
+    return sass_strdup(input_string);
+  }
+
+  void ADDCALL sass_free_c_string (char* input_string)
+  {
+    if (input_string)  free(input_string);
+  }
+
   Sass_File_Context* ADDCALL sass_make_file_context(const char* input_path)
   {
     struct Sass_File_Context* ctx = (struct Sass_File_Context*) calloc(1, sizeof(struct Sass_File_Context));
--- a/src/sass_context.hpp
+++ b/src/sass_context.hpp
@@ -54,7 +54,7 @@ struct Sass_Options : Sass_Output_Options {
   // Directly inserted in source maps
   char* source_map_root;
 
-  // Custom functions that can be called from sccs code
+  // Custom functions that can be called from Sass code
   Sass_Function_List c_functions;
 
   // List of custom importers
--- a/docs/api-context-internal.md
+++ b/docs/api-context-internal.md
@@ -78,7 +78,7 @@ struct Sass_Options {
   // Directly inserted in source maps
   char* source_map_root;
 
-  // Custom functions that can be called from sccs code
+  // Custom functions that can be called from Sass code
   Sass_C_Function_List c_functions;
 
   // Callback to overload imports
--- a/docs/api-context.md
+++ b/docs/api-context.md
@@ -88,7 +88,7 @@ char* source_map_file;
 char* source_map_root;
 ```
 ```C
-// Custom functions that can be called from sccs code
+// Custom functions that can be called from Sass code
 Sass_C_Function_List c_functions;
 ```
 ```C
@@ -154,6 +154,13 @@ struct Sass_Data_Context; // : Sass_Context
 
 // Create and initialize an option struct
 struct Sass_Options* sass_make_options (void);
+
+// Create a copy of C string
+const char* sass_copy_c_string (const char* input_string);
+
+// Delete a copied C string
+void sass_free_c_string (char* input_string);
+
 // Create and initialize a specific context
 struct Sass_File_Context* sass_make_file_context (const char* input_path);
 struct Sass_Data_Context* sass_make_data_context (char* source_string);
@@ -168,12 +175,12 @@ struct Sass_Compiler* sass_make_data_compiler (struct Sass_Data_Context* data_ct
 
 // Execute the different compilation steps individually
 // Usefull if you only want to query the included files
-int sass_compiler_parse(struct Sass_Compiler* compiler);
-int sass_compiler_execute(struct Sass_Compiler* compiler);
+int sass_compiler_parse (struct Sass_Compiler* compiler);
+int sass_compiler_execute (struct Sass_Compiler* compiler);
 
 // Release all memory allocated with the compiler
 // This does _not_ include any contexts or options
-void sass_delete_compiler(struct Sass_Compiler* compiler);
+void sass_delete_compiler (struct Sass_Compiler* compiler);
 
 // Release all memory allocated and also ourself
 void sass_delete_file_context (struct Sass_File_Context* ctx);
--- a/include/sass/context.h
+++ b/include/sass/context.h
@@ -30,6 +30,13 @@ enum Sass_Compiler_State {
 
 // Create and initialize an option struct
 ADDAPI struct Sass_Options* ADDCALL sass_make_options (void);
+
+// Create a copy of C string
+ADDAPI char* ADDCALL sass_copy_c_string (const char* input_string);
+
+// Delete a copied C string
+ADDAPI void ADDCALL sass_free_c_string (char* input_string);
+
 // Create and initialize a specific context
 ADDAPI struct Sass_File_Context* ADDCALL sass_make_file_context (const char* input_path);
 ADDAPI struct Sass_Data_Context* ADDCALL sass_make_data_context (char* source_string);


am11:
@xzyfer, needed to expose `sass_strdup` to API surface for P/Invoke stuff, where string allocated in dotnet is passed to libsass works fine until libsass tries to `free()` it. I have encountered this issue with `source_string` in case of `Sass_Data_Context`. 

Made it a general purpose API function for making C strings, considering it might be reused when implementing the custom function, custom importers, (experimental) custom headers etc. In C# code now I have three cases covered to deal with strings (this PR helps in Case 2):

```c#
if (string.IsNullOrEmpty(sassOptions.Data))
{
    Sass.InternalContextPtr = sass_make_file_context(sassOptions.InputFile);
    // Case 1: dotnet string passed as is to CRT via interop services.
    // this works just fine because InputFile was allocated by dotnet and will be 
    // deallocated by dotnet GC. libsass will duplicate it and deallocate the clone on
    // freeing the context.
    // libsass uses IMPLEMENT_SASS_OPTION_STRING_ACCESSOR macro for
    // setting input_file option.
}
else
{
    Sass.InternalContextPtr = sass_make_data_context(sass_make_c_string(sassOptions.Data));
    // Case 2: we don't need to store the ptr returned by sass_make_c_string locally, as libsass
    // will take care of deallocation.
}

Sass.InternalOptionsPtr = sass_context_get_options(Sass.InternalContextPtr);
Sass.InternalIndentPtr = Marshal.StringToCoTaskMemAnsi(sassOptions.Indent);
sass_option_set_indent(sassOptionsInternalPtr, Sass.InternalIndentPtr);
// Case 3: dotnet should take care of allocation and deallocation, but
// since libsass is using IMPLEMENT_SASS_OPTION_ACCESSOR macro for setting this option
// (as opposed to IMPLEMENT_SASS_OPTION_STRING_ACCESSOR, why?), which
// does not duplicate the string, we have to reallocate memory on process heap and then call
// `Marshal.FreeCoTaskMem(Sass.InternalIndentPtr);` in C# class dtor (IDisposable's
// Dispose()).
```

*(based on the comment about Case 3, should we switch usages of `IMPLEMENT_SASS_OPTION_ACESSOR` macro with `IMPLEMENT_SASS_OPTION_STRING_ACCESSOR` for all stringly options?)*

The non-generic alternative of this change is to add an overload of `sass_make_data_context`, which accepts `Sass_Value `(as we have maker for `Sass_Value` already on API surface area).

src/sass_context.hpp
```diff
@@ -153,4 +153,4 @@ struct Sass_Compiler {
   Sass::Block* root;
 };
 
-#endif
\ No newline at end of file
+#endif
```
drewwells: :+1:

mgreter:
I guess it doesn't really hurt, but I would prefer `sass_copy_c_string`. Makes it a bit more standing out and IMO even describes the purpose a bit better. And I guess I will have to badger you to also update the documentation for the C-API :wink: 

Again thanks for your contributions!

mgreter:
LGTM! Thanks a lot!

am11:
@mgreter, changed the function name, I agree that `sass_copy_c_string` sounds more appropriate. =)

src/sass_context.cpp
```diff
@@ -375,6 +375,11 @@ extern "C" {
     return options;
   }
 
+  const char* ADDCALL sass_copy_c_string (const char* input_string)
```
mgreter: Not 100% sure, but I feel `char*` without const would be more appropriate here, since you are indeed responsible to delete (and therefore change) the memory. Thinking about it, we may should also provide a `sass_free_c_string` function then?

src/sass_context.cpp
```diff
@@ -112,7 +112,7 @@ extern "C" {
       c_ctx->error_file = sass_strdup(e.pstate.path);
       c_ctx->error_line = e.pstate.line+1;
       c_ctx->error_column = e.pstate.column+1;
-      c_ctx->error_src = e.pstate.src;
+      c_ctx->error_src = sass_strdup(e.pstate.src);
```
mgreter: Are you sure about this? Who is freeing it?

src/sass_context.cpp
mgreter: Just checked and it should be `char*` since ie `sass_context_take_error_message` also return a `char*` without const, and the have the same purpose. Seeing this a `sass_free_c_string` actually should be there to free these.

mgreter:
Added a few comment, sorry to nitpick further. But I cannot give my blessing for the change in `sass_context` for `sass_strdup(e.pstate.src)`. And we should also have a `sass_free_c_string`.

Hope you got some time to also add these? Should be copy paste mostly while ammending this commit! You may also want to elaborate why you added the string copy in context?

src/sass_context.cpp
am11: @mgreter, libsass is actually taking ownership of this string and will free it. This is only for data context's data string only, which could be a large string. In node-sass binding, we are passing the memory to libsass, which libsass free's as part of `sass_delete_data_context`.

mgreter: Not sure if I follow? If you request a copy it's your reponsability to free it!?

am11: The string passed to sass_copy_c_string will be deallocated by the called. The cloned string by strdup is returned to caller as a string, which will further passed as char* to sass_make_data_context. On deleting the context, libsass free this string. So everything is freed in the end.

src/sass_context.cpp
am11: Added a free call in 9874428. I think no other down-stream is using it. When we get the syntax error and call `sass_context_get_error_src` after the compilation, it returns garbage, because libsass deletes cpp_ctx right after compilation. Caught this when debugging.

mgreter: Ah I see. Did you check if it's the only property affected by this?

mgreter: I don't feel too confortable to copy the whole source if avoidable. We should see if we can delay the de-allocation then. I do see the point that it's propably only affecting error cases. But I would like to check first if we can delay it.

mgreter:
Might be wort to split into two commits. Would love to see one adding copy and free functions, while the second fixes the "premature" de-allocation issue.

am11:
@mgreter, `sass_copy_c_string` is only for `sass_make_data_context`. I think `sass_copy_c_string` name is the confusion, which begs the counterpart `sass_free_c_string`.
Should I instead add a `bool clone` parameter to `sass_make_data_context` and strdup the source_string if clone is `true`? Note that in this case libsass is taking the ownership of this string and freeing it on deleting the context.

src/sass_context.cpp
am11: Sure meanwhile I am reverting it back.

> Did you check if it's the only property affected by this?

I am make the SassError object on libsass.net which captures everything that libsass API has to offer. I made code errors in Sass and hit couple of different exceptions, but this seems to be the only one which was returning garbage.

mgreter:
The point of `sass_copy_c_string` is to copy a c string, but you don't know for sure how to free it. To have a 100% sane API we need the counterpart, so we can implement and change whatever we use in the copy function. This also is true for every "take" function, since those internaly also come from the same copy call. That's why I also want a free counterpart. Hope that made more sense!

am11:
I have added the `sass_free_c_string` to complete the pair. 

Agree that reference counting will handle these situation more robustly library-wide. IMO, at the moment, `source_string` is the only odd case where libsass is mixing the concerns. IMO, the usual two cases are:

a) Caller allocates the memory, caller passes to lib, lib uses it caller deallocates it after some function (like `compile context`) returns.

b) Caller allocates the memory, lib copies it (strdup), caller deallocates the memory, concerns are separated, so lib de-allocates the cloned copy without caring about consumer.

However, in case of `sass_make_data_context`, caller allocates the memory and pass it to `sass_make_data_context` as `source_string`, libsass does not clone it but assign the ptr as is and later deallocates it after the compilation. So if the caller is a managed runtime (like .NET), the caller will have to somehow tell the garbage collector (usually running as a separate thread or service) to shut its eye on that chunk of memory (i.e. memory pinning), which is bit of extra hassle to manage.

Given this situation, should we make `sass_make_data_context`'s behaviour fit into `a` or `b` described above? =)

mgreter:
Will need to think about it a bit more. But from a gut standpoint (and AFAIR) we currently try to do b) in all cases with one big exception. We want you to overturn the source string for data context to libsass. The reason is that this makes sense in pretty much all use cases, because you normaly are done with that memory after you have passed it to libsass. But I also have wondered if we want to allow consumers to tell us if we should free the string or not. Problem is that caller now has the obligation to keep the memory alive as long as libsass needs to use it, so passing it seemed the best choice (and was IMO consistent with the old sass_interface). So that option now would only serve a purpose when the caller actually wants to hold on to the source string, which I doubt has any use cases at all (well I guess you have given one when we pass that pointer back to the caller for the error object).

am11:
As per the discussion, i have made the alternative version for this where caller can tell libsass to clone the source string memory via a boolean flag, and therefore would not be obliged to keep the memory alive: https://github.com/am11/libsass/commit/f076061efaa0a130f8d3d1ae5e1f07b5f3a50a66.

mgreter:
I don't really like that option, since it is pretty redundant to specify an option to have libsass do something,, as you know exactly what to do. I will need to think about the big picture more deeply.

It also came to my mind that you probably destroy the context object before accessing that memory, since libsass should keep the memory alive until you call `sass_delete_context`. So from that standpoint we do have a defined life-cycle. But I agree that is not very obvious. We also lack a `take` function for the error object in regard to the file source. Need to check these, as I cannot tell with certainty.

mgreter:
I just replied to @saper on slack regarding https://github.com/saper/sass-crash-test and since my answer is closely related to this, I will post the same here for future reference. 

> @saper: Checked out your samples in that sass-crash-test repo and compiled and tried them all and AFAICS all that fail (002 works ok for me) are a result of "wrong usage" and I will try to explain why, but will need some paragraphes to do so fully. I fail to see how our API should be broken if used correctly.
001 and 003 fail because ie char data[60] is on the stack which will be automatically deallocated once the stack variable goes out of scope. Libsass must hold on to that memory, since we might need it to stay around for error reportings. And we cannot delegate that duty to the implementer. There are IMO only two ways to solve this. Always make a copy when the string is first passed to libsass, to make both sides independent of each other, which will mean we have two copies of the full source around in memory. The other possibility is to take over ownership of the memory. And that only works if the memory is on the heap (we need to have control over when that memory is deallocated). So by that API contract you have to pass a heap variable to libsass, since you cannot delegate/pass ownership/life-cycle of a stack variable.
Main reason I choose the second approach, and AFAIR it the same way as the previously and now obsolete interface worked, is as following: There is simply no real use case where you would have the source input in a stack variable (beside made up cases). To have the source variable on the stack you need to know how much space you need for it on compile time. If you know that I'm pretty sure you also know what sass code you will have. In other words having it on the stack pretty much translates to knowing it is static (at compile time). If you know the source you pass on compile time, you would also know the expected result on compile time, so why even bother to pass it to libsass? Normal use
case is implementer loads the source and really just wants to pass it to libsass, and the API was mainly built for exactly that use case. You must not deallocate the source memory on your own, once you passed it to libsass, ever, that is one contract the libsass API has.
Having the source input on the stack implicitly violates that contract.
has a leak in strdup("file.css"). Only the input source life-cycle is passed to libsass. Libsass takes a copy of all other strings you pass (as with the filename). In that case the strdup is superflous and leads to the memory leak. I actually haven't tested it, but from reading the code I'm very confident that is the problem there.

```
struct Sass_File_Context* sass_make_file_context (const char* input_path);
struct Sass_Data_Context* sass_make_data_context (char* source_string);
```

> As you can see one expects a const char* while the other expects char*. And finally 005 also seems to work:

```
Importing(#0): lala
Importing(#1): delta
/* import */
/* import */
/* First */
a {
  color: red; }
```

So in the end I seem to come to the conclusion that we are already doing the right thing, but need to communicate the contracts more clearly. We cannot either depend on implementer to keep the source input memory around for us, since he can't know how long we will need to have it around. And once we pass it back to implementer via ie. error object, implementer cannot trust us to keep that memory alive as long as he needs it (so he should make a copy in that case). I guess the contract that we keep that memory alive until implementer calls `sass_destroy_context` is a fair and well defined border. We may could offer a similar function like I did with the other "take" functions. So implementer could request owenership of memory via reported errors.

am11:
:+1: 

In order to make memory lifecycles independent:

* All the memory passed to libsass should be copied. 
* After compilation, when the consumer is done extracting out the results, consumer must call `sass_delete_context` which would delete all the constituents of context (and those of intermediary context aka: cpp_context); until that time, libsass will not free the memory.

mgreter:
@am11 I don't see a compelling reason or use case why we would need to copy source input memory. I really don't see any use case. And it would be a breaking change for all implementors.

am11:
@mgreter, sure, alright. I will figure out some other way to handle my use case.

drewwells:
I like the part about all memory is kept until sass_delete_context. Does it
not work that way currently?

saper:
@mgreter I am sure you will disagree with me, but here it is:

I wrote those examples on purpose to see how the API breaks. We are programming C here and not Pascal and C pointer semantics does not tell you *anything* about where does the memory come from.

One reason why our API is broken is that you cannot pass a static or stack-allocated memory. Suppose I write a library that wraps libsass and I also cannot guarantee where the memory comes from.

Second reason is sometimes you need to use a custom allocator, for example because you operate in the environment like V8 or Python virtual machine. I am unable to pass strings created on those platforms directly because they will not survive libsass operations. Copying strings is taking time and is really bad solution. 

Also by not using `const char *` one eliminates lots of possibilities to optimize the code by the compiler, because the compiler never knows if there isn't a pointer that is going to modify that memory in unexpected ways. 

I do not understand the point where you are afraid that the stack memory can be deallocated in-between. Except for callbacks (imports, functions) I don't see the issue here.

The change in the API to "own" all the memory was done to work around some double-free problems in the library instead of fixing them. Yes, I know memory management is difficult and C++ does not make it easier.

Typically C APIs require the caller to return a large output buffer to be used as the output. If it's too small, one truncates the result or returns the error code. A custom memory allocator in libsass could be used to store C++ objects in the caller-supplied buffer as well (that would be actually a good use for the custom allocator for `new`). 

libsass did use caller-provided memory not so long ago at least for its data, so it's not impossible.

am11:
@drewwells, there are two topics in this thread:

* the error_source is holding garbage after compilation (prematurely disposing). @mgreter said that will be fixed separately by re-ensuing the lifecycle of every exportable object. To my knowledge, rest of the objects do not have this problem.

* this PR is mainly about providing a way for external process to allocate char* memory on libsass process heap, which will help in `make_data_context`'s input string argument; when an external runtime calls the lib via FFI. To my knowledge none of the down streams at this point are consuming libsass directly via FFI, but interfacing it via a C++ wrapper. For example, we can try Ruby-FFI or lua-ffi and get into the same situation (at least on Windows) where I am at with .NET P/Invoke. Note that with libsass-net, we already have the C++/CLS wrapper which is again a C++ wrapper around libsass, not calling libsass via FFI. The hip way is to get it working via P/Invoke, essentially for CoreCLR and Mono targets support (for Linux, OSX and BSD support). With that being said, I have few other options (in case this PR does get rejected :blush:):
  * On Windows, P/Invoke kernel32's `VirtualAllocEx` function (or file-memory mapping, or named pipes) to allocate string on libsass heap for libsass, so libsass can destroy the memory without getting AccessVoilation and likes. On Unix, if the parent/child processes have the same restrictions pertaining to memory access as Windows (one allocates memory on its own heap and pass to the other, and when other tries to free it, the process horks), p/invoke suitable libc functions to inject heap memory for each operating system.
  * Keep this patch in fork and point submodule accordingly. This will keep the libsass-net code clean and portable.
  * (are there other options? please feel free to enlighten me please..)

xoofx:
Sorry to bump into this PR, as I required almost something very similar to @am11 is looking in the PR #1973 

Though, I would like to add a simple sass_string_alloc function instead of a sass_string_dup (or copy from c), as I would like to avoid duplicating the allocation in C# (as I can dump directly a wchar string to utf8 string, instead of dumping to my own allocated utf8 string, and then copy/reallocate the same utf8 string in libsass)

xoofx:
So @mgreter Would you like me to make a PR with a new `sass_allocate_memory` and `sass_free_memory` as you suggested [here](https://github.com/sass/libsass/pull/1973#issuecomment-205729714)?

mgreter:
I created a proposal in https://github.com/sass/libsass/pull/1983. Waiting for feedback!
Thanks for the contributions and feedbacks here (@am11 et al)!

Issue #1750, opened 2015-11-23
Issues when importing a file that imports other files not relative to the first file


jkrehm:
I use grunt-sass 1.1.0 that uses node-sass 3.4.2 that uses libsass 3.3.2. Since updating node-sass from 3.4.1 to 3.4.2 (which bumped the libsass version) our builds don't work and I've tracked the problem down to libsass 3.3.2, but my C++ is so rusty that I'm not sure how to fix the issue and submit a pull request.

The scenario: our stylesheet (style.scss) imports bootstrap-sass's [_bootstrap.scss](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/_bootstrap.scss). This in turn imports its own files, relative to itself. Those relative files are not found unless I add Bootstrap's directories to the include-path array. I think libsass is looking relative to the initial file, but not relative to the imported files. [Font Awesome](https://github.com/FortAwesome/Font-Awesome/blob/master/scss/font-awesome.scss) has the same issue.

I would really rather not include vendor paths in our include-path array, especially since it wasn't required before. Perhaps it's something that was introduced with 610acbf? (e78b314 references that commit as having another relative related issue) Again, my C++ is practically non-existent at this point.

JohnAlbin:
So… your Sass file has something like this:

```scss
// Assuming this file is in project-root/sass.
@import "../bower_components/bootstrap-sass/assets/stylesheets/bootstrap";
```

?

jkrehm:
Yep. That's not the exact code, but it's close enough to get the idea.

xzyfer:
I'm unable to reproduce this locally.

```
|-- test.scss
|-- foo/
  |-- test.scss
  |-- bar/
    |-- test.scss
```
```
/* test.scss */
@import "foo/test";
```
```
/* foo/test.scss */
@import "bar/test";
```
```
/* foo/bar/test.scss */
```
~~~~~~
Ruby Sass
```
/* test.scss */
/* foo/test.scss */
/* foo/bar/test.scss */
```
LibSass 3.3.2
```
/* test.scss */
/* foo/test.scss */
/* foo/bar/test.scss */
```

jkrehm:
I think I narrowed down the issue to when the include path is something like this: `../www-root`.

File layout:
![layout](https://cloud.githubusercontent.com/assets/999845/11508060/12f006b8-9814-11e5-9798-39ddfe258d3c.png)

```
/* www-root/styles/style.scss */
[...]

PR #1826, opened 2016-01-06
Fix `join_paths` when left path is not canonical

mgreter:
We blindly count `/./` as an actual directory, thus the
results were wrong when the right side starts with `../`.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -203,8 +203,10 @@ namespace Sass {
       if (l[l.length()-1] != '/') l += '/';
 
       while ((r.length() > 3) && ((r.substr(0, 3) == "../") || (r.substr(0, 3)) == "..\\")) {
-        r = r.substr(3);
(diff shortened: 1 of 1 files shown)


xzyfer:
Is this related to #1750?

If not do you think you can look into it? Feels like it might be a regression in the recent context refactor.

mgreter:
Hmm, could be, but no real way to test it with libsass. Saw it from a perl-libsass integration test ...

mgreter:
BTW. bug in `join_paths` was introduced with https://github.com/sass/libsass/commit/5f9bc0528785c0dca106ff3e10d1c932e74c3ffe, but might be that we triggered it with some recent changes.

xzyfer:
Interesting. Ship when ready.

mgreter:
Waiting for :traffic_light:

Issue #1822, opened 2016-01-02
Ampersand glued to the selector should be rejected


saper:
This is from https://github.com/sass/node-sass/issues/1324:

````
.btn {
    .open& {
        color: #000;
    }
}
````

as of https://github.com/sass/libsass/tree/13ae62def291c4934ad04d0d527e84b4021d60a3 produces:

````
.open& {
  color: #000; }
````

and this is rejected by Ruby Sass.

(I hope this is not a duplicate, the closest match I could find was https://github.com/sass/libsass/issues/1491)

**Comment by saper** [comment]
This should be the proper behaviour as of Sass 4.0 per https://github.com/sass/sass/issues/1425

**Comment by xzyfer** [comment]
We'll leave this for the time being to see what happens with this in Ruby Sass.

**Comment by arkonan** [comment]
If sass/sass#1425 is trying to match LESS's behavior I would expect 
```
.btn {
    .open& {
        color: #000;
    }
}
```
to produce
```
.open.btn {
  color: #000;
}
```
(rather than output the `&` into the CSS)

xzyfer:
sass/sass#1425 is the correct place for that discussion.

PR #1828, opened 2016-01-06
Error on invalid parent selector placements

mgreter:
Fixes https://github.com/sass/libsass/issues/1822


xzyfer:
More generally `&` changes the selector ident lexer so that it's an alternate to the first character.

PR #1835, opened 2016-01-07
Include <cstdint> unconditionally

saper:
A follow up to #1721

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -14,6 +14,7 @@
 #include "sass/base.h"
 #include "utf8.h"
 
+#include <cstdint>
 #include <cstdlib>
 #include <cmath>
 #include <cctype>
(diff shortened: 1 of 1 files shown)


saper:
Not sure what PR ready means ... but hey, green looks so much better!

mgreter:
It's mostly just usefull for issues, but I tend to flag PRs that I feel can be "blindly" merged if needed. Just a minor nitpick, a comment in the source for such conditionals would be usefull. Other than that :+1:

xzyfer:
I'm all for this. We've talked bout VS2013 before. The only reason we haven't addressed it yet is that we don't have a plan for testing it in parallel with VS2015.

https://github.com/sass/libsass/pull/1719#issue-116363736

saper:
This change removes VS version conditional - we should always include this header because we use width int types. I tested it with VS2013 and VS2015 was tested by appveyor.

We should be able to run tests for VS2015 as an additional environment if we want.

xzyfer:
Rebuilding. If CI is still happy :ship:

PR #1836, opened 2016-01-07
Use Visual Studio 2013 for node/Windows XP compatibility

saper:
* Use Visual Studio 2013 to ensure the same
  platform as the node engines are currently built

* Include <cstdint> unconditionally to prevent
  missing type declarations. (follow up to https://github.com/sass/libsass/pull/1721)

Fixes: https://github.com/sass/node-sass/issues/1283

More information why we need this change https://github.com/sass/node-sass/issues/1283#issuecomment-169845062

This pull request is against libsass release 3.3.2 (not master)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2015
+os: Visual Studio 2013
 
 environment:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -14,6 +14,7 @@
 #include "sass/base.h"
 #include "utf8.h"
 
+#include <cstdint>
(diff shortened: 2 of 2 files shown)


saper:
Unfortunately the tests fail since the use the newest sass-spec, not the one we've had with 3.3.2 release.

mgreter:
Is this still needed with latest master?
I checked with MSVC 2013 community!

saper:
The first commit here is the same as #1835, so master is now compiling on VC2013.
The second commit is to enable building again on VC2013 on AppVeyor, so we can check if node-sass still compiles.

saper:
(This PR was against 3.3, not master).

To avoid confusion, submitted https://github.com/sass/libsass/pull/1888 to supersede this.

PR #1841, opened 2016-01-08
Refactor/milestore 3.4

mgreter:
Should pass https://github.com/sass/sass-spec/pull/669

@xzyfer I'm pushing this here, since I had to fix some stuff in parser for the binary expression handling and could no longer wait. I consider the code base of this branch stable, but its need some more cleaning up. I'm also pretty sure that a few flags and ifs are no longer needed with this refactor.

Beside the fixes that are open in the other PRs, this mainly improves and fixes how we parse values and value schemas. I also added a pretty extensive test suite for this. It passes all expected tests against 3.4.21, minus the open sass bugs, plus the new extended test suite.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -2260,9 +2260,10 @@ namespace Sass {
     std::string str("");
     auto end = this->end();
     auto start = this->begin();
+    std::string sep(compressed ? "," : ", ");
     while (start < end && *start) {
       Complex_Selector* sel = *start;
-      if (!str.empty()) str += ", ";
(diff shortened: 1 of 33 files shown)


mgreter:
We got :green_heart: against ruby sass master (upcoming 3.4.21)

mgreter:
Rebased branch and renuked latest sass-specs ...

xzyfer:
There's really no way I can do a meaning review on a PR this big :(

Do what you need to do I guess.

Issue #1834, opened 2016-01-07
No file information for empty source string


k0kubun:
I'm using libsass via sassc-ruby.

I got a following backtrace for empty source scss.

```
SassC::SyntaxError: Error: Data context created with empty source string
/Users/k0kubun/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/sassc-1.8.2/lib/sassc/engine.rb:41:in `render'
```

This may be raised in https://github.com/sass/libsass/blob/3.3.2/src/sass_context.cpp#L390.

I fixed sassc-ruby locally to see filename of empty scss and I found my empty scss. I want to see filename by default for this error. While I can fix sassc-ruby to see it, I think this can be resolved in libsass layer.

PR #1845, opened 2016-01-09
Disable error for empty source string on context

mgreter:
Fixes https://github.com/sass/libsass/issues/1834


k0kubun:
:cool: 

Thank you!

Issue #1802, opened 2015-12-27
Prefix constants with a prefix to prevent collisions


xzyfer:
> I think we should just rename all constants and prefix them with something like LIBSASS_ to avoid issues like the SEC one. Otherwise it'll be the endless chase...
>https://github.com/sass/libsass/pull/1782#issuecomment-166031233.

I believe we start prefixing some constants with `SASS_` 3.3. We should finish the job.

mgreter:
IMO it is okay as it is now. So far only C-API stuff is prefixed with `SASS_`, since we have a `Sass` namespace on the C++ side. The issues we had were due to `using namespace std` which has been dropped, so it should be ok now. I really would like to encourage to work with namespaces in C++ (i.e. Units now live in `SassUnit` namespace). There will always be compiler quirks as ie. the one with `std::round` that surfaced on cygwin when we prefixed all identifiers from std.

saper:
What about just having `LIBSASS_` prefix on all our constants?

C++ namespaces will not solve `SEC` issue afaik (the preprocessor will still try to expand it!)

mgreter:
I absolutely would hate that, but I guess that's already clear. So we would have `Sass::SassUnit::LIBSASS_SEC` just because one compiler can't keep it together? I say let's keep it as it is with the `#undef SEC` since that macro should never be defined in the first place!

saper:
It is not the compiler, those are the header files. If you don't like it we can drop SassUnit, since that's a bit too much. it's a known problem in the C world and most libraries try to namespace like this, this is nothing unusual. Windows is also adding random symbols defined here and there, you might by lucky with one SDK or version setting, but not with another. We can't change a whole world.

QuLogic:
Is the `LIB` part really necessary though?

mgreter:
So I took the pleasure to install smartos on a virtual machine, which took me 3 hours [due to a bug] [1] in their installer :facepunch: Once I had it running it took me another hour to figure out how to [install] [2] gcc (`pkgin in scmgit-base gcc47; export PATH="$PATH:/opt/local/gcc47/bin/"`). Once I was that far it took me a couple of minutes to figure out how to avoid these macros to be exported:

In src/units.cpp:
```
#undef __EXTENSIONS__
#include <stdexcept>
#include "units.hpp"
```

[...]

PR #1847, opened 2016-01-10
Fix SmartOS compilation once and for all

mgreter:
Verified by compiling inside VMWare image (the problems with `SEC` were present):
```
SunOS smartos 5.11 joyent_20160108T173524Z i86pc i386 i86pc
```

Let's see what CI thinks ... https://github.com/sass/libsass/issues/1802

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,10 @@ CP       ?= cp -a
 MKDIR    ?= mkdir
 RMDIR    ?= rmdir
 WINDRES  ?= windres
-ifeq ($(OS),SunOS)  # Solaris/Illumos flavors
+# Solaris/Illumos flavors
+# ginstall from coreutils
+ifeq ($(OS),SunOS)
 INSTALL  ?= ginstall
-PREFIX   ?= /opt/local
 endif
 INSTALL  ?= install
 CFLAGS   ?= -Wall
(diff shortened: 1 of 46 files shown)


mgreter:
The last commit [reverts the SunOS specifics] [1] in the Makefile:

1) We need to have `libtool` installed and `ginstall` is not available by default too!
2) We install to `/` on all platforms (use `PREFIX="/opt/local"` to override).

SmartOS needs: `pkgin install automake autoconf libtool`!
Then we can also build libsass via autotools as [documented] [2].

[1]: https://github.com/sass/libsass/commit/2e0e379411e5a20dda88c9936a4126f6ac323ef8
[2]: https://github.com/sass/libsass/blob/master/docs/build-with-autotools.md

xzyfer:
@jaredmorrow @djensen47 I'd like to hear your thoughts here

mgreter:
Looks like `ginstall` would still be the correct one to use. libtool install doesn't seem to do the right thing, was asuming it would. But no idea how to get gnu-install on to SmartOS. Maybe if someone can argue why we cannot install into `/` on Solaris, please enlighten me.

mgreter:
Verfied both install methods and documented my findings. Very very sure this chapter can be closed!

mgreter:
I'm going to merge this on my responsibility. Beside CI I have tested it with Visual Studio 2013 Community, SmartOS 5.11, Gentoo Linux, Cygwin and MinGW. Not all combinations, but a fair share.
Want to move forward and concentrate to bring sass 3.4.21 parity!
As always feel free to open issues when problems arise.
Have a nice day!

Issue #1768, opened 2015-12-01
@debug'ing an empty list should produce ()


xzyfer:
```
@debug(());
@debug(foo, (), bar);
@debug(foo () bar);
@debug((foo: (), bar: baz));
```
Ruby Sass
```
test.scss:1 DEBUG: ()
test.scss:2 DEBUG: foo, (), bar
test.scss:3 DEBUG: foo () bar
test.scss:4 DEBUG: (foo: (), bar: baz)
```
LibSass 3.3.2
```
/test.scss:1 DEBUG:
/test.scss:2 DEBUG: foo, bar
/test.scss:3 DEBUG: foo bar
/test.scss:4 DEBUG: (foo: , bar: baz)
```
Spec https://github.com/sass/sass-spec/pull/633

mgreter:
With the new `to_sass` mode this one became easy :smile:

PR #1862, opened 2016-01-17
Use `to_sass` for `@debug`, `@warn` and `@error` message

mgreter:
Specs: https://github.com/sass/sass-spec/pull/702 (fixes: https://github.com/sass/libsass/issues/1768)

### DIFF
--- a/src/eval.cpp
+++ b/src/eval.cpp
@@ -337,7 +337,7 @@ namespace Sass {
 
     }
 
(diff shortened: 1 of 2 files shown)


xzyfer:
So much nicer.

mgreter:
All thanks to the decoupling of context from the inspect visitor, took me quite some time and refactorings to finally make these happen.

Issue #1856, opened 2016-01-15
Incorrect 'Invalid UTF-8' error with a specific input length


am11:
Input (data context):

```scss
xxxŢxxxŢŢŢ{y:z;
```

Ruby Sass outputs:

```
Invalid CSS after "xxxŢxxxŢŢŢ{y:z;": expected "}", was ""
```

LibSass outputs:

```
Invalid UTF-8 on line undefined at column undefined
```

Drop first character from the input:

```scss
xxŢxxxŢŢŢ{y:z;
```

and now LibSass produces correct error:

```
Invalid CSS after "xxŢxxxŢŢŢ{y:z;": expected "}", was "" on line 1 at column 20
```

Note in case of file context (e.g. `sassc test.scss`), even with one `x` dropped does not suffice; at the minimum it should be:

```scss
xxxŢŢŢ{y:z;
```

to get the error right.

mgreter:
This should be fixed by https://github.com/sass/libsass/pull/1875, but spec tests show an issue with the spec runner and unicode, so the specs cannot be merged at that point. https://github.com/sass/sass-spec/issues/714

am11:
Thanks! :+1: :tada:

PR #1875, opened 2016-01-17
Make generating source excerpt on error unicode aware

mgreter:
Fixes https://github.com/sass/libsass/issues/1856

### DIFF
--- a/src/utf8_string.cpp
+++ b/src/utf8_string.cpp
@@ -69,6 +69,8 @@ namespace Sass {
       }
     }
 
+    #ifdef _WIN32
+
     // utf16 functions
     using std::wstring;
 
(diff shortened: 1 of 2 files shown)


src/utf8_string.cpp
```diff
@@ -94,5 +96,7 @@ namespace Sass {
       return utf16;
     }
 
+    #endif
+
```
am11: @mgreter, I kept MinGW and MSVC implementations separate here: https://github.com/sass/libsass/blob/e6ac0a2/src/file.cpp#L28-L44 (as GCC hasn't implemented C++11 `<codecvt>`, so MinGW implementation should make a win32 API call for conversion instead).
IMO, if this solution works for both, we should probably use the same there. Either way, we should probably refactor this out at one place in `util.cpp` for DRY.

PR #1884, opened 2016-01-20
Ensure we keep around memory for parser source

mgreter:
Fixes an issue exposed by MSVC debug build as seen in https://github.com/sass/libsass/pull/1883


mgreter:
Sorry, but I have to say appveyor often acts like b****.
Makes it difficult to tell who is to blame for a failed CI!
It "failed" this PR because I closed https://github.com/sass/libsass/pull/1883, which contained the same commit as this PR. Restarting CI now with hopefully :green_heart: in 20 Minutes.

PR #1888, opened 2016-01-20
Switch back to Visual Studio 2013

saper:
Resolve incompatibility with node that
is still built using VS2013


xzyfer:
:+1:

xzyfer:
For context: https://github.com/sass/node-sass/issues/1283

Issue #1656, opened 2015-10-28
Add option to control CSS inline import behaviour


xzyfer:
We've added the ability to inline `@import` CSS partials. This is significantly different from the Ruby Sass behaviour (which does not allow this at all) and is understandable [causing pain](https://github.com/sass/node-sass/issues/1222#issuecomment-151763178).

We need a context option to enable/disable this feature. It should be disabled by default.

mgreter:
I think the wording is a bit misleading. We only allow imports of css partials (stuff ending with .css will left as an actual import in the code). Also Ruby Sass does not allow this by default, but a few tools (AFAIK compass too) enable this by default for Ruby Sass. I'm also not sure if we really want to disable it by default, since this would be another potentially breaking change.

```
class CSSImporter < Sass::Importers::Filesystem
  def extensions
    super.merge('css' => :scss)
  end
end
Sass::Plugin.options[:filesystem_importer] = CSSImporter
```

xzyfer:
Full disclosure I still think this feature should be removed. Our directive is to do what Ruby Sass does. Anything deviation from their implementation breaks portability, and is indistinguishable from a bug.

>Also Ruby Sass does not allow this by default, but a few tools (AFAIK compass too) enable this by default for Ruby Sass

This isn't our job. This belongs in the realm of tools like wellington and eyeglass.

> I'm also not sure if we really want to disable it by default

I vehemently believe it should. 

> this would be another potentially breaking change

The current behaviour is broken as far as Sass users are concerned.

mgreter:
I guess it's clear that I'm in favor of having this feature and just wanted to point out that the breakage is mainly only because I've implemented the error message for ambigous imports for libsass 3.3.0. Beside that we only differ by default in partial resolving (also includeing css extension), and I would argue that normally you don't generate or have a css with the same name beside a *partial* import. Unsure when I will come around to implement that as an option on the C-API. Contributers welcome ;)

esr360:
I think the ideal situation would be that they both support this features. The next best thing would be as @xzyfer says and to at least have them be consistent, even if that means neither of them supporting it.

You should be able to switch from libsass to ruby sass in theory without anything breaking. Having something which would knowingly prevent this from happening imo is a bad thing.

xzyfer:
> You should be able to switch from libsass to ruby sass in theory without anything breaking

Exactly.

> The next best thing would be as @xzyfer says and to at least have them be consistent

This is what we will do.

[...]

PR #1893, opened 2016-01-22
Change default from `scss` to `sass` syntax

mgreter:
- https://github.com/sass/libsass/pull/1341
- https://github.com/sass/sass/issues/1726

Ruby sass uses the extension, so to be on par we must too. Auto-detection of the format may be fancier, but it's not what ruby sass does AFAIK. Personally I think think this should be changed with sass 4.0 (//CC @chriseppstein) and scss should be the defaut, if extensions doesn't explicitly match .sass.

Otherwise https://github.com/sass/libsass/issues/1219 will stay open forever!

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -404,7 +404,7 @@ namespace Sass {
       }
       for(size_t i=0; i<extension.size();++i)
         extension[i] = tolower(extension[i]);
-      if (extension == ".sass" && contents != 0) {
+      if (extension != ".scss" && contents != 0) {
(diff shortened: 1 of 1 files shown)


xzyfer:
:+1: :100: :100:

xzyfer:
This is probably a good opportunity to disable `.css` importing by default and put it behind a flag.

mgreter:
@xzyfer that would be https://github.com/sass/libsass/issues/1656, let's keep these seperated.

xzyfer:
Sorry meant to close this not merge it.

We've decided not to change the existing behaviour until 4.0. See https://github.com/sass/libsass/issues/1656#issuecomment-220745629

Issue #1890, opened 2016-01-21
@at-root (without: media) selector outside from wapper


sou-lab:
When use "@at-root (without: media)".
`.box` is outside from `.wrap`

``` scss
.wrap {
  @media (min-width: 480px) { 
    display: block;
    @at-root (without: media){ 
      .box { 
        display: inline-block;
      }
    } 
  }
}
```

LibSass

``` css
@media (min-width: 480px) {
  .wrap {
    display: block;
  }
}

.box {
  display: inline-block;
}

```

Ruby Sass 3.4

``` css
@media (min-width: 480px) {
  .wrap {
    display: block;
  }
}
.wrap .box {
  display: inline-block;
}
```

Spec added https://github.com/sass/sass-spec/pull/749
##

sou-lab:
if `.box` at root

```scss
.wrap {
  @media (min-width: 480px) { 
    display: block;
    @at-root (without: media rule) {
    // or  @at-root (without: all) {
      .box { 
        display: inline-block;
      }
    } 
  }
}
```

mgreter:
This is basically a duplicate of https://github.com/sass/libsass/issues/1661 and https://github.com/sass/libsass/issues/1873

xzyfer:
I've compared our at-root implementation with Ruby Sass. The at-root is being resolved correctly in this case. The problem here is that the `.box` selector is not fully resolved (parentized) in the expand/eval stage.

mgreter:
Related to https://github.com/sass/libsass/issues/2444 and https://github.com/sass/libsass/issues/1651

PR #1900, opened 2016-01-27
[WIP] POC of refactor to support at-root correctly

mgreter:
@xzyfer this is a very early WIP PR which currently should pass all spec tests (AFAICS minus one error spec) and fixes about ~5 currently open issues. Many unused code duplicates of failed trials need to be removed and consolidation needs to be done for quite a few aspects. This branch will act as a starting point for that work, so I'm posting it here since sharing is caring.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -5,6 +5,7 @@
 #include "extend.hpp"
 #include "emitter.hpp"
 #include "color_maps.hpp"
+#include "debugger.hpp"
 #include <set>
 #include <iomanip>
 #include <algorithm>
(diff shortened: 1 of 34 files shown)


xzyfer:
Can you please reference the issues this fixes?

For the record we currently support with/without and there [test coverage](https://github.com/sass/sass-spec/tree/master/spec/libsass/at-root/with_without). it'd be good to clear about what specifically this is aiming to fix.

mgreter:
@xzyfer we need to move the parentization to the cssize stage, since we otherwise "freeze" it too early. Extends happen after the eval stage and therefore parent references will be wrong.

xzyfer:
I'm pretty sure this is fixed in #2089. We can look at at-root again when we have a specific known failure.

xzyfer:
After some further investigation there are still problem related to at-root but they stem from broken parentizing, not the at-root implementation itself. Specific issues are #2006 and #1890 both of which are incorrectly parentized before reaching CSSize.

The issue is certainly with our parentizing, so any fix should not touch the cssize visitor since it is line for line compliant with Ruby Sass.

PR #1902, opened 2016-01-31
Fix a few typos in the parser's comments

kaelig:
I can't code in C/C++ (yet) but I can use a spell checker, so here's my $2.


xzyfer:
Thanks

Issue #1891, opened 2016-01-21
missing byte order mark (aka bom)


gabriellecrawford:
This is definitely not my area of expertise, so bear with me, but I think in compressed form the byte order mark (bom) is missing. 

here's my sass file
```
$oj-fwk-icon-clock: "\e60a" !default;

.clock:before {
  content: $oj-fwk-icon-clock;
}
```

this is the output for the uncompressed case, clearly the unicode has been detected:
```
@charset "UTF-8";
.clock:before {
  content: ""; }
```

However in the compressed version there's no @charset, from reading the sass doc linked below the compressed version should have a BOM instead
http://sass-lang.com/documentation/file.SASS_REFERENCE.html#encodings

I followed advice on detecting the BOM here:
https://www.w3.org/International/questions/qa-byte-order-mark

I uploaded my compressed css file here and it says there's no BOM
http://validator.w3.org/i18n-checker/#validate-by-upload+

mgreter:
Should work from libsass side: https://github.com/sass/libsass/blob/master/src/output.cpp#L72-L75, please open an issue with the implementor you are using. Thanks!

gabriellecrawford:
I'm sorry, but I don't understand this answer, by implementor do you mean node-sass? My team is using libsass and grunt...

mgreter:
Yes, you never stated which upstream tool you use, so if you use node-sass open an issue with them.

gabriellecrawford:
thank you! I'm wondering one more thing, can file a separate issue depending on your answer. If I explicitly put @charset "UTF-8"; at the beginning of my sass file it's stripped, shouldn't it be passed through?

mgreter:
No, we only support utf-8 output and we auto-detect if such a charset is needed or not. If one is present in input and is not really needed, we would strip it. I think that ruby sass does it the same way (it supports importing different charsets though, libsass has no means to convert ie. latin1 to utf8).

PR #1914, opened 2016-02-09
Handle signed char when considering @charset

asottile:
Resolves dahlia/libsass-python#117

On platforms where `char` is unsigned the `@charset` (or UTF-8 BOM) was not present in the output.  This seems to fix it.


xzyfer:
LGTM.

xzyfer:
This is potentially related to https://github.com/sass/libsass/issues/1891

PR #1921, opened 2016-02-17
Fix import resolution order changing with custom importers

xzyfer:
When using the following syntax with a custom importer

```css
@import "foo", "bar";
```

Paths that are matched by a custom importer get pushed into the
queue ahead of those that don't. This means if `bar` is matched
by a custom importer the import order will be changed to
`"bar", "foo"` instead of the intended `"foo", "bar"`. This can
easily break user code.

See sass/node-sass#1381 for an example

/cc @nowells

/cc @mgreter does this look sane to you? Passes the Node Sass tests.

### DIFF
--- a/src/debugger.hpp
+++ b/src/debugger.hpp
@@ -357,6 +357,7 @@ inline void debug_ast(AST_Node* node, std::string ind, Env* env)
     Import_Stub* block = dynamic_cast<Import_Stub*>(node);
     std::cerr << ind << "Import_Stub " << block;
     std::cerr << " (" << pstate_source_position(node) << ")";
+    std::cerr << " [" << block->imp_path() << "] ";
     std::cerr << " " << block->tabs() << std::endl;
   } else if (dynamic_cast<Import*>(node)) {
     Import* block = dynamic_cast<Import*>(node);
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -290,12 +290,7 @@ namespace Sass {
     do {
       while (lex< block_comment >());
       if (lex< quoted_string >()) {
-        if (!ctx.call_importers(unquote(std::string(lexed)), path, pstate, imp))
-        {
-          // push single file import
-          // import_single_file(imp, lexed);
-          to_import.push_back(std::pair<std::string,Function_Call*>(std::string(lexed), 0));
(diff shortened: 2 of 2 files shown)


nowells:
Awesome fix @xzyfer! Thanks for fixing this!

mgreter:
@xzyfer did you verify https://github.com/sass/libsass/issues/1262 does still work as expected? On a quick glance I don't see any obvious problems, but I also don't see how it could go out of order. I guess that was happening in call_importers? Might have some time this weekend to test it via perl-libsass.

xzyfer:
@mgreter 

> did you verify #1262 does still work as expected? 

Yes. All our specs and node sass specs pass.

>I also don't see how it could go out of order. I guess that was happening in call_importers? 

Correct. In the case of `@import "foo", "bar"` where `bar` is handled by a custom importer. It gets out of order because `foo` is pushed onto the `to_import` vector, and `bar` is pushed on to the import queue in `call_importers`. Only later when when is `foo` pushed onto the importer queue from the `to_import` vector.

In the case that neither or both files are handled by custom importers the files are pushed on to the import queue or `to_import` vector in the correct order.

xzyfer:
> Might have some time this weekend to test it via perl-libsass

Yeah I pinged you so you could test this with perl-libsass because you have better test coverage for custom importers than node sass and LibSass.

mgreter:
Just to ask, but do we have an actual test case for #1262 in sass specs (or do you have one in node sass). Not sure I do in perl-libsass. Anyway, will try to get some time this weekend to check this! Beeing busy with another project atm.

xzyfer:
Hmm good point, add one should be trivial.

We have a [test in PR](https://github.com/sass/node-sass/pull/998) in node-sass that hasn't been merged. I'll look into getting it up to date and test this patch this arvo.

xzyfer:
I've confirmed this patch still passes #1262. In my investigation I noticed some other issues with custom importers. I'll create some new issues and PRs later today.

mgreter:
My existing spec tests still work, so LGTM!

NickHeiner:
Awesome. Thanks for merging this. I'm excited for the next release so I can start using this fix!

PR #1925, opened 2016-02-23
Bugfix regressions 3.3.3

mgreter:
#1901, #1907, #1916

PR #1958, opened 2016-03-19
Fix regression with parentize of wrapped selector

xzyfer:
This is simply #1925 rebased off master (3.3.4). This change was not merged in 3.3.4 because it's reasoning wasn't explained in the PR.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -961,6 +961,38 @@ namespace Sass {
     return ss;
   }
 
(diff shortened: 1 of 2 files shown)


mgreter:
Not sure and it doesn't seem to fix any open issues ATM, so rescoped for now.
Not closing it now just to have it on the radar, probably later if nothing turns up.

mgreter:
Closing as propably obsolete

PR #1963, opened 2016-03-26
Fix import file resolution to match Ruby Sass

### LINKED ISSUE #1656: Add option to control CSS inline import behaviour

xzyfer:
This patch address two inconsistencies with how `@import` is
resolved compared to Ruby Sass.
- `.css` should not be `@import`able
- `.sass` should take precedence over `.scss` files

Fixes #1656

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -305,7 +305,7 @@ namespace Sass {
       std::string filename = join_paths(root, file);
       // supported extensions
       const std::vector<std::string> exts = {
-        ".scss", ".sass", ".css"
+        ".sass", ".scss"
       };
       // split the filename
       std::string base(dir_name(file));


mgreter:
IMO this should wait for 3.4.0 (breaking change) and we should add an opt in config to preserve the current behavior if wanted. We added it because it was widely requested and ruby sass also allows to alter file extensions for including partials.

martinf55:
Commenting just so I can find this again easily.

xzyfer:
@martinf55 in future click the subscribe button on the left instead of commenting on issues.

xzyfer:
We have decided keeping the current behaviour, and not breaking the code for people who now depending on this, is the lesser evil. 

This will be addressed in 4.0.

PR #1964, opened 2016-03-26
[RFC] Add an option for allowing the old .css import behaviour

xzyfer:
The broken behaviour was removed in #1963. There appears to be some support for this, although I'm very much against it. Additionally we have no way presently to test flags like this.

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -77,6 +77,8 @@ namespace Sass {
     c_importers             (std::vector<Sass_Importer_Entry>()),
     c_functions             (std::vector<Sass_Function_Entry>()),
 
+    css_imports             (false),
+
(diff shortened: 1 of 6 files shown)


mgreter:
Looks quite ok too me, but instead of creating a specific option for another extension (.css) I would prefer to have the full `std::vector<std::string>` exposed as far as possible. This means bridging between C-API and C++ will be a bit more complicated. But we may can work with a similar apporach as with `push_include_dir` (IMO we don't need full manipulation support on the C side).
The C-API part would then look something like this:

`sass_option_push_import_extension($options, "css");`

This would be very close to ruby sass, as they also allow to alter the list of extensions it will try to resolve.

PR #1967, opened 2016-03-29
Ensure sass spec is using specs for 3.4

xzyfer:
Sass spec has been updated to support specs for multiple versions
of Sass in sass/sass-spec#745. It currently defaults to 4.0 but we
want to run against 3.4.


chriseppstein:
Merged as 10d82926d3683752bd64705ab93b4a812a2c7747

PR #1973, opened 2016-04-04
Add sass_string_alloc public API function to allow to allocate a string owned by libsass

xoofx:
Hi there!

While developing a wrapper of libsass for .NET (I can't use [libsass.net](https://github.com/darrenkopp/libsass-net) as it is using a wrapping technique that is not portable on Linux...etc.), I had to add a new function API `sass_string_alloc` in order to be able to allocate a string that I have to pass to libsass (and owned by libsass)

From a public API point of view, there is currently no way to do this. Typically, when wrapping libsass in C#, I can only access the DLL and don't have access to any C function like `malloc`. Also, it might be good to enforce a public API for this, so that you can hide the internals about requiring malloc or not, track allocated strings...etc.

Hope this can make it, as I don't have any workaround for this, let me know if there is any issues (when I saw the comment `Sadly, sass_strdup is not portable`, I was unsure about this problem)

Thanks!

### DIFF
--- a/include/sass/base.h
+++ b/include/sass/base.h
@@ -63,6 +63,7 @@ enum Sass_Output_Style {
 };
 
 // Some convenient string helper function
+ADDAPI char* ADDCALL sass_string_alloc(size_t length); // Allocates a string of length+1 (terminated by zero)
 ADDAPI char* ADDCALL sass_string_quote (const char* str, const char quote_mark);
 ADDAPI char* ADDCALL sass_string_unquote (const char* str);
 
--- a/src/sass.cpp
+++ b/src/sass.cpp
@@ -11,6 +11,12 @@
 extern "C" {
   using namespace Sass;
 
+  // Allocates a string owned by libsass
+  char* ADDCALL sass_string_alloc(size_t length)
+  {
+	  return sass_stralloc(length);
+  }
+
   // caller must free the returned memory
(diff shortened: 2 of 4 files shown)


xzyfer:
@xoofx if I'm understanding correctly this seems related to #1821. Can you please take a look and confirm whether that would address your issue?

xoofx:
@xzyfer indeed, that's quite related!, sorry I didn't look much further at existing PR. Though, I would prefer to have an alloc function instead of only a string duplicate, so that I can avoid double allocating while I can dump directly the UTF8 string to the allocated slot.

xzyfer:
@xoofo I think it'd be better to contribute your usecase to the existing issue. There's already a lot of discussion there I'd prefer to have a single "right way" to address this problem.

mgreter:
Closing as duplicate of https://github.com/sass/libsass/pull/1821. I can say that I personaly also prefer a way to avoid unnecessary memory copies and I already came to pretty much the same conclusion that we need to expose our memory allocator somehow. So it seems to make sense to even expose the lowest level `sass_allocate_memory`. The use case would be so implementors allocate the buffer first and can fill it with data themselve before passing it to libsass.

mgreter:
BTW. the C-API changes are scheduled for the next middle version increase (3.4.0). We are not following semver strictly as we want to keep version numbers to match with ruby sass (so we basically have one try left to get it right for 3.4).

Issue #1871, opened 2016-01-17
API bug: Preserve include paths which were pushed to linked list


am11:
At present, consumers have two choices to set include paths: `sass_option_set_include_path` and `sass_option_push_include_path`.

The first one requires a string with concatenated paths, joined by path separator. LibSass internally split the string with platform-aware path separators.

Most consumers, including node-sass, are using this approach and it works.

The second one lets the consumer to insert one path at a time to internal `include_paths` linked list.

With second approach, the paths are pushed correctly, however, parser ignores those paths. This is because when `cpp_ctx` is instantiated, the existing `include_paths` are not copied from the original context. See https://github.com/sass/libsass/blob/1c92e8f/src/context.cpp#L95 (the line is commented out).

IMO. this can either be fixed in the existing form or `sass_option_push_include_path` should be removed from API surface area and docs. :)

mgreter:
This was disabled in https://github.com/sass/libsass/commit/ea573488891ed34342d56d41cc5cc388c27859ef, but I don't remember why ...

am11:
I am setting it like this: https://github.com/am11/libsass-net/blob/72dde69/src/SafeSassContextHandle.cs#L182-L188.

I also debugged the code around `include_path[s]` and found that `sass_option_push_include_path` does get called from different code paths, but strings inserted into linked list before `compile()` are ignored during the compilation. It seems like `options->include_paths` and `context->include_paths` are redundant and the one in options is not copied in context ctor (and hence remain unused).

If possible, can we unify both and have one `include_paths` linked list (which gets flattened as array later as well, another overhead)?

PR #1974, opened 2016-04-04
Fix sass_option_push_include_path / sass_option_push_plugin_path 

xoofx:
This is a fix for #1871 

Trying to use `sass_option_push_include_path` / `sass_option_push_plugin_path` I discovered that they were completely unplugged, while these APIs are useful to push path when the list is already prepared on the user side and we don't want to concatenate the includes (that will be splitted anyway by libsass)

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -92,9 +92,9 @@ namespace Sass {
 
     // collect more paths from different options
     collect_include_paths(c_options.include_path);
-    // collect_include_paths(c_options.include_paths);
+    collect_include_paths(c_options.include_paths);
     collect_plugin_paths(c_options.plugin_path);
-    // collect_plugin_paths(c_options.plugin_paths);
+    collect_plugin_paths(c_options.plugin_paths);
 
     // load plugins and register custom behaviors
     for(auto plug : plugin_paths) plugins.load_plugins(plug);
@@ -162,7 +162,6 @@ namespace Sass {
 
   void Context::collect_include_paths(const char* paths_str)
   {
-
     if (paths_str) {
       const char* beg = paths_str;
       const char* end = Prelexer::find_first<PATH_SEP>(beg);
@@ -185,17 +184,17 @@ namespace Sass {
     }
   }
 
-  void Context::collect_include_paths(const char** paths_array)
+  void Context::collect_include_paths(string_list* paths_array)
   {
-    if (!paths_array) return;
-    for (size_t i = 0; paths_array[i]; i++) {
-      collect_include_paths(paths_array[i]);
+    while (paths_array)
+    {
+      collect_include_paths(paths_array->string);
+      paths_array = paths_array->next;
     }
   }
 
   void Context::collect_plugin_paths(const char* paths_str)
   {
-
     if (paths_str) {
       const char* beg = paths_str;
       const char* end = Prelexer::find_first<PATH_SEP>(beg);
@@ -218,14 +217,14 @@ namespace Sass {
     }
   }
 
-  void Context::collect_plugin_paths(const char** paths_array)
+  void Context::collect_plugin_paths(string_list* paths_array)
   {
-    if (!paths_array) return;
-    for (size_t i = 0; paths_array[i]; i++) {
-      collect_plugin_paths(paths_array[i]);
+    while (paths_array)
+    {
+      collect_plugin_paths(paths_array->string);
+      paths_array = paths_array->next;
     }
   }
-
 
   // resolve the imp_path in base_path or include_paths
   // looks for alternatives and returns a list from one directory
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -198,44 +198,6 @@ extern "C" {
   static Sass_Compiler* sass_prepare_context (Sass_Context* c_ctx, Context* cpp_ctx) throw()
   {
     try {
-
-      // convert include path linked list to static array
-      struct string_list* inc = c_ctx->include_paths;
-      // very poor loop to get the length of the linked list
-      size_t inc_size = 0; while (inc) { inc_size ++; inc = inc->next; }
-      // create char* array to hold all paths plus null terminator
-      const char** include_paths = (const char**) calloc(inc_size + 1, sizeof(char*));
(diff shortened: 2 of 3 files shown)


src/context.cpp
```diff
@@ -185,12 +185,15 @@ namespace Sass {
...
 
-  void Context::collect_include_paths(const char** paths_array)
+  void Context::collect_include_paths(string_list* paths_array)
   {
     if (!paths_array) return;
-    for (size_t i = 0; paths_array[i]; i++) {
-      collect_include_paths(paths_array[i]);
-    }
+	
+	while (paths_array)
```
xzyfer: nit: indentation looks off

xzyfer:
@xoofx thanks for this. My gut tells me we may need both implementations for the sake of upstreams that already implement the old `char` interface.

Could you also please PR how this would look with [SassC](https://github.com/sass/sassc)?

src/context.cpp
xzyfer: Could be that you're using tabs?

xzyfer:
@xoofx never mind my previous comment, I'm hungry.

@am11 PTAL?

xoofx:
> @xoofx never mind my previous comment, I'm hungry.
No, I did a fix and push force in the meantime ;)

You can see above a PR for sassc using this API.

am11:
@xzyfer, I remember chatting about it with @mgreter on Slack that while `include_path` a string is functional, `include_paths` (plural) string array is not and has no affect on the output. By it looks, this change fixes that issue (I haven't tested it).

mgreter:
I think there was a compelling reason I disabled these, but I can't remember exactly what it was. @xoofx can you say how widely you tested it? We need to make sure it works with both APIs and preferably doesn't break if someone is funky enough to use both (IMO that we have a fixed order here is fine as doing so should not be encouraged anyway).

Other than that LGTM!

xzyfer:
We have scheduled breaking API changes for 4.0 so I'm :+1: to shipping this and getting userland feedback.

xzyfer:
I'm going to be AFK for the next week. @mgreter please ship this if you have no objections.

mgreter:
I don't even consider this one a breaking change but additional feature? But need to test this first. I hope I have some unit tests in perl-libsass for this. Maybe you can see and give feedback if node-sass is happy with the changes?

xzyfer:
I won't have time this week, maybe @saper will. The SassC change looks promising (https://github.com/sass/sassc/pull/166/files)

xoofx:
@mgreter I tested it only on my wrapper and in the sassc branch. I have checked that using `sass_option_set_include_path` together with `sass_option_push_include_path` was correctly appending paths. It is not really a breaking change nor an additional feature but really just a bug fix. Previous API were just no-op.

src/context.cpp
```diff
@@ -185,11 +185,14 @@ namespace Sass {
     }
   }
 
-  void Context::collect_include_paths(const char** paths_array)
+  void Context::collect_include_paths(string_list* paths_array)
   {
     if (!paths_array) return;
```
mgreter: This would no longer be needed (same while clause below)

src/context.cpp
```diff
@@ -218,15 +221,17 @@ namespace Sass {
     }
   }
 
-  void Context::collect_plugin_paths(const char** paths_array)
+  void Context::collect_plugin_paths(string_list* paths_array)
   {
     if (!paths_array) return;
```
mgreter: This would no longer be needed (same while clause below)

src/sass_context.cpp
```diff
@@ -198,44 +198,6 @@ extern "C" {
...
-      // copy over the paths
-      for (size_t i = 0; imp; i++) {
-        plugin_paths[i] = imp->string;
-        imp = imp->next;
-      }
-
-      // free intermediate data
-      free(include_paths);
-      free(plugin_paths);
-
```
mgreter: Maybe you konw/can explain why this is no longer needed?
I'm also a bit worried about https://github.com/sass/libsass/issues/1871#issuecomment-172405988
Last check should be if there are any memory leaks.
But it doesn't break perl-libsass, so LGTM so far.

mgreter:
Sorry to be a bit nitpicky @xoofx. Any chance you have made an "end to end" test with sassc that requres the include path via cmd line?

src/sass_context.cpp
xoofx: Well, the previous code was basically a no-op (filling include_paths and then after freeing it).
That being said, there is a slight issue with c_ctx->include_paths that will not be released.

But I'm really not enough sure how ownership is currently working in libsass, if you could enlighten me. Looking at the code, I would assume that `include_paths` should be free in the [sass_clear_context](https://github.com/sass/libsass/blob/master/src/sass_context.cpp#L565). Is it correct?

xoofx: ah, actually, https://github.com/sass/libsass/blob/master/src/sass_context.cpp#L527 seems to release things, should be ok then? :sweat_smile:

xoofx:
I have tested it on the command line, so yes, I made a test, though quite basic. I was assuming also that there are some unit test using it?

src/sass_context.cpp
mgreter: Well there is `std::vector<std::string> plugin_paths` in `context.hpp` and `struct string_list* plugin_paths` in `sass_context.hpp`. But yes, normally we can expect memory to be alive until the main context is freed. There were also a lot of refactorings in the C++ part to re-use more of the C structs directly.

mgreter:
Would like to wait for feedback from node-sass before merging (//CC @xzyfer, @am11, @saper) if their tests still pass. Other than that I don't see a reason we cannot give it a spin in 3.3.5 as incubating feature.

mgreter:
@xoofx can you rebase this to latest master so CI passes. Thanks!

xoofx:
> @xoofx can you rebase this to latest master so CI passes. Thanks!

done

xzyfer:
@mgreter I've updated node-sass to LibSass master + this patch. Test pass. I am not using the updated interface however because include paths as a delimited string is much cleaner for us.

xzyfer:
I'm happy to merge this and ship 3.3.5.

PR #1983, opened 2016-04-06
Expose mem allocator on C-API

mgreter:
Combining https://github.com/sass/libsass/pull/1973 and https://github.com/sass/libsass/pull/1821

Adds the following new C-API functions:

- `void* sass_alloc_memory(size_t size)`
- `char* sass_copy_c_string(const char* str)`
- `void sass_free_memory(void* ptr)`

IMO this should be future proof and both linked issues should be covered.

- [x] Update documentation
- [x] Delete obsolete interface
- [x] Get rid of `Sass::sass_strdup`

I suggest to add them with https://github.com/sass/libsass/pull/1974 to 3.3.5 as incubating features for 3.4.0.

//CC @xoofx, @am11, @drewwells


xoofx:
ship it! :sweat_smile:

xoofx:
Also a broader plan would be to replace all internal malloc/free so that they go through this API to allocate things (and internally also all C++ objects would use the same allocator), so that custom mem allocator could then be pluggable.

am11:
LGTM - Thanks @mgreter ! 🎉

xzyfer:
I plan to release 3.3.5 this week. I'd do it sooner but I'm at a conference. 

@mgreter can you please get started on some release notes for this change?

am11:
@mgreter,

> Adds the following new C-API functions:
• void* sass_alloc_memory(size_t size) 
• char* sass_copy_c_string(const char* str) 
• void sass_free_memory(void* ptr) 


can we `ADDAPI`fy these three signatures as well, so they are visible for FFI (P/Invoke)? Currently `dumpbin /exports libsass.dll` does not list them.

At the moment I am using local patch to make it work:

```diff
 // to allocate buffer to be filled
-void* sass_alloc_memory(size_t size);
+ADDAPI void* ADDCALL sass_alloc_memory(size_t size);
 // to allocate a buffer from existing string
-char* sass_copy_c_string(const char* str);
+ADDAPI char* ADDCALL sass_copy_c_string(const char* str);
 // to free overtaken memory when done
-void sass_free_memory(void* ptr);
+ADDAPI void ADDCALL sass_free_memory(void* ptr);
```

PR #1985, opened 2016-04-07
Fix issue with (too strict) assertion in MSVC debug mode

mgreter:
A newly activated spec test triggered a debug assertion with MSVC and `isspace`.
The char in question was an `ö` which as `char` seems to have the value `-4`.
Therefore the assertion was triggered (IMO MSVC should know how to deal with chars).
Anyway, changing it to `unsigned char` fixed the issue (hope that works with gcc too).

http://stackoverflow.com/questions/28513332/debug-assertation-failed-c-1-c-255


xzyfer:
FYI we can into an similar issue reported by the python binding team in 3.3.4.

Issue #1417, opened 2015-07-29
Slashes should not be interpreted as division in an argument of a missing function


andrew-skybound:
Slashes should not be interpreted as division when they appear in arguments for missing functions according to the same rules as when a slash appears in a declaration.

Input:
```
@function test($a, $b) {
  @return ($a $b);
}
p {
  foo: 1px / 2px;
  foo: 1px / round(1.5);
  foo: test(1px / 2px, 1px / round(1.5));
  foo: missing(1px / 2px, 1px / round(1.5));        // output will be incorrect in libsass
  foo: call(missing, 1px / 2px, 1px / round(1.5));  // output will be incorrect in libsass
}
```

Ruby SASS Output:
```
p {
  foo: 1px / 2px;
  foo: 0.5px;
  foo: 0.5 0.5px;
  foo: missing(1px/2px, 0.5px);
  foo: missing(1px/2px, 0.5px);
}
```

Libsass Output:
```
p {
  foo: 1px/2px;
  foo: 0.5px;
  foo: 0.5 0.5px;
  foo: missing(0.5, 0.5px);
  foo: missing(0.5, 0.5px);
}
```

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/462

PR #1986, opened 2016-04-08
Patch division delay (keep slash) with function calls

mgreter:
Fixes https://github.com/sass/libsass/issues/1417
Spec sass/sass-spec#762


xzyfer:
This PR has introduced a regression with the sassdash library. I'm currently working on a reduced test case. Given old this issue is I'd prefer to revert this patch and rescope the fix.

PR #1995, opened 2016-04-13
Update to use upcoming sass-spec refactor

chriseppstein:
This is not merge-ready, I want to test changes to the libsass build against my sass-spec branch in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
     - os: osx
       compiler: gcc
     - os: osx
-      env: AUTOTOOLS=no COVERAGE=yes BUILD=static
+      env: AUTOTOOLS=no BUILD=static
 
 script: ./script/ci-build-libsass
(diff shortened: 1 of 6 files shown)


xzyfer:
The test failures appear to be related to changes in whitespace/new line normalisation.
AppVeyor appear to be running sass-spec@master which is why it's passing. You'll need to update `appveyor.yml` to checkout your branch.

xzyfer:
Interesting, AppVeyor failed sass-spec (invalid -V flag) but didn't fail the build. We'll need to look in to that separately.

chriseppstein:
There was a bug in the new spec runner that caused tests that needed to be "cleaned" to not be. That and some tests with a todo in their name weren't marked as pending yet in the metadata. Once those were sorted out, all the specs appear to be passing now.

chriseppstein:
merged as 10d82926d3683752bd64705ab93b4a812a2c7747

Issue #1392, opened 2015-07-27
@extend loops output selectors in the wrong order


xzyfer:
``` scss
.foo {a: b; @extend .bar}
.bar {c: d; @extend .baz}
.baz {e: f; @extend .foo}
```

Ruby Sass

``` css
.foo, .baz, .bar { a: b; }
.bar, .foo, .baz { c: d; }
.baz, .bar, .foo { e: f; }
```

LibSass

``` css
.baz, .bar, .foo { a: b; }
.foo, .baz, .bar { c: d; }
.bar, .foo, .baz { e: f; }
```

Spec https://github.com/sass/sass-spec/tree/master/spec/libsass-todo-tests/180_test_basic_extend_loop
Spec https://github.com/sass/sass-spec/tree/master/spec/libsass-todo-tests/181_test_three_level_extend_loop
##

mgreter:
Have seen this one already, IMO it only happens with circular extends ...

xzyfer:
As pointed out by @andybenedict in https://github.com/sass/libsass/issues/1539 this can actually cause broken code when attribute selectors are involved.

```scss
.foo{
  &[disabled]{
    display: none;
  }
}

.bar {
  @extend .foo;
}
```
Libsass
```
.foo[disabled], [disabled].bar {
  display: none;
}
```

mgreter:
@chriseppstein maybe you have any pointers as to how the ordering works?

mgreter:
This is another interestingly similar issue:

```scss
.thing {
    color: red;
}
.thing[disabled] {
    @extend .thing;
    background: blue;
}
```

ruby sass:
```css
.thing, .thing[disabled] {
  color: red; }
.thing[disabled] {
  background: blue; }
```

libsass:
```css
.thing, .thing[disabled] {
  color: red; }
[disabled].thing {
  background: blue; }
```

chriseppstein:
I'd really like to make sass-spec agnostic of non-semantic differences like this. The order doesn't matter and I don't think libsass should work very hard to achieve that level of consistency when there are bigger fish to fry.

mgreter:
@xzyfer you pointed out that "by @andybenedict in #1539 this can actually cause broken code when attribute selectors are involved." Can you provide an example of this? I agree with @chriseppstein that it should not make any difference!?

xzyfer:
You're both correct. I realised my mistake but hadn't removed the comment. This is why I removed the milestone as per #2174

[...]

PR #2001, opened 2016-04-15
Refactor/directive parsing

### LINKED ISSUE #1233: Block comments should be preserved in "static" at rules
When using @'s it appears to fail.

```scss                  
@-moz-keyframes animatetoptop /* Firefox */ line 429
{
from {width:0%}
to {width:100%}
}
```
I'll get the following error.
```
[16:36:40] gulp-notify: [Compile Error] stylesheets\apps\tmm\_results.scss
429:17  top-level directive must be terminated by ';'
```

I'm using gulp-sass: 2.0.1

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/16485592-block-comments-should-be-preserved-in-static-at-rules?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgreter:
Spec test https://github.com/sass/sass-spec/pull/764 (excerpt from https://github.com/sass/libsass/pull/1900, fixes #1233)

- libsass-todo-issues\issue_1233\expected.nested
- libsass-todo-tests\scss-tests\127_test_moz_document_interpolation\expected.nested
- libsass-todo-tests\libsass\moz-document\basic\expected.nested
- libsass-todo-tests\libsass\moz-document\interpolated\expected.nested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # Miscellaneous stuff
 
+/sassc
+/sass-spec
+
 VERSION
 .DS_Store
 .sass-cache
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -60,7 +60,7 @@ namespace Sass {
   bool Compound_Selector::has_parent_ref()
   {
     for (Simple_Selector* s : *this) {
-      if (s->has_parent_ref()) return true;
+      if (s && s->has_parent_ref()) return true;
     }
     return false;
   }
@@ -1293,7 +1293,7 @@ namespace Sass {
   bool Selector_List::has_parent_ref()
   {
     for (Complex_Selector* s : *this) {
-      if (s->has_parent_ref()) return true;
+      if (s && s->has_parent_ref()) return true;
     }
     return false;
   }
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -7,8 +7,11 @@
 #include "color_maps.hpp"
 #include <set>
 #include <iomanip>
-#include <algorithm>
 #include <iostream>
+#include <algorithm>
+#include <functional>
+#include <cctype>
+#include <locale>
 
 namespace Sass {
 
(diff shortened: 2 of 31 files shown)


xzyfer:
If you're going to make this massive of a change I suggest doing your best to move more inline with Ruby Sass implementation like with cssize.

The closer our internal data flow and logic is to Ruby the easier it's going to be keep features in sync, and prevent weird edge cases.

mgreter:
If you check the commits this is exactly what is happening here!

xzyfer:
I wasn't insinuating it wasn't what you were doing. I just wanted to make sure we were on the same page.

mgreter:
Just assumed, because the main reason why this change is that massive is because it is following ruby sass closely in the first place :wink:

mgreter:
Just to be clear, I agree to follow ruby sass closely when it comes to the parser and the internal ast in general. But we can (and should) do better in libsass with manipulation and processing. I guess ruby sass choose to do quite a few operations (I'm specially refering to extend) in a monolithic approach. This makes sense in a language with dynamic function dispatching, since adding to many method calls in tight loops (code hotspots) can lead to bad performance. We don't have this constraint in c++ and should embrace the possibility of micro functions for selector manipulations. Another reason for that is that writing big monolithic functions in C++ (or any strict types language for that matter) tends to get tiresome, specially when a lot of loops are involved. But so far extend works and I don't want to touch any of that directly ported `Node` class if possible. Which reminds me of https://github.com/sass/libsass/issues/1392 ...

xzyfer:
I agree. I tend to avoid the extend code this exact reason.

xzyfer:
You'll probably have to rebase your sass-spec branch of master.

PR #2021, opened 2016-04-21
Expose the libsass language version via API.

chriseppstein:
I'm not sure if this 100% right. I can't seem to make autoconf work on my local env. But it builds and prints "[NA]" in the patched version of sassc that I will also submit.

Reason: Tracking the sass language version explicitly helps users correctly understand which version of the language they are using, it will also help sass-spec automatically derive the version of the tests to run without requiring `-V` all the time.

### DIFF
--- a/include/sass/base.h
+++ b/include/sass/base.h
@@ -79,6 +79,9 @@ ADDAPI char* ADDCALL sass_resolve_file (const char* path, const char* incs[]);
 // Get compiled libsass version
 ADDAPI const char* ADDCALL libsass_version(void);
 
+// Get compiled libsass language
+ADDAPI const char* ADDCALL libsass_language_version(void);
+
 #ifdef __cplusplus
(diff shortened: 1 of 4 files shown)


mgreter:
I guess this is hardcoded now and we figure out later when and how we gonna update/use it?
It probably should also be documented in https://github.com/sass/libsass/blob/master/docs/api-doc.md
Note: I didn't search for other places it should be added in the documentation (was the first I found).

chriseppstein:
> I guess this is hardcoded now and we figure out later when and how we gonna update/use it?

I mean somewhere you have to write down "this is the version we're implementing". I assume it will be updated when development branch for 3.5 starts. I don't see how it can be data-driven or what value that would provide. Am I understanding correctly?

> It probably should also be documented

I will look around and add docs where it seems warranted.

Issue #2017, opened 2016-04-21
Segfault with @extends and :not


xzyfer:
Originally reported by @dryoma in https://github.com/sass/node-sass/issues/1465

```sass
foo {
  bar: baz;
}

@mixin link() {
  a:not(.btn) {
    color: red;
  }
}

foo {
  @include link;
  @extend .btn;
  @include link;
}
```
Ruby Sass
```css
foo {
  bar: baz; }

foo a:not(.btn):not(foo) {
  color: red; }
foo a:not(.btn):not(foo) {
  color: red; }
```
LibSass 3.3.3+
```
[1]    51362 segmentation fault  /Users/michael/Projects/Sass/sassc/bin/sassc test.scss
```
LibSass 3.3.2
```
foo {
  bar: baz; }

foo a:not(.btn) {
  color: red; }

foo a:not(.btn) {
  color: red; }
```

mgreter:
I get `Internal Error: deque::_M_new_elements_at_front`

mgreter:
The segfault was easy to fix as most of the time (another missing check).
It was triggered by a recent refactor of mine and the results look as "expected".

libsass 3.5.5+
```css
foo {
  bar: baz; }
foo a:not(.btn, foo) {
  color: red; }
foo a:not(.btn, foo) {
  color: red; }
```

libsass 3.3.2 (sassmeister)
```css
foo {
  bar: baz; }
foo a:not(.btn) {
  color: red; }
foo a:not(.btn) {
  color: red; }
```

So it was silently dropping the `:not(foo)` selector, which means it could potentially break the expected css (if compared to ruby sass). Therefore I consider the now invalid css that is output more accurate, since it basically also shows that we now have the right selectors, but not in the correct form yet. And I'm pretty sure that is a duplicate of https://github.com/sass/libsass/issues/1960

mgreter:
Continued in https://github.com/sass/libsass/issues/2031

PR #2022, opened 2016-04-21
Add check in weave to avoid segfault

mgreter:
Fixes https://github.com/sass/libsass/issues/2017 (only the segfault)


mgreter:
So this was also the root cause for https://github.com/sass/node-sass/issues/1464

Issue #2023, opened 2016-04-21
Error when ID selector begins with a number


mahtd:
Regression from libsass 3.3.4

When the ID selector starts with a number, I'm seeing this error.

Input:

```
#1a {
    color: red;
}
```

libsass 3.3.4:

```
$ sassc --version
sassc: 3.3.0
libsass: 3.3.4
sass2scss: 1.0.5
$ sassc test.scss
#1a {
  color: red; }
```

libsass 3.3.5:

```
$ sassc --version
sassc: 3.3.0
libsass: 3.3.5
sass2scss: 1.0.6
$ sassc test.scss
Error: Invalid CSS after "#": expected "{", was "#1a {"
        on line 1 of test.scss
>> #1a {
   ^
```

nex3:
I'm pretty sure the new behavior here is incorrect—according to [the spec](https://drafts.csswg.org/selectors-4/#id-selectors), ID names may only be identifiers.

xzyfer:
That's true, but Sass allows it and "do what says does" has been our guiding principal.

PR #2026, opened 2016-04-21
Relax parsing rules for type selector to start with number

mgreter:
Fixes https://github.com/sass/libsass/issues/2023

### DIFF
--- a/src/prelexer.cpp
+++ b/src/prelexer.cpp
@@ -586,7 +586,7 @@ namespace Sass {
     }
 
     const char* hyphens_and_identifier(const char* src) {
-      return sequence< zero_plus< exactly< '-' > >, identifier >(src);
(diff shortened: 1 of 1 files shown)


xzyfer:
Please add test

mgreter:
Issue is marked accordingly, will do them all in a batch later

mgreter:
@xzyfer by the way how are new spec tests supposed to be added with the new spec test structure? I just saw that we now have multiple folders for `libsass-closed-issues`? //CC @chriseppstein

xzyfer:
My understanding is that we've gone back to one input and the output for the nested syntax. The other libsass-closed-issues folders are for different output styles which are treated separately.

mgreter:
That's how I handled it for the missing spec tests (there are still some open with closed issues, but I should have covered all the simpler once).

Issue #2051, opened 2016-04-26
Bus error: 10 in a mixed global-local extend


ckknight:
`node-sass` version: ~~3.8.3~~ 3.6.0, using included `libsass`
On Mac OS X 10.11
Using either Node.js 4.2 or 5.6
Configured to use local CSS modules by default.
----

Input code:
```scss
:global(.thing) {
    color: red;
}

:global(.thing[disabled]) {
    @extend .thing;
    background: blue;
}
```

Result: `Bus error: 10`

----
Both the following work without issue:
```scss
:global(.thing) {
    color: red;
}

:global(.thing[disabled]) {
    @extend :global(.thing);
    background: blue;
}
```

```scss
.thing {
    color: red;
}

.thing[disabled] {
    @extend .thing;
    background: blue;
}
```
****
Spec added https://github.com/sass/sass-spec/pull/849

xzyfer:
I'm confused, there is no node-sass 3.8.3. Please read their documentation
to determine your correct version.

ckknight:
Sorry about that. It's `3.6.0`. I read the wrong number.

xzyfer:
Thanks for the report @ckknight. You're correct that the first example causes a segfault.

It's also worth noting that neither of your examples are valid Sass. When we patch this segfault both your examples will return the expect error.

```
Error: Invalid CSS after ":global(": expected pseudo_expr, was ".thing)"
        on line 13 of test.scss
  Use --trace for backtrace.
```

xzyfer:
For more information on why this will stop working please read medium.com/@xzyfer/why-node-sass-broke-your-code-and-semver-1b3e409c57b9

If you think this should be supported in Sass please open an issue with the [official Sass project](https://github.com/sass/sass/issues).

mgreter:
This exhibits a tricky one that seems somewhat related to #1392:
It can also be triggered by this valid code:

```scss
:not(.thing) {
    color: red;
}
:not(.thing[disabled]) {
    @extend .thing;
    background: blue;
}
```

ruby sass:
```
:not(.thing) {
  color: red; }
:not(.thing[disabled]):not([disabled]:not(.thing[disabled])) {
  background: blue; }
```

The problem is that libsass tries to extend `.thing` of the wrapped selector itself (and finds another `@extend` inside), so it goes in an endless loop until the call stack is exhausted.

A related test case would therefore be:
```scss
.thing {
    color: red;
}
.thing[disabled] {
    @extend .thing;
[...]

PR #2052, opened 2016-04-27
Fix issue with extend and recursions

mgreter:
Addresses https://github.com/sass/libsass/issues/2051 (and should fix https://github.com/sass/libsass/issues/2054)

I tried to re-use the existing approach with `std::set<Compound_Selector> seen`. But I have no idea how the extend code overall is supposed to work, since the implementation is pretty hairy. I guess this case was simply not implemented. I recently introduced changes to allow extending of wrapped selectors. This lead to a new code path to call `extendSelectorList`, which lead to more exposure of this issue.

used test case:
```scss
:not(.thing) {
    color: red;
}
:not(.thing[disabled]) {
    @extend .thing;
    background: blue;
}
```

now produces:
```css
:not(.thing) {
  color: red; }
:not(.thing[disabled]):not([disabled]:not(.thing[disabled])) {
  background: blue; }
```

So it fixes the segfault and the output is correct.
Specs still pass, but implications for this change are unknown.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -1889,6 +1889,9 @@ namespace Sass {
     virtual void set_media_block(Media_Block* mb) {
       media_block(mb);
     }
+    virtual bool has_wrapped_selector() {
+      return false;
+    }
   };
   inline Selector::~Selector() { }
 
@@ -2185,6 +2188,10 @@ namespace Sass {
       }
       return hash_;
     }
+    virtual bool has_wrapped_selector()
+    {
+      return true;
+    }
     virtual unsigned long specificity()
     {
       return selector_ ? selector_->specificity() : 0;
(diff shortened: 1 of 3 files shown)


mgreter:
This should now also fully pass

```scss
:not(.thing) {
    color: red;
}
:not(.thing[disabled]) {
    @extend .thing;
    background: blue;
}
```

@xzyfer I'm in favor of merging this in lack of any other solution. Sassmeister seems to indicate that this has segfaulted even at node-sass 3.3.6. This patch here at least seems to enable to most common cases with `:not` selector. As I said, I cannot really estimate the impact of the changes, beside trusting our spec tests. If you agree feel free to hit merge.

https://ci.appveyor.com/project/sass/libsass/build/1.0.2426/job/5laspavt0upkreqb#L278
```
The following tests pass but were marked as TODO for libsass:
C:/projects/libsass/sass-spec/spec/libsass-todo-issues/issue_2054
```

xzyfer:
I'll run this through my citgm if all goes well we can ship it. This
probably won't land in time for the next node-sass.

zanderi:
I realize this is an edge case but we ran across it and it took a long time to figure out what was going on. In the end we removed the @extend from the basic styling of the btns and added the properties to each one in order to fix it. But the following code (simplified for demonstration) emits a large amount of selectors that end up breaking the css.

`.btn{
  display:block;
  padding:6px 12px;
}

.btn-basic{
  @extend .btn;
  color:#FFF;
}

.btn-blue{
  @extend .btn-basic;
  color:blue;
}

h1, h2, p, a:not(.btn) {
	color: #FFFFFF;
}

a:not(.btn) {
	text-decoration: underline;
}

a:not(.btn):hover {
	color: #007ad1;
}`

Issue #2042, opened 2016-04-23
Math operations.


dangreen:
Sample
```scss
.wizard-editor {
    transform: scale(50/1);
    transform: scale((50/1));
    transform: scale( (50/1) );
}  
```

libsass 3.5.5
```css
.wizard-editor {
  transform: scale(50/1);
  transform: scale(50/1);
  transform: scale(50/1); }
```

libsass 3.3.2 (sassmeister)
```css
.wizard-editor {
  transform: scale(50);
  transform: scale(50);
  transform: scale(50); }
```

expected
```css
.wizard-editor {
  transform: scale(50/1);
  transform: scale(50);
  transform: scale(50); }
```

mgreter:
This was quite probably changed with https://github.com/sass/libsass/pull/1986. Sorry about that regression, but as we can see this has not been covered by our spec test yet (failing is always a painful and steady way to progress).

Issue #2057, opened 2016-04-29
Regression with call()


lunelson:
The following code was working correctly up until one of the last minor bumps.
```scss
@function apply($args) { @return call($args...); }
@function foo($a, $b, $c){ @return $a $b $c; }

.fntest {
  output: call(foo,1,2,3);
  output: apply(foo 1 2 3);
}
```
Returned correctly, as late as `v3.3.4`:
```css
.fntest {
  output: 1 2 3;
  output: 1 2 3; }
```
now with `v3.3.6` returns:
```css
.fntest {
  output: 1 2 3;
  output: foo 1 2 3...(); }
```

mgreter:
Probably a duplicate of https://github.com/sass/libsass/issues/2034

Issue #2034, opened 2016-04-22
Regression: Call function does not destructure variables


TxHawks:
When doing:
```scss
$args: mix, #fff, #000, 20%;

.foo {
  color: call($args...);
}
```

Ruby Sass will output:
```css
.foo {
  color: #333333; 
}
```

But Libsass will output:
```css
.foo {
  color: mix, #fff, #000, 20%...(); 
}
```

This is happening with: `node-sass 3.5.3` wrapping `libsass 3.3.5`.

mgreter:
This is because an early check was added in function call evaluation, since it was needed to avoid the evaluation of the arguments for `calc` functions (`call(calc, ...)`). The normal function evaluation has a very messed up logic to deduct the arguments for the ast node (`Argument_List`, restargs, etc). It probably makes sense to refactor the whole list handling including https://github.com/sass/libsass/issues/1930#issuecomment-213189134, instead of reimplementig the whole argument deduction logic for call invokations.

xzyfer:
I refactored the argument logic to match Ruby Sass in 3.3.3. It should be correct. The issue is more likely which how we're evaling `call`.

mgreter:
Bug happens here: https://github.com/sass/libsass/blob/master/src/eval.cpp#L842

mgreter:
This is a more complete spec test:
```scss
$args: mix, #fff, #000, 20%;
$params: #fff, #000, 20%;

.foo {
  test: call($args...);
  test: mix(#fff, #000, 20%);
  test: call(mix, #fff, #000, 20%);
  test: call(mix, $params...);
}
```

Currently yields:
```
.foo {
  test: mix, #fff, #000, 20%...();
  test: #333333;
  test: #333333;
  test: #333333; }
```

Expected is:
```css
.foo {
  test: #333333;
  test: #333333;
  test: #333333;
  test: #333333; }
```

So we can probably get away here with one additional check. @xzyfer I also do not like to add such patches, but I see it as a legit way to add the needed behavior until someone (mostly me) gets around to do the actual refactoring. And it has proven to help to have these patches in place, specially if they are documented accordingly. The only small thing that worries me a little is that this pattern may break lazy evaluation as fixed in https://github.com/sass/libsass/pull/1986. To put it short we tend to get away with those patches quite far (and I do keep cleaning up after them).

PR #2059, opened 2016-04-30
Fix and refactor delayed evaluation flag

mgreter:
Fixes https://github.com/sass/libsass/issues/2042, https://github.com/sass/libsass/issues/2034 and https://github.com/sass/libsass/issues/2057 (Spec tests https://github.com/sass/sass-spec/pull/856).

This is pretty much a complete rewrite of how we handle delayed values:

```c++
// Notes about delayed: some ast nodes can have delayed evaluation so
// they can preserve their original semantics if needed. This is most
// prominently exhibited by the division operation, since it is not
// only a valid operation, but also a valid css statement (i.e. for
// fonts, as in `16px/24px`). When parsing lists and expression we
// unwrap single items from lists and other operations. A nested list
// must not be delayed, only the items of the first level sometimes
// are delayed (as with argument lists). To achieve this we need to
// pass status to the list parser, so this can be set correctly.
// Another case with delayed values are colors. In compressed mode
// only processed values get compressed (other are left as written).
```

Issue #2042 was never fully working, but its behavior was changed with #1986 from undelayed to delayed. From my experience libsass struggled to emulate these behaviors from the beginning. First I figured out why #2042 was never working. It turns out that we have to only set delayed on certain parsed ast nodes. And whoever calls parse_list must tell it if values on the first level must be delayed (//CC @xzyfer). This is the case for declarations and argument lists (and others), but not for nested lists (with exceptions).

Once implemented pretty much all other stuff related to delayed values broke. So I set out to fix them one by one, and most were easy. The trickiest one revealed itself in the following sass:

```scss
.foo {
  eval: foobar(true, ((yellow 3) red), 1/2);
}  
```

This compiles to (in compressed :exclamation:  mode):
```css
.foo{eval:foobar(true, #ff0 3 red, 1/2)}
```

It shows that undelaying colors vs divisions are really two different things. Because in the sample above the color got undelayed, while the division did not. Another key code in the PR is that binary expression never get delayed when they contain nested binary expressions. With this logic in place I was able to remove most of the complicated code regarding delayed value handling.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -74,6 +74,21 @@ namespace Sass {
     ltrim();
   }
 
+  void Argument::set_delayed(bool delayed)
+  {
+    if (value_) value_->set_delayed(delayed);
+    is_delayed(delayed);
+  }
+
+  void Arguments::set_delayed(bool delayed)
+  {
+    for (Argument* arg : elements()) {
+      if (arg) arg->set_delayed(delayed);
+    }
+    is_delayed(delayed);
+  }
+
+
   bool At_Root_Query::exclude(std::string str)
   {
     bool with = feature() && unquote(feature()->to_string()).compare("with") == 0;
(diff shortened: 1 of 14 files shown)


drewwells:
👍

PR #2060, opened 2016-04-30
Split AST source into multiple compile units

mgreter:
This splits the massive AST source files into

- nodes (base)
- containers
- blocks (has_blocks)
- expressions
- statements
- values
- selectors

This is only a first step and we should probably do the same for some other massive source files (parser, eval and function). This also helps to keep memory requirements lower during compiling, although I'm not sure if we loose any noticeable performance from weaker optimization (related to the discussion about unified builds). I haven't put too much thought into which includes to put where, was more interested to get it to compile on all platforms. From a code maintainance point this is definitely a step in the right direction, so if CI goes green I would like to ship this (//CC @xzyfer).


mgreter:
Seems to not some work on the autotools side. Have to test localy.

xzyfer:
I need time to look over this. Going to bed now. Will make time tomorrow.

mgreter:
Note: with autotools 2.69 we get a new warning:
```
automake-1.15: warning: possible forward-incompatibility.
automake-1.15: At least a source file is in a subdirectory, but the 'subdir-objects'
automake-1.15: automake option hasn't been enabled.  For now, the corresponding output
automake-1.15: object file(s) will be placed in the top-level directory.  However,
automake-1.15: this behaviour will change in future Automake versions: they will
automake-1.15: unconditionally cause object files to be placed in the same subdirectory
automake-1.15: of the corresponding sources.
automake-1.15: You are advised to start using 'subdir-objects' option throughout your
automake-1.15: project, to avoid future incompatibilities.
```

This happens also on the old code. Unfortunately building sassc/tester fails if I enable `subdir-objects` also for the main makefile. For now I only enabled it on the src directory, since it is needed to build correctly for this patch as it makes sense to use a folder structure when spliting up big source chunks.

I also looked at another warning, which seems to be a bug with autotools and GNUMakefile.am:
http://www.archivum.info/bug-libtool@gnu.org/2008-11/00081/Re-libtool-version-2.2.6-only-checks-Makefile.am-for-ACLOCAL_AMFLAGS.html

src/ast/blocks.cpp
```diff
@@ -0,0 +1,33 @@
+#include "../sass.hpp"
+#include "values.hpp"
```
xzyfer: Should this be `blocks.hpp`?

src/ast/expressions.cpp
```diff
@@ -0,0 +1,139 @@
+#include "../sass.hpp"
+#include "values.hpp"
```
xzyfer: ditto?

src/ast/nodes.cpp
```diff
@@ -0,0 +1,40 @@
+#include "../sass.hpp"
+#include "values.hpp"
```
xzyfer: ditto

src/ast/statements.cpp
```diff
@@ -0,0 +1,19 @@
+#include "../sass.hpp"
+#include "values.hpp"
```
xzyfer: ditto

xzyfer:
All in all :+1:
Not sold on `containers` but we change easily change it later if it's annoying enough.

src/ast/nodes.hpp
```diff
@@ -0,0 +1,194 @@
...
+  inline Statement::~Statement() { }
+}
+
+
+
+/////////////////////////////////////////////////////////////////////////////////////
+// Hash method specializations for std::unordered_map to work with Sass::Expression
+/////////////////////////////////////////////////////////////////////////////////////
+
+namespace std {
```
xzyfer: I think this belongs in `containers`. It's critical to understanding `Hashed`. Not entire sure how that'll work given it's a fwd decl.

mgreter:
As I said, haven't put much thought into the header includes.
For now I just made it work, so I agree this should be refined further.

xzyfer:
Yep just leaving some thoughts here so that they're somewhere.

Issue #1732, opened 2015-11-13
Including a `@mixin` when outside of selector still outputs properties


ezekg:
As mentioned in https://github.com/ezekg/flint/issues/40#issuecomment-156465316 and similar to https://github.com/sass/libsass/issues/1585.

### Reproducable test case:
```scss
@mixin a {
  b: c;
}

@include a();
```

### Compiling with Ruby Sass results in:

```
Error: Properties are only allowed within rules, directives, mixin includes, or other properties.
```

Cmd: `sass test.scss` 

### Compiling with Libsass results in:

```css
b: c;
```

Cmd: `sassc test.scss`

xzyfer:
Thanks for the report @ezekg.

xzyfer:
Spec added https://github.com/sass/sass-spec/pull/613

mgreter:
This is a bit more complicated since the check cannot be done in parser, since the following is valid:

```scss
@mixin a {
  b: c;
}
foo {
  @include a();
}
```

So we need to do this check after the mixin is expanded.

mgreter:
WIP to fix this in https://github.com/sass/libsass/pull/1900

xzyfer:
Reassiging this to 3.4. This doesn't prevent people to from migrating to LibSass I suspect browsers will simply ignore these erroneous properties.

PR #2061, opened 2016-04-30
Error when declarations are printed without ruleset

mgreter:
Fixes https://github.com/sass/libsass/issues/1732 - Another one of the category "should error" :exclamation: 

`Properties are only allowed within rules, directives, mixin includes, or other properties.`

### DIFF
--- a/src/cssize.cpp
+++ b/src/cssize.cpp
@@ -493,6 +493,15 @@ namespace Sass {
     return flatten(result);
   }
 
+  Statement* Cssize::operator()(Declaration* d)
+  {
+    if (block_stack.size() < 2) {
+      throw Exception::InvalidSass(d->pstate(), "Properties are only allowed "
+        "within rules, directives, mixin includes, or other properties.");
+    }
+    return static_cast<Statement*>(d);
+  }
+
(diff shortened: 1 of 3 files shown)


xzyfer:
Let's not do this. This check does not belong in cssize. The cssize class is a 1-1 port from Ruby, no need to change that.

Ruby solves this a `CheckNesting` visitor. We an equivalent since there is an entire class of nesting errors we cannot catch properly atm.

mgreter:
IMO this can be moved over once such a `CheckNesting` visitor exists. I don't see a reason why this should not be correct, or can you come up with a non failing test?

xzyfer:
This bug doesn't prevent Sass code being ported to LibSass so IMHO it's a high priority. IMO a fix for this bug is not worth the tech debt. 

If you really want to fix this it's trivial to add a `CheckNesting` visitor [before cssize runs](https://github.com/sass/sass/blob/c7886e4ce87c87d1778fa89a1ff653a08595b3f9/lib/sass/tree/root_node.rb#L37).

mgreter:
Superseeded by https://github.com/sass/libsass/pull/2062

PR #2062, opened 2016-05-01
Error when declarations are printed without ruleset

### LINKED ISSUE #1732: Including a `@mixin` when outside of selector still outputs properties

xzyfer:
This is introduces a vastly reduced `CheckNesting` visitor from Ruby
Sass as discussed in #2061. Doing this properly will require some
small changes to the parser, and probably eval, which currently try
to do some nesting checking of their own.

This is just first step to properly introducing the proper nesting
checking.

Spec sass/sass-spec#857
Closes #2061
Fixes #1732

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -1,0 +1,66 @@
+#include "sass.hpp"
+#include <vector>
+
+#include "check_nesting.hpp"
+#include "context.hpp"
+// #include "debugger.hpp"
+
+namespace Sass {
+
+  CheckNesting::CheckNesting(Context& ctx)
+  : ctx(ctx),
+    parent_stack(std::vector<AST_Node*>())
+  { }
+
+  AST_Node* CheckNesting::parent()
+  {
+    if (parent_stack.size() > 0)
+      return parent_stack.back();
+    return 0;
+  }
+
+  Statement* CheckNesting::operator()(Block* b)
+  {
+    parent_stack.push_back(b);
+    append_block(b);
+    parent_stack.pop_back();
+    return b;
+  }
+
+  Statement* CheckNesting::operator()(Declaration* d)
+  {
+    if (!is_valid_prop_parent(parent())) {
+      throw Exception::InvalidSass(d->pstate(), "Properties are only allowed "
+        "within rules, directives, mixin includes, or other properties.");
+    }
+    return static_cast<Statement*>(d);
+  }
+
+  Statement* CheckNesting::fallback_impl(AST_Node* n)
+  {
+    return static_cast<Statement*>(n);
+  }
+
+  bool CheckNesting::is_valid_prop_parent(AST_Node* p) 
+  {
+    if (Definition* def = dynamic_cast<Definition*>(p)) {
+      return def->type() == Definition::MIXIN;
+    }
+
+    return dynamic_cast<Ruleset*>(p) ||
+           dynamic_cast<Keyframe_Rule*>(p) ||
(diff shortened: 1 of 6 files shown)


src/check_nesting.cpp
```diff
@@ -0,0 +1,53 @@
...
+    Block* bb = SASS_MEMORY_NEW(ctx.mem, Block, b->pstate(), b->length(), b->is_root());
+    block_stack.push_back(bb);
+    append_block(b);
+    block_stack.pop_back();
+    return bb;
+  }
+
+  Statement* CheckNesting::operator()(Declaration* d)
+  {
+    if (block_stack.size() < 2) {
```
xzyfer: This check [still needs work](https://github.com/sass/sass/blob/63586f1e51c0aec47a7afcc8cd17aa03f32f0a29/lib/sass/tree/visitors/check_nesting.rb#L140-L141) and the test case will need to be updated.

src/check_nesting.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace Sass {
+
+  CheckNesting::CheckNesting(Context& ctx)
+  : ctx(ctx),
+    block_stack(std::vector<Block*>())
+  { }
+
+  Statement* CheckNesting::operator()(Block* b)
+  {
+    Block* bb = SASS_MEMORY_NEW(ctx.mem, Block, b->pstate(), b->length(), b->is_root());
```
mgreter: Is it really necessary to make another copy? I figure this should not be needed since we do not really alter anything here? Creating too many clones is the main reason why libsass is slower than it could be (I already removed some cloning in eval some time ago, which lead to the big performance gain). The other reason we are slower than needed is that we visit too many visitors, but I guess there is not much too avoid that ...

mgreter:
Other than my comments, LGTM!

xzyfer:
Beefed out the specs in https://github.com/sass/sass-spec/pull/857

src/check_nesting.cpp
xzyfer: No we don't need the copy. I slimmed this down a lot, fixed the check, and beefed out the spec. This visitor should never mutate the nodes.

mgreter: IMHO if it should never mutate than the `append_block` is also superflous. Doesn't it always replace the children with itself anyway under that condition? Otherwise good work!

xzyfer: Yep I've also removed that locally. Just pushed my latest. Tracking down an error mismatch atm.

mgreter: :+1:

Issue #1585, opened 2015-10-05
Passing `@content` to `@at-root` allows properties outside of rules


lunelson:
Passing a `@content` block to an `@at-root` rule allows properties to be written at the root without being inside any rule or directive:
```scss
@mixin bar() {
  @at-root { @content; }
}

.test {
  @include bar() {
    color: yellow;
    .nest2 {
      color: green;
    }
  }
}
```
expected:
```bash
Properties are only allowed within rules, directives, mixin includes, or other properties.
```
actual:
```css
color: yellow;
.test .nest2 {
  color: green; }
```

xzyfer:
I'm still curious about this. I'll take a look at it separately from #2064.

xzyfer:
I have a WIP for that may also solve #2006.

xzyfer:
I'm pretty close on this. Should have something soon.

xzyfer:
Made a lot of progress on this. Down to a couple failings due us incorrectly handling propsets.

mgreter:
@xzyfer can you post a WIP branch for this?

xzyfer:
@mgreter I'm at work atm but I'll be shipping this over a handful of PRs this week. It required touching more code than I had expected.

xzyfer:
@mgreter WIP PR in #2086. It's super messy but works as expected is like for like port from Ruby Sass.

PR #2064, opened 2016-05-02
Clean up some dead code in check nesting

xzyfer:
- fix warning about unused `ctx`
- remove commented out code
- don't reassign root to the check nesting visitor

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -2,14 +2,11 @@
 #include <vector>
 
 #include "check_nesting.hpp"
-#include "context.hpp"
-// #include "debugger.hpp"
 
 namespace Sass {
 
-  CheckNesting::CheckNesting(Context& ctx)
-  : ctx(ctx),
-    parent_stack(std::vector<AST_Node*>())
+  CheckNesting::CheckNesting()
+  : parent_stack(std::vector<AST_Node*>())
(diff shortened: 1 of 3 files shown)


mgreter:
LGTM!

mgreter:
This would probably also fix https://github.com/sass/libsass/issues/1585 if the check is added after the extend phase. I wonder why this check is called after each phase, shouldn't one at the end be enough?
```diff
      Extend extend(*this, subset_map);
      // extend tree nodes
      root->perform(&extend);
    }
+    // check proper nesting of ast nodes
+    root = root->perform(&check_nesting)->block();

    // clean up by removing empty placeholders
    // ToDo: maybe we can do this somewhere else?
```

xzyfer:
Ruby Sass only checks nesting [before and after eval](https://github.com/sass/sass/blob/c7886e4ce87c87d1778fa89a1ff653a08595b3f9/lib/sass/tree/root_node.rb#L35-L37). I expect this to prevent unnecessary duplication and guarding in the parser and eval.

You're correct adding check after extend would fix #1585 but Ruby Sass doesn't do it. That issue is indicative of a bigger issue. Since it's an edge case and doesn't prevent people adopting LibSass it's not worth introducing technical debt for.

PR #2066, opened 2016-05-02
Fix minor issue with handling "\a" in interpolations

mgreter:
Fixes https://github.com/sass/libsass/issues/1786

The check currently needed in interpolation indicates that we probably should not get a list
there in the first place. Normally single list items are not wrapped in a list container.


xzyfer:
lgtm

PR #2067, opened 2016-05-02
Fix error message for missing arguments

mgreter:
Updates one error message from bind class to conform to ruby sass:

```
Error: Mixin colors is missing argument $color.
        on line 7 of foo.scss
```

Interestingly we already have a case for this in todo-issues/83


xzyfer:
Spec activated in https://github.com/sass/sass-spec/pull/869

Issue #1930, opened 2016-02-25
Support maps in list context (index, nth, etc)


madlandproject:
Example file, sasstest.scss : 

```scss
$test-map : (
        mykey    : 10px,
        otherKey : 20px,
        moarKey  : 30px
);


@mixin testMixin($key-name) {

  // get index in map
  $item-value : map-get($test-map, $key-name);
  $item-index : index($test-map, ($key-name $item-value) );

  @debug $item-index;

}

@include testMixin(otherKey);
```

Expected output (using ruby sass 3.4.21) :
sasstest.scss:14 DEBUG: 2

Output with libsass : 
sasstest.scss:14 DEBUG: null

mgreter:
I do see the difference, but wonder if this is actually indended behavior. IMHO this is quite a "funky" feature so to speak (never seen any similar "index" function with "hash-maps") //CC @chriseppstein 

```scss
$test-map : (
  key : 20px,
);

foo {
  bar: index($test-map, ("key" 20px) );
  baz: index($test-map, ("key", 20px) );
  ibar: inspect(index($test-map, ("key" 20px) )); // 1
  ibaz: inspect(index($test-map, ("key", 20px) )); // null
}
```

Yields in ruby sass:
```
foo {
  bar: 1;
  ibar: 1;
  ibaz: null; }
```

xzyfer:
For context, when a map is used with index (instead of a list) it's treated a list of lists. i.e. `(foo: bar, bar: baz)`, is `(foo bar, bar baz)`.

This applies for all list functions. This is documented in the Sass changelog for 3.3.

>All the existing list functions also work on maps, treating them as lists of pairs. For example, nth((foo: 1, bar: 2), 1) returns foo 1. Maps can also be used with @each, using the new multiple assignment feature (see below):
>[source](http://sass-lang.com/documentation/file.SASS_CHANGELOG.html#sassscript_maps)

mgreter:
Would be great if someone could come up with a complete list, or better spec test suite for all related functions, since this doesn't seem to be related to index only (I don't think `nth` has that support)!?

mgreter:
On another note this should probably be connected to a refactoring with selector lists, since all three types seem to share some common properties (iterable etc). Would probably help we had this covered in AST directly and not having to implement the logic again in each function (nth, inspect, etc).

chriseppstein:
Sass maps are ordered. As such, the pairs within them have an intrinsic index based on insertion order.

[...]

PR #2068, opened 2016-05-02
Implement indexed map access (incubating feature)

mgreter:
Fixes https://github.com/sass/libsass/issues/1930

Was even simpler than anticipated. Had to add a `deque` object to the hash map implementation. I also tried to switch to a `std::map`, but it doesn't seem to support a custom hasher and only works with compare operator (which is not good, since we cannot really compare different types of ast nodes). It wouldn't had much benefit anyway (we would only get an ordered iterator over all pairs). As always with hash maps, it more of a guess which implementation would lead to the best performance.

IMO this implementation is pretty close to being optimal (in terms of performance). Random/Indexed access is constant `O(1)` due to the use of `deque`. I didn't put too much thought into the API, but it should be straight forward to use. You can fetch key or value by index. As I already wrote in other threads, the list/map interfaces/types/classes should be refactored and unified at some point (basically hiding all the logic currently implemented in functions.cpp and elsewhere).

If you can point me to other functions that allow random/indexed access to maps, please add them in this thread. Will see if I can also implement them. And yes, this is still missing proper spec test coverage.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -303,6 +303,7 @@ namespace Sass {
     std::vector<Expression*> list_;
   protected:
     size_t hash_;
+    std::deque<Expression*> ordered;
     Expression* duplicate_key_;
     void reset_hash() { hash_ = 0; }
     void reset_duplicate_key() { duplicate_key_ = 0; }
@@ -321,14 +322,23 @@ namespace Sass {
     Hashed& operator<<(std::pair<Expression*, Expression*> p)
     {
       reset_hash();
-
+      ordered.push_back(p.first);
+      ordered.push_back(p.second);
       if (!has(p.first)) list_.push_back(p.first);
       else if (!duplicate_key_) duplicate_key_ = p.first;
-
       elements_[p.first] = p.second;
-
       adjust_after_pushing(p);
       return *this;
+    }
+    Expression* key(size_t i) {
+      return ordered.at(i * 2);
+    }
+    Expression* value(size_t i) {
+      return ordered.at(i * 2 + 1);
+    }
+    Expression* value(size_t i, Expression* v) {
+      ordered[i * 2 + 1] = v;
+      return v;
     }
     Hashed& operator+=(Hashed* h)
     {
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -1248,6 +1248,10 @@ namespace Sass {
         if (index < 0 || index > len - 1) error("index out of bounds for `" + std::string(sig) + "`", pstate);
         // return (*sl)[static_cast<int>(index)];
         Listize listize(ctx.mem);
+        if (m) {
+          List* l = SASS_MEMORY_NEW(ctx.mem, List, pstate, 2);
+          *l << m->key(index) << m->value(index);
+        }
         return (*sl)[static_cast<int>(index)]->perform(&listize);
       }
       List* l = dynamic_cast<List*>(env["$list"]);
@@ -1265,8 +1269,8 @@ namespace Sass {
 
       if (m) {
         l = SASS_MEMORY_NEW(ctx.mem, List, pstate, 1);
-        *l << m->keys()[static_cast<unsigned int>(index)];
-        *l << m->at(m->keys()[static_cast<unsigned int>(index)]);
+        *l << m->key(index);
+        *l << m->value(index);
         return l;
       }
       else {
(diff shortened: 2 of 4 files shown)


xzyfer:
> it doesn't seem to support a custom hasher

This is why unordered map is used. We maintainer the ordering by keeping the keys ordered in a vector on insert.

xzyfer:
Can you elaborate why you need the `ordered` vector? `Hashed` is already ordered. The has itself is an unordered list but the insert order of the keys is tracked.

It looks like `Hashed`'s iterator has been broken. It should iterate over the `list_` vector and look up the value in the `elements_` hash.

src/ast.hpp
```diff
@@ -321,15 +322,24 @@ namespace Sass {
...
       if (!has(p.first)) list_.push_back(p.first);
       else if (!duplicate_key_) duplicate_key_ = p.first;
-
       elements_[p.first] = p.second;
-
       adjust_after_pushing(p);
       return *this;
     }
+    Expression* key(size_t i) {
+      return ordered.at(i * 2);
```
xzyfer: `keys().at(i)`

src/ast.hpp
```diff
@@ -321,15 +322,24 @@ namespace Sass {
...
+    }
+    Expression* value(size_t i) {
+      return ordered.at(i * 2 + 1);
```
xzyfer: `elements().at(keys(i))`

src/functions.cpp
```diff
@@ -1265,8 +1269,8 @@ namespace Sass {
 
       if (m) {
         l = SASS_MEMORY_NEW(ctx.mem, List, pstate, 1);
-        *l << m->keys()[static_cast<unsigned int>(index)];
```
xzyfer: Why does this need to change?

mgreter:
@xzyfer Can you elaborate where "we maintain the ordering by keeping the keys ordered in a vector on insert". I did not find any vector where we "keep" the insertion order, that is exactly why I have added "deque".

xzyfer:
FWIW I don't get email updates when editing comments. It's worth just posting a new comment so I'm not email back out of context replies. Timezones 😞

mgreter:
I also added the values to the "vector", since it should make it faster to iterate over values then by fetching each key from the hash map itself. It may also be more effecient to use deque instead of a vector. We can also easily implement optimal iterators over values and keys. To access keys or values you just have to multiply the index by two (and adding 1 for values).

xzyfer:
The extra memory doesn't seem worth it. Is there a use case for iterating over values?

xzyfer:
The overwhelming majority of map usage in Sass is direct key look up. We should optimise for that use case above all else.

mgreter:
I agree that direct key look up is the most important one, but I would like to guarantee that it doesn't degrade too much. But I need to make some performance assessments first, to determine if this is the case with repeated key access vs. random access via deque. IMO the memory overhead is very little compared to the actual data that is stored (4 to 8 additional bytes per item, so 1MB should fit 128k items on x64 arch). We do not have many places where this is currently used. On a first glimpse it seems to only be the case in map destructors and we could probably refactor that to use the iterator (no need to destroy in insert order).

On the other hand there are the C-API functions that basically have to traverse a map in this way if they need to do so. It seems that i.e. `sass_map_get_value` etc. are broken anyway currently, since it returns `v->map.pairs[i].value` (`i` of type `size_t`). It looks the performance implications for access by key are "Average case O(1), worst case O(size()).". I figure that depends on how well our hash function avoids collisions. On the other hand each access via keys has to go through the hashing function. So there is always a constant overhead for that. But as I said, not sure how that translates to real world scenarios, so again not sure if the memory/cpu tradeoff is worth it.

But I think it probably makes sense to use "deque" instead of vector in "Hashed", as appending stuff is cheaper for deque and random index access is also O(1). There are always [pros and cons] [1].

[1]: http://www.gotw.ca/gotw/054.htm

xzyfer:
I'm 👍  for make `list_` a `deque`

xzyfer:
The tradeoff in code complexity is also something to consider. I'll take slight less efficient, user friendly, maintainable code instead of overly complicated, over optimized code any day.

xzyfer:
With that said adopting deque isn't necessary to fix the issue at hand. I'd
avoid conflating the two is this PR.
On 3 May 2016 9:38 PM, "Marcel Greter" <notifications@github.com> wrote:

I agree that direct key look up is the most important one, but I would like
to guarantee that it doesn't degrade too much. But I need to make some
performance assessments first, to determine if this is the case with
repeated key access vs. random access via deque. IMO the memory overhead is
very little compared to the actual data that is stored (4 to 8 additional
bytes per item, so 1MB should fit 128k items on x64 arch). We do not have
many places where this is currently used. On a first glimpse it seems to
only be the case in map destructors and we could probably refactor that to
use the iterator (no need to destroy in insert order).

On the other hand there are the C-API functions that basically have to
traverse a map in this way if they need to do so. It seems that i.e.
sass_map_get_value etc. are broken anyway currently, since it returns
v->map.pairs[i].value (i of type size_t). It looks the performance
implications for access by key are "Average case O(1), worst case
O(size()).". I figure that depends on how well our hash function avoids
collisions. On the other hand each access via keys has to go through the
hashing function. So there is always a constant overhead for that. But as I
said, not sure how that translates to real world scenarios, so again not
sure if the memory/cpu tradeoff is worth it.

But I think it probably makes sense to use "deque" instead of vector in
"Hashed", as appending stuff is cheaper for deque and random index access
is also O(1). There are always pros and cons
<http://www.gotw.ca/gotw/054.htm>.

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/sass/libsass/pull/2068#issuecomment-216502581>

PR #2077, opened 2016-05-07
Remove placeholder for Sass lang version

am11:
Based on the commentary in PR #2021, there is no need to define
a placeholder for language version in `version.h.in` file. Instead we
define the constant in `version.h` for non-gcc compilers.

With this change MSVC and clang compiled outputs will not show [NA]
anymore.

### DIFF
--- a/include/sass/version.h
+++ b/include/sass/version.h
@@ -6,7 +6,7 @@
 #endif
 
 #ifndef LIBSASS_LANGUAGE_VERSION
-#define LIBSASS_LANGUAGE_VERSION "[NA]"
+#define LIBSASS_LANGUAGE_VERSION "3.4"
 #endif
 
 #endif


am11:
cc @mgreter

mgreter:
LGTM, but as I wrote earlier, It doesn't serve much use for libsass ATM, until we start to actually implement different language versions. Although we are pretty close to sass 3.4, the underyling code is really not ready to move on to 3.5 atm. (let aside suporting multiple versions of the language) There are just too many places where implementiation differs from sass implementation. We do try to get these in line with ruby sass, but here are a lot of subtile differences where I personaly don't know how to implement the behavior 100% correctly in libsass (even after looking at the sass/ruby code). So for these cases we either need someone who knows the sass specs inside out and can provide fixes for the C++ code. But for now I think we'll be stuck with sass 3.4 for quite a long time ... so LGTM!

xzyfer:
@mgreter feel free to ask me in these cases. I know the language spec very well and know my way around the Ruby code base.

xzyfer:
These failures are strange.

xzyfer:
Rebuilding the failed build to see if it's a blip.

xzyfer:
CI looks better \o/

mgreter:
@xzyfer https://github.com/sass/libsass/issues/1392 for example ... I don't know why this is different than in ruby sass ...

PR #2078, opened 2016-05-08
Export memory allocator functions to public API

am11:
There functions were missing __cdecl (required by FFI consumption).

cc @mgreter


mgreter:
Seems I messed this up when updating the documentation pages. Thanks @am11

PR #2085, opened 2016-05-18
[DNM] Consolidate the Declaration and Propset AST nodes

### LINKED ISSUE #1585: Passing `@content` to `@at-root` allows properties outside of rules

xzyfer:
In order to fix #1585 we need to introduce the `Trace` node from
Ruby Sass. This node interacts with propsets in an interesting way.
Have two node to represent the single Ruby `Prop` node caused a
lot of trouble with correct implementing check nesting logic for
`Trace`.

The trouble with implementing `Trace` were exaggerated by us
incorrectly eval propsets into declarations in the expand step.
This PR moves the de-nesting of prop nodes the cssize step to
[better match Ruby Sass][1].

Later we should rename this node `Prop` but it's out of scope of
this PR. This should allow my work on porting the `Trace` node over
from Ruby Sass to fix #1585.

[1]: https://github.com/sass/sass/blob/63586f1e51c0aec47a7afcc8cd17aa03f32f0a29/lib/sass/tree/visitors/cssize.rb#L149-L166

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -481,18 +481,6 @@ namespace Sass {
     ATTACH_OPERATIONS()
   };
 
-  /////////////////////////////////////////////////////////
-  // Nested declaration sets (i.e., namespaced properties).
-  /////////////////////////////////////////////////////////
-  class Propset : public Has_Block {
-    ADD_PROPERTY(String*, property_fragment)
-  public:
-    Propset(ParserState pstate, String* pf, Block* b = 0)
-    : Has_Block(pstate, b), property_fragment_(pf)
-    { }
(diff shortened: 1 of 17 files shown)


xzyfer:
This currently has tests that fail due to an incomplete implementation of the check nesting visitor which is resolved in #2086.

PR #2086, opened 2016-05-18
[DNM] Implement the Trace node

### LINKED ISSUE #1585: Passing `@content` to `@at-root` allows properties outside of rules

xzyfer:
The Trace node is a simple wrapper around the result of a mixin call 
and `@content`. It exists as a way to group the resulting properties
as a block but maintain some metadata about the initiating mixin call 
or `@content` block.

It also plays an important rule is bubble those blocks with `@at-root` 
and nested property sets. This required implementing the final
 missing `cssize` methods. As a result we should a complete `@at-root` 
implementation.

`Trace` "wrapping" other nodes didn't play nice with the naive check
nesting implementation I lazily wrote so I also implemented 99% of the 
check nesting visitor like-for-like with Ruby Sass. The missing piece is 
handling `@imports` which is the only part I didn't need to touch in this 
work.

Spec sass/sass-spec#868
Fixes #1585 

## TODO
- [ ] blocked on #2085
- [x] enable specs (sass/sass-spec#868)

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -6,52 +6,374 @@
 namespace Sass {
 
   CheckNesting::CheckNesting()
-  : parent_stack(std::vector<AST_Node*>())
+  : parents(std::vector<Statement*>()),
+    parent(0),
+    current_mixin_definition(0)
   { }
 
-  AST_Node* CheckNesting::parent()
-  {
-    if (parent_stack.size() > 0)
-      return parent_stack.back();
-    return 0;
+  Statement* CheckNesting::before(Statement* s) {
+      if (this->should_visit(s)) return s;
+      return 0;
   }
 
-  Statement* CheckNesting::operator()(Block* b)
-  {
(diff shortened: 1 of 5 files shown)


xzyfer:
@mgreter this PR currently has 10 failings test that require #2085. Unfortunately #2086 has 9 failing test that require this PR.

My plan to clean up the two PRs in a clean state for review, then create a new PR that merges the two that will be merged.

xzyfer:
This PR currently has 2 failings test, 1 is fixed by #2085, which currently has it's own failing test which depends on this PR.

PR #2089, opened 2016-05-19
Implement the Trace AST node

### LINKED ISSUE #1585: Passing `@content` to `@at-root` allows properties outside of rules

xzyfer:
This PR is the combination of #2085 and #2086. 
They have their own detailed descriptions I wont repeat here.

Unfortunately those two PRs each have a single failing spec
which is due to a missing implementation provided by the
alternative PR. Those PRs have been marked as DNM and
will closed by this PR. I'm doing this so that master stays green
in case a revert is required.

Spec sass/sass-spec#868
Closes #2085 
Closes #2086
Fixes #1585

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -42,7 +42,7 @@ namespace Sass {
     return static_cast<Statement*>(n);
   }
 
-  bool CheckNesting::is_valid_prop_parent(AST_Node* p) 
+  bool CheckNesting::is_valid_prop_parent(AST_Node* p)
   {
     if (Definition* def = dynamic_cast<Definition*>(p)) {
       return def->type() == Definition::MIXIN;
@@ -50,7 +50,7 @@ namespace Sass {
 
     return dynamic_cast<Ruleset*>(p) ||
            dynamic_cast<Keyframe_Rule*>(p) ||
-           dynamic_cast<Propset*>(p) ||
+           dynamic_cast<Declaration*>(p) ||
            dynamic_cast<Directive*>(p) ||
            dynamic_cast<Mixin_Call*>(p);
   }
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -6,52 +6,374 @@
 namespace Sass {
 
   CheckNesting::CheckNesting()
-  : parent_stack(std::vector<AST_Node*>())
+  : parents(std::vector<Statement*>()),
+    parent(0),
+    current_mixin_definition(0)
   { }
 
-  AST_Node* CheckNesting::parent()
-  {
-    if (parent_stack.size() > 0)
-      return parent_stack.back();
-    return 0;
+  Statement* CheckNesting::before(Statement* s) {
+      if (this->should_visit(s)) return s;
+      return 0;
   }
 
-  Statement* CheckNesting::operator()(Block* b)
-  {
-    parent_stack.push_back(b);
+  Statement* CheckNesting::visit_children(Statement* parent) {
 
-    for (auto n : *b) {
-      n->perform(this);
+    Statement* old_parent = this->parent;
+
+    if (dynamic_cast<At_Root_Block*>(parent)) {
+      std::vector<Statement*> old_parents = this->parents;
+      std::vector<Statement*> new_parents;
+
+      for (size_t i = 0, L = this->parents.size(); i < L; i++) {
+        Statement* p = this->parents.at(i);
+        if (!dynamic_cast<At_Root_Block*>(parent)->exclude_node(p)) {
+          new_parents.push_back(p);
+        }
+      }
+      this->parents = new_parents;
+
+      for (size_t i = this->parents.size(); i > 0; i--) {
+        Statement* p = 0;
+        Statement* gp = 0;
+        if (i > 0) p = this->parents.at(i - 1);
+        if (i > 1) gp = this->parents.at(i - 2);
+
+        if (!this->is_transparent_parent(p, gp)) {
+          this->parent = p;
+          break;
+        }
+      }
+
+      At_Root_Block* ar = dynamic_cast<At_Root_Block*>(parent);
+      Statement* ret = this->visit_children(ar->block());
+
+      this->parent = old_parent;
+      this->parents = old_parents;
+
+      return ret;
     }
 
-    parent_stack.pop_back();
+
+    if (!this->is_transparent_parent(parent, old_parent)) {
+      this->parent = parent;
+    }
+
+    this->parents.push_back(parent);
+
+    Block* b = dynamic_cast<Block*>(parent);
+
+    if (!b) {
+      if (Has_Block* bb = dynamic_cast<Has_Block*>(parent)) {
+        b = bb->block();
+      }
+    }
+
+    if (b) {
+      for (auto n : *b) {
+        n->perform(this);
+      }
+    }
+
+    this->parent = old_parent;
+    this->parents.pop_back();
+
     return b;
   }
 
-  Statement* CheckNesting::operator()(Declaration* d)
+
+  Statement* CheckNesting::operator()(Block* b)
   {
-    if (!is_valid_prop_parent(parent())) {
-      throw Exception::InvalidSass(d->pstate(), "Properties are only allowed "
-        "within rules, directives, mixin includes, or other properties.");
-    }
-    return static_cast<Statement*>(d);
+    return this->visit_children(b);
   }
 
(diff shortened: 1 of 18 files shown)


mgreter:
Overall: LGTM.

From what I see in the code, the `Trace` AST is pretty tightly coupled to mixins, so maybe `TraceMixin` would be a bit more clear about that. And it might make sense to only store the pointer to the mixin `c` instead of a string from `c->name()`?

I will add some additional minor nitpicking directly to the code in this PR ...

src/check_nesting.cpp
```diff
@@ -6,52 +6,413 @@
...
+  //       parent->pstate(),
+  //       "CSS import directives may only be used at the root of a document."
+  //     );
+  //   }
+
+  //   return true;
+  // }
+
+  bool CheckNesting::is_invalid_mixin_definition_parent(Statement* parent)
+  {
```
mgreter: has `bool` signature but only returns false?

src/check_nesting.cpp
```diff
@@ -6,52 +6,413 @@
...
+          "Mixins may not be defined within control directives or other mixins."
+        );
+      }
+    }
+
+    return false;
+  }
+
+  bool CheckNesting::is_invalid_function_parent(Statement* parent)
+  {
```
mgreter: same as above and also applies to functions below

src/expand.cpp
```diff
@@ -86,6 +87,7 @@ namespace Sass {
 
   Statement* Expand::operator()(Ruleset* r)
   {
+    bool old_at_root_without_rule = this->at_root_without_rule;
```
mgreter: You should be able to write that as `LOCAL_FLAG(at_root_without_rule, false);`
The macro will create a small stack object that will ensure the flag is reset once it goes out of scope. IMO that is much safer than resetting the flag on your own, since it will also handle cases where we exit via a throw. I have added that [Macro (and the helper class)] [1] some time ago.

[1]: https://github.com/sass/libsass/blob/826827a6a52b107d471958fc34dcafbb847ea6e4/src/ast_def_macros.hpp#L5-L27

xzyfer:
>so maybe TraceMixin would be a bit more clear about that

I decided to keep it Trace because that's the Ruby name. I would also like to come back to renaming `Declaration` to `Prop`. It helps if we're all speaking the same language.

> And it might make sense to only store the pointer to the mixin

The implement doesn't care about the mixin itself, just it's name and pstate. I presume the name for displaying error messages somewhere I haven't seen yet.

src/expand.cpp
xzyfer: Ah I see, I the macro but wans't 100% clear on when it pop off the stack.

src/check_nesting.cpp
xzyfer: There either return false or thrown error. The Ruby Sass implement is return the error string or false which is gross in C++.

src/expand.cpp
xzyfer: Unfortunately this doesn't work in all cases like here when I need to pop off that stack at a specific point. Waiting until the end of the function would alter the implementation. IMO we should aim to use macro where possible and assume if we're not it's intentional.

src/check_nesting.cpp
mgreter: ACK. But wouldn't it be more sane to just make them `void` then? Not really a big deal, but signature just indicates something it is not actually doing this way.

xzyfer: I think void might actually make the usage nicer.

src/expand.cpp
mgreter: I case you need better control you could create a variable scope local enough for your use-case.

mgreter:
@xzyfer so the name is actually not used so far? Leaving the `Trace` to basically be a Block node with some special meaning (as in being a result of a mixing call)?

src/expand.cpp
mgreter: After a second look, shouldn't it be reset on [line 100] [1] too?

[1]: https://github.com/sass/libsass/pull/2089/files#diff-ff25c22e06664808e8ba36fcd39b6439R100

src/expand.cpp
```diff
@@ -98,6 +100,8 @@ namespace Sass {
       return k;
     }
 
+    this->at_root_without_rule = false;
```
mgreter: On a third look I don't see why it shouldn't work here if you add the `LOCAL_FLAG` call here ... Since you seem to reset it just right before you exit the function (and therefore this variable scope).

xzyfer:
@mgreter I explain in more detail in #2086 

>The Trace node is a simple wrapper around the result of a mixin call 
and `@content`. It exists as a way to group the resulting properties
as a block but maintain some metadata about the initiating mixin call 
or `@content` block.

>It also plays an important rule is bubble those blocks with `@at-root` 
and nested property sets. This required implementing the final
 missing `cssize` methods. As a result we should a complete `@at-root` 
implementation.

mgreter:
@xzyfer I fail to see what the "metadata" is ... maybe you can elaborate a bit more on that.

src/expand.cpp
xzyfer: Sorry, you're correct.

src/check_nesting.cpp
mgreter: Agree. It probably would make sense to just name it "check_mixin_definition_parent" etc. That would also make it more clear that they actually do throw an error (I don't really expect that from something prefixed "is_").

xzyfer:
I'm not 100% sure myself. As far as I can tell the Trace is a block
around the resulting output of a mixin call. Instead of appending
the individual resulting nodes to the parent block, the Trace block 
is appended. The Trace node keeps a reference to the pstate of
the mixin that created and it's name.

The primary usecase appears to be in maintaining a backtrace in
case the resulting nodes end up in a location that's invalid. FWIW
I don't believe I'm correctly handling the backtrace in this PR. Since
it's not covered by specs it's a manual process.

This is the explanation I got from @nex3 

>it allows us to reconstruct a stack trace that includes mixin invocations even after mixins have been resolved
>it wraps the output of a mixin invocation and remembers that mixin's name and line/filename information
>so that if, later on, we run into an error in that output
>we can tell the user "this was caused by including this mixin"

src/check_nesting.cpp
xzyfer: Made them `void` and removed the `is_` prefix. This makes the method names the same as Ruby Sass.

mgreter:
OK, so it is pretty much the same thing I need for exposing import stack on C-API. After imports are resolved (in eval) we have the full AST Tree with `Function_Call` and once that call is evaluated, the context from where these calls were imported is lost (since it is not preserved in the AST Tree). So the easiest approach that came to my mind was to add an AST Node just to preserve that information, which is basically the same from what I get from your explanation ...

xzyfer:
That's my understanding. The node was originally called MixinTraceNode in Ruby Sass but was renamed to TraceNode but is still so far only used for mixins.

mgreter:
OK, that clears some things up. I will see if I can re-use the `Trace` Node for what I need with imports. BTW I'm aware of the situation with the various StackTraces. We have something in `Env` and also various stack objects in various phases. I think we should refactor and unify all these cases in one smart class that can be easily handled and queried. But I think we first need to add the traces feature for the imports and (custom) functions. This work should also include the `--trace` cli option.

xzyfer:
> I will see if I can re-use the Trace Node for what I need with imports

I would avoid this. Ruby Sass is explicit about it's usecase.

># A solely static node left over after a mixin include or @content has been performed.
># Its sole purpose is to wrap exceptions to add to the backtrace.
>[Source](https://github.com/sass/sass/blob/stable/lib/sass/tree/trace_node.rb#L4-L5)

There are various assumption made in places like check nesting and cssize visitors around this.

xzyfer:
I addressed the feedback, and did some additional clean up.
It's late here so I'm not going to wait up for CI. 
I'll ship this in the morning.

xzyfer:
I've also made sure the respective PRs this references are up to date with the changes made here so they serve as an accurate reference to the intentions of this merge.

mgreter:
Nice work! Thx for that PR!

PR #2109, opened 2016-06-19
Use mkdir -p

AMDmi3:
Use mkdir's -p flag for to not fail on trying to create already existing directories


xzyfer:
This has come up a couple times. There are compatibility issues with `-p`. We're open to a real fix but it's much more involved than this.

https://github.com/sass/libsass/issues/1433
https://github.com/sass/libsass/pull/1365#issuecomment-123325642
https://github.com/sass/libsass/pull/795#discussion_r22396485

PR #2118, opened 2016-07-03
Fix issue with invalid duplicate keys

mgreter:
Keys `1` and `1px` were considered duplicates if they end
up in the same bucket. In sass these are indeed equal but
that does not seem to apply when used as hash keys.

@xzyfer feel free to refactor as you see fit.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -2004,6 +2004,21 @@ namespace Sass {
     return false;
   }
 
+  bool Number::eq (const Expression& rhs) const
+  {
+    if (const Number* r = dynamic_cast<const Number*>(&rhs)) {
+      size_t lhs_units = numerator_units_.size() + denominator_units_.size();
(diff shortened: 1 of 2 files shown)


mgreter:
http://www.cpantesters.org/cpan/report/276835f4-40a5-11e6-8807-814d1da4c10f
http://www.cpantesters.org/cpan/report/739e97e8-4068-11e6-8807-814d1da4c10f
http://www.cpantesters.org/cpan/report/958a4a1e-402c-11e6-8807-814d1da4c10f

xzyfer:
Thanks @mgreter. Looks good. I'll write a test and merge soon.

mgreter:
@xzyfer this is an edge case and only happens if the hash table implementation puts both keys in the same bucket for some reason (it should not). This is not the case in any of libsass' CI envs. I only saw it with some compilers and perl cpan "CI" (see links above).

xzyfer:
👍

Issue #2120, opened 2016-07-11
Malformed URL exception when contains double-byte characters


roydukkey:
The input provided below is a logical example for when double-byte character may be used. However, to verify this issue on a smaller URL, the following can be tested and produces the same exception: [네이버.com](http://네이버.com). Single quotes around the URL will work around the issue.

input.scss
```sass
@import url(//fonts.googleapis.com/css?family=Signika+Negative&text=“”)
```

libsass 3.3.6
```css
Error: malformed URL
        on line 1 of stdin
>> @import url(//fonts.googleapis.com/css?family=Signika+Negative&text=“”)
   --------^
```

ruby sass 3.4.21
```css
@charset "UTF-8";
@import url(//fonts.googleapis.com/css?family=Signika+Negative&text=“”);
```

version info:
```cmd
$ node-sass --version
node-sass       3.8.0   (Wrapper)       [JavaScript]
libsass         3.3.6   (Sass Compiler) [C/C++]
```

PR #2125, opened 2016-07-19
Allow unicode chars in unquoted urls

mgreter:
Fixes https://github.com/sass/libsass/issues/2120


xzyfer:
I don't know about this. It breaks away from the Ruby Sass implementation which suggests the real bug is else where.

PR #2129, opened 2016-07-21
Clarify API documentation of custom importers.

### LINKED ISSUE #2128: Documentation of parameter srcmap of sass_make_import_entry
Hey,

I'd like to know how the parameter `srcmap` of `sass_make_import(_entry)` is supposed to be used. I checked some wrapper implementations but the only thing I could find was [this unit test in libsass-python](https://github.com/dahlia/libsass-python/blob/master/sasstests.py#L332) but the author seems to be clueless as well. [sassc-ruby](https://github.com/sass/sassc-ruby/blob/master/lib/sassc/import_handler.rb#L39) on the other hand calls this parameter `source_map_path` but seems to always pass nil. [perl-libsass](https://github.com/sass/perl-libsass/blob/develop/Sass.xs#L501) refers to this parameter as a json string (which matches libsass-python) but includes no unit test where it is used.
It would be nice if someone could give me an example or hint about this parameter. Thanks 😄 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36209690-documentation-of-parameter-srcmap-of-sass_make_import_entry?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DerPapst:
Describe parameter srcmap of sass_make_import() in more detail.

Closes #2128

### DIFF
--- a/docs/api-importer.md
+++ b/docs/api-importer.md
@@ -11,6 +11,8 @@ rv[0] = sass_make_import(rel, abs, source, srcmap);
 
 Every import will then be included in LibSass. You are allowed to only return a file path without any loaded source. This way you can ie. implement rewrite rules for import paths and leave the loading part for LibSass.
 
+Please note that LibSass doesn't use the srcmap parameter yet. It has been added to not deprecate the C-API once support has been implemented. It will be used to re-map the actual sourcemap with the provided ones.
+
 ### Basic Usage
 
 ```C


xzyfer:
Thanks @DerPapst

Issue #2156, opened 2016-08-31
The unquote function is not consistent with Ruby Sass


trevorsg:
### Title: unquote("\"hello\" and \"goodbye\"") incorrectly strips the first and last quotes.

input.scss
```scss
@debug unquote("\"hello\" and \"goodbye\"");
```

[libsass] [1]
```css
DEBUG: hello" and "goodbye
```

ruby sass 3.4.19
```css
DEBUG: "hello" and "goodbye"
```

version info:
```cmd
$ sassc.exe --version
sassc: [NA] (master: 077af530)
libsass: [NA] (master: 014a61b6)
sass2scss: 1.0.6
sass: 3.4
```

[1]: http://libsass.ocbnet.ch/srcmap/#dGVzdCB7CiAgY29udGVudDogYmFyOyB9Cg==

mgreter:
Seems to be a duplicate of https://github.com/sass/libsass/issues/2153

Issue #2153, opened 2016-08-19
Concatenating an empty string w/ a quoted string will unquote the string


trevorsg:
### "\"foo\"" + ""  ==> foo

input.scss
```scss
@debug "\"foo\"" + "";
// --> foo (expected/ruby sass: "foo")
@debug "" + "\"foo\"";
// --> foo (expected/ruby sass: "foo")
@debug "" + "\"foo";
// --> "foo (as expected)
@debug "\"foo\"" + "bar";
// --> "foo"bar (as expected)
```

version info (Windows):
```cmd
$ sassc.exe --version
sassc: [NA] (master: 077af530)
libsass: [NA] (master: 014a61b6)
sass2scss: 1.0.6
sass: 3.4
```

[1]: http://libsass.ocbnet.ch/srcmap/#dGVzdCB7CiAgY29udGVudDogYmFyOyB9Cg==

xzyfer:
I can confirm this issue. Spec added https://github.com/sass/sass-spec/pull/892

PR #2160, opened 2016-09-02
Remove unquoting for @debug and @warn statements

mgreter:
- Fixes https://github.com/sass/libsass/issues/2153
- Fixes https://github.com/sass/libsass/issues/2156

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -1580,10 +1580,13 @@ namespace Sass {
   ////////////////////////////////////////////////////////
   class String_Quoted : public String_Constant {
   public:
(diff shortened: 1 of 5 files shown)


xzyfer:
It's worth double checking the #2153 spec. It doesn't use @debug or @warn.

https://github.com/sass/sass-spec/pull/892

mgreter:
@xzyfer thanks, didn't see it. IMO it would be worth to ammend the original issue report with such information or additional test cases, since it was not that easy to spot.

xzyfer:
@mgreter I commented a like to the specific however I agree adding it to
the issue description is more visible. Will do so in the future.

Issue #2158, opened 2016-08-31
CSS comment inside of @supports incorrect CSS output


zachleat:
Input:
```scss
@supports ( box-shadow: none ) {
	/* Shadow on first body column */
	.sizechart {
		position: relative;
	}
}
```

libsass 3.3.6
```css
@supports (box-shadow: none) { {
    /* Shadow on first body column */
  }
  .sizechart {
    position: relative;
  }
}
```

Sass 3.4.21
```css
@supports (box-shadow: none) {
  /* Shadow on first body column */
  .sizechart {
    position: relative;
  }
}
```

[Sassmeister Link](http://www.sassmeister.com/gist/47391852c9759e6e0dd7e15cfc3fcfd8)

PR #2161, opened 2016-09-02
Remove hoistable flag and code paths

mgreter:
Fixes https://github.com/sass/libsass/issues/2158


mgreter:
I ended up removing the hoistable flag completey. I don't know what the purpose of hoistable was or why it would be needed. I couldn't find anything "hoist" related in ruby sass code ...

PR #2163, opened 2016-09-02
Use data error code 65 to match Ruby Sass

### LINKED ISSUE #2162: Data error code of 1 doesn't match Ruby Sass return values
While looking at getting the newer sass-spec running for node-sass, I noticed that there is a missmatch in the error codes. It looks like this was changed in Sass 3.4.5 https://github.com/sass/sass/blob/4764015f54a191ebe6fb5f912cab16b28fc324c3/doc-src/SASS_CHANGELOG.md#345-19-september-2014 to match the FreeBSD exit codes https://www.freebsd.org/cgi/man.cgi?query=sysexits.

This might be a breaking change to fix, so I thought I'd ask before trying to PR something.
##

nschonni:
Fixes #2162 
This was guess work, but ties into https://github.com/sass/sassc/pull/182 to see if we can start to match the Ruby Sass data error

### DIFF
--- a/src/sass_context.cpp
+++ b/src/sass_context.cpp
@@ -109,7 +109,7 @@ extern "C" {
       }
 
       JsonNode* json_err = json_mkobject();
-      json_append_member(json_err, "status", json_mknumber(1));
+      json_append_member(json_err, "status", json_mknumber(65));
       json_append_member(json_err, "file", json_mkstring(e.pstate.path));
       json_append_member(json_err, "line", json_mknumber((double)(e.pstate.line+1)));
       json_append_member(json_err, "column", json_mknumber((double)(e.pstate.column+1)));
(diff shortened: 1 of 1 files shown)


mgreter:
The mapping to OS semantics should be done in sassc. The code 65 means "ERROR_NETWORK_ACCESS_DENIED" in windows. Therefore this should go completely to sassc (just return 65 as exit code there if a libsass error occured).

xzyfer:
As mentioned in https://github.com/sass/sassc/pull/182#issuecomment-244284837

We should probably expose some agnostic error code. Possibly part to the API.

mgreter:
IMO we should keep it as it is in libsass (it is already agnostic). I think this is something implementers have to do. I think we can give the recommendation, that if libsass return an error code of 1, the corresponding OS exit codes should be 65 on POSIX compatible OSs (*nix) and 13 on Windows.

nschonni:
Closing this one since Ruby now returns 13 on Windows. I'll update the issue to reflect this

PR #2171, opened 2016-09-06
Improve memory footprint when evaluating in loops

mgreter:
We are cloning/copying too many objects when evaluating loops. i.e. a simple `map-get` call will always clone the full map that is passed, which can lead to an absurd amount of allocated memory. But we cannot really cache the results as 1) variables can be re-assigned and 2) some code paths (i.e. equality checks) evaluate binary expressions different than normaly. This fix will only cache "delayed" results, which should cover 90% of the cases. This should also improve performance in certain cases by a lot.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -141,7 +141,7 @@ namespace Sass {
                bool d = false, bool e = false, bool i = false, Concrete_Type ct = NONE)
     : AST_Node(pstate),
       is_delayed_(d),
-      is_expanded_(d),
+      is_expanded_(e),
       is_interpolant_(i),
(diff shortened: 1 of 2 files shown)


mgreter:
This would be ready but is currently blocked by https://github.com/sass/sassc/issues/187

xzyfer:
I have a feel delayed values are not long for this earth

xzyfer:
This broke a lot of eyeglass tests i.e. https://github.com/sass-eyeglass/eyeglass/blob/master/test/test_assets.js#L31-L52

I need to investigate further after some sleep.

PR #2178, opened 2016-09-12
Support list functions on maps

### LINKED ISSUE #1930: Support maps in list context (index, nth, etc)

xzyfer:
When Sass added maps the updated the list functions to operate on lists. With a list function gets a map it's treated as a list of key value lists. i.e. `(a: 1, b: 2)` becomes `(a 1), (b 2)`. When I originally implemented maps I only added map support for a subset of list functions.

- [x] append
- [x] index
- [x] join
- [x] length
- [x] nth
- [x] set-nth
- [x] zip

Fixes #1930
Specs sass/sass-spec#919

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -2226,4 +2226,20 @@ namespace Sass {
     }
   }
 
+  //////////////////////////////////////////////////////////////////////////////////////////
+  // Convert map to (key, value) list.
+  //////////////////////////////////////////////////////////////////////////////////////////
+  List* Map::to_list(Context& ctx, ParserState& pstate) {
+    List* ret = SASS_MEMORY_NEW(ctx.mem, List, pstate, length(), SASS_COMMA);
+
+    for (auto key : keys()) {
+      List* l = SASS_MEMORY_NEW(ctx.mem, List, pstate, 2);
+      *l << key;
+      *l << at(key);
+      *ret << l;
+    }
+
+    return ret;
+  }
+
 }
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -924,6 +924,7 @@ namespace Sass {
     std::string type() { return "map"; }
     static std::string type_name() { return "map"; }
     bool is_invisible() const { return empty(); }
+    List* to_list(Context& ctx, ParserState& pstate);
 
     virtual size_t hash()
     {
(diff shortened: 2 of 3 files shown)


xzyfer:
I want to build some confidence in error handling but otherwise this is pretty close.

src/ast.cpp
```diff
@@ -2226,4 +2226,20 @@ namespace Sass {
     }
   }
 
+  //////////////////////////////////////////////////////////////////////////////////////////
+  // Convert map to (key, value) list.
+  //////////////////////////////////////////////////////////////////////////////////////////
+  List* Map::to_list(Context& ctx, ParserState& pstate) {
```
xzyfer: This might be some what wasteful but it's equivalent to what Ruby Sass does and makes list functions easy to reason about for newcomers.

xzyfer: As an aside it bugs that the AST doesn't have access to the context. Not being about to create new nodes from AST methods gets rather painful. In this case it might be nicer to add a `List` constructor that takes a `Map` argument.

xzyfer: > In this case it might be nicer to add a List constructor that takes a Map argument.
 Add a line note

This runs into the issue of not being able to create new list nodes in AST nodes.

Issue #2185, opened 2016-09-22
at-root does not work if there is an internal if statement


speedplane:
### `at-root` does not work if there is an internal `if` statement

input.scss
```
$bar: true;
p {
	@if $bar {
		@at-root {
			div {
				width: 100px;
			}
		}
	}
	
	@at-root {
		@if $bar {
			div {
				height: 100px;
			}
		}
	}
}
```

[libsass 3.5.5] [1]
```
div {
  width: 100px;
}

p div {
  height: 100px;
}
```

Expected result:
```
div {
  width: 100px;
}

div {
  height: 100px;
}
```
[1]: http://libsass.ocbnet.ch/srcmap/#dGVzdCB7CiAgY29udGVudDogYmFyOyB9Cg==

mgreter:
Updated the test to also include all? other similar statements:

```scss
$bar: true;

@mixin mixin() {
    mix: in;
}

@mixin include() {
    @content;
}

p {
    @at-root {
        @if $bar {
            #if {
                height: 100px;
            }
        }
        @if (not $bar) {
        } @elseif($bar) {
            #elseif {
                width: 100px;
            }
        }
        @if (not $bar) {
        } @else {
            #else {
                width: 100px;
            }
        }
        @for $i from 1 through 1 {
            #for {
                foo: bar#{$i};
            }
        }
        $i: 0;
        @while ($i == 0) {
            $i: $i + 1;
            #while {
                foo: bar#{$i};
            }
        }
        @each $i in (1) {
            #each {
                foo: bar#{$i};
            }
        }
        #inc {
            @include mixin();
            @include include() {
                inc: lude;
            }
        }
        @supports (display: flex) {
            display: flex;
        }
        @foo {
           bar: bat;
        }
    }
}
```

PR #2187, opened 2016-09-24
Fix @at-root for loop and logic statements

mgreter:
The root flag should be inherited from the parent, since
the statements are transparent (no CSS blocks in output).

Fixes  https://github.com/sass/libsass/issues/2185


speedplane:
That was fast, thank you!

Issue #2149, opened 2016-08-16
/ interpreted as division instead of separator


chriseppstein:
### Title: bare `/` should not be interpreted as division with bare literals

Originally filed:
https://github.com/sass/sass/issues/2118

For more info see:
https://github.com/sass/sass/issues/667#issuecomment-14086477

input.scss
```scss
.foo {
  background: url('background.png') -10px -10px/110% no-repeat
}
```

[libsass 3.5.5] [1]
```txt
// error
{
  "formatted": "Error: -0.09091px/% isn't a valid CSS value.\n        on line 2 of foo.scss\n>>   background: url('background.png') -10px -10px/110% no-repeat\n   ------------------------------------------^\n",
  "message": "-0.09091px/% isn't a valid CSS value.",
  "column": 43,
  "line": 2,
  "file": "/Users/ceppstei/Projects/broccoli-multi-filter/foo.scss",
  "status": 1
}
```

ruby sass 3.4.22
```css
.foo {
  background: url("background.png") -10px -10px/110% no-repeat; }
```

version info:
```cmd
$ node-sass --version
node-sass      	3.8.0  	(Wrapper)      	[JavaScript]
libsass        	3.3.6  	(Sass Compiler)	[C/C++]
```

Spec added https://github.com/sass/sass-spec/pull/909

gakimball:
The compiler seems to only try division if one or both of the operands is negative.

This will compile as-is:

```scss
.selector {
  background: #fff url('') 100px 20px / 100%;
}
```

This will produce an error:

```scss
.selector {
  background: #fff url('') 100px -20px / 100%;
}
```

For anyone who needs a temporary workaround, you can interpolate the negative value to bypass the error.

```scss
.selector {
  background: #fff url('') 100px #{-20px} / 100%;
}
```

PR #2189, opened 2016-09-25
Delay Unary_Expression according to parsed operand

mgreter:
Fixes https://github.com/sass/libsass/issues/2149

### DIFF
--- a/src/debugger.hpp
+++ b/src/debugger.hpp
@@ -552,6 +552,7 @@ inline void debug_ast(AST_Node* node, std::string ind, Env* env)
     Unary_Expression* expression = dynamic_cast<Unary_Expression*>(node);
     std::cerr << ind << "Unary_Expression " << expression;
     std::cerr << " [interpolant: " << expression->is_interpolant() << "] ";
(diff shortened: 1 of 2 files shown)


xzyfer:
This caused a regression that was hidden by a bug in Sass spec - https://github.com/sass/sass-spec/issues/932

mgreter:
My bisect shows the regression happened much earlier in 734668f1803bc7ad5315f05bc026a1a2d593349a

It basically boils down that I did not anticipate the following sass behavior (3.4.22):

```scss
h1 {
  delayd: #{5% / 2};
  delayd: #{+5% / 2};
  evaled: #{-5% / 2};
}
```

```css
h1 {
  delayd: 5%/2;
  delayd: 5%/2;
  evaled: -2.5%;
}
```

xzyfer:
@mgreter do you think you'll have a chance to look into this? Otherwise I may have to back this patch out of 3.4.

xzyfer:
@mgreter I talked to @nex3, she confirmed the `evaled` case is bug in Sass.

>that's a bug, the last one should be -5%/2

PR #2196, opened 2016-09-28
Don't probe tests in CI

xzyfer:
Sass spec has a bug with probe todo that causes some failures to be
hidden. As a result #2189 caused a regression.

/cc @mgreter


xzyfer:
The sass spec bug in question is https://github.com/sass/sass-spec/issues/932

PR #2210, opened 2016-10-19
Remove -ldl flag from OpenBSD too

parhs:
Probably OpenBSD should be included too. I couldn't compile it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,8 +125,10 @@ endif
 
 ifneq (Windows,$(UNAME))
 	ifneq (FreeBSD,$(UNAME))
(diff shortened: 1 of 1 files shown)


mgreter:
Seems reasonable. @parhs Mind PRing and testing [sassc] [1] too?

[1]: https://github.com/sass/sassc/blob/master/Makefile#L138-L141

xzyfer:
No objections from me. @saper have thoughts?

parhs:
@mgreter I was looking for that thanks. I'll make a pull request there too. Did  test here fail because of timeout?

Issue #2205, opened 2016-10-13
Using an overloaded functions with call() causes fatal error


JohnAlbin:
### Title: Using an overloaded functions with call() causes fatal error

If you try to use an overloaded function inside the `call()` function, it will work just fine in Ruby Sass, but libSass will throw an error. For example, `rgba()` can be called with either 4 params ($red, $green, $blue, $alpha) or with 2 params ($hex, $alpha).

input.scss
```scss
// Use the 2 parameter form of rgba().
$params: #fff .5;
test {
  color: call(rgba, $params...);
}
```

libSass 3.3.6
```css
overloaded function `rgba` given wrong number of arguments on line 4 at column 10
```

ruby sass 3.4.21
```css
test {
  color: rgba(255, 255, 255, 0.5);
}
```

version info:
```cmd
$ node-sass --version
node-sass	3.10.1	(Wrapper)	[JavaScript]
libsass  	3.3.6	(Sass Compiler)	[C/C++]
```

The error is thrown from this line: https://github.com/sass/libsass/blob/master/src/eval.cpp#L834

JohnAlbin:
I found a simpler test case that didn't require using `call()`, so I've updated the above test case and the bug description. Both the old and the new test case produce the same error message.

PR #2215, opened 2016-10-20
Fix calling overloaded stub functions with rest arguments

mgreter:
Fixes https://github.com/sass/libsass/issues/2205


xzyfer:
lgtm

PR #2216, opened 2016-10-21
Bugfixes for source maps

mgreter:
Fixes https://github.com/sass/libsass/issues/2204
Replaces https://github.com/sass/libsass/pull/2208

- Fixes parent selector mappings
- Fixes media block/query mappings
- Fixes range over binary expressions
- Don't include semicolon for statics
- Fixes variable assignment mappings

I have already deployed the changes to my source map inspector!

### Parent selector mappings
[SourceMap Inspector] [1] (click [inspect] and use keyboard arrows to navigate)
```scss
.foo {
  background: red;
  &-bar {
    background: green;
  }
}
```

### Media Block expressions
[SourceMap Inspector] [2] (click [inspect] and use keyboard arrows to navigate)
```scss
.foo {
  background: red;
  @media only screen {
    background: green;
  }
}
```

### Binary expressions
[SourceMap Inspector] [3] (click [inspect] and use keyboard arrows to navigate)
```scss
.foo {
  bar: (1px+3px);
}
```

### Static values
[SourceMap Inspector] [4] (click [inspect] and use keyboard arrows to navigate)
```scss
.foo {
  bar: green   ;
}
```

### Variables (last assignment)
[SourceMap Inspector] [5] (click [inspect] and use keyboard arrows to navigate)
```scss
$var: bar;
$var: (foo + $var);
test { val: $var; }
```

[1]: http://libsass.ocbnet.ch/srcmap/#LmZvbyB7CiAgYmFja2dyb3VuZDogcmVkOwogICYtYmFyIHsKICAgIGJhY2tncm91bmQ6IGdyZWVuOwogIH0KfQ==
[2]: http://libsass.ocbnet.ch/srcmap/#LmZvbyB7CiAgYmFja2dyb3VuZDogcmVkOwogIEBtZWRpYSBvbmx5IHNjcmVlbiB7CiAgICBiYWNrZ3JvdW5kOiBncmVlbjsKICB9Cn0=
[3]: http://libsass.ocbnet.ch/srcmap/#LmZvbyB7CiAgYmFyOiAoMXB4KzNweCk7Cn0=
[4]: http://libsass.ocbnet.ch/srcmap/#LmZvbyB7CiAgYmFyOiBncmVlbiAgIDsKfQ==
[5]: http://libsass.ocbnet.ch/srcmap/#JHZhcjogYmFyOwokdmFyOiAoZm9vICsgJHZhcik7CnRlc3QgeyB2YWw6ICR2YXI7IH0K

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -122,6 +122,11 @@ namespace Sass {
     pstate_.offset += pstate - pstate_ + pstate.offset;
   }
 
+  void AST_Node::set_pstate_offset(const Offset& offset)
+  {
+    pstate_.offset = offset;
+  }
+
   inline bool is_ns_eq(const std::string& l, const std::string& r)
   {
     if (l.empty() && r.empty()) return true;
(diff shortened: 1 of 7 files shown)


src/parser.cpp
```diff
@@ -1527,6 +1546,9 @@ namespace Sass {
   {
     lex< static_value >();
     Token str(lexed);
+    // static valuse always have trailing white-
```
nschonni: valuse -> values

src/ast.cpp
```diff
@@ -1150,9 +1163,18 @@ namespace Sass {
...
-                s->append(ctx, ss);
+                // adjust for parent selector (1 char)
+                ParserState nstate = (*h)[0]->pstate();
+                nstate.offset.column += 1;
+                nstate.column -= 1;
+                (*h)[0]->pstate(nstate);
+                // keep old parser state
+                s->pstate(pstate());
+                // append new tail
+                  s->append(ctx, ss);
```
nschonni: Indent

mgreter:
thx @nschonni. Fixed the formatting and type!

xzyfer:
Thank you @mgreter. Please try to keep PRs focused. This PR is address 4 different issue of differing merit.

I would prefer not merge the makefile change. The existing order is obvious and self documenting.

xzyfer:
I've noticed this happening recently. I don't think it's related to the build time. The build completes but CI hangs for over 30minutes seemingly waiting to execute sass-spec.

```
Fetching info for PR 2193
Run options: --seed 42180
# Running:
```

xzyfer:
@mgreter I have cherry picked 314360e and 51ff53f on to master. Please back out f8d9dec, then feel free to ship this. 

I'll be feature freezing master after this merge so I release a 3.4 RC this weekend.

mgreter:
Hmm, https://github.com/sass/libsass/commit/f8d9decfdf860d4c91a4ec29cc17b62c80c445db saves me 1 minute compile time. But whatever ...

xzyfer:
I understand as a power user this is of benefit to you. My reasoning is

- it's something that we can't guarantee the correctness of, so it will regress
- it's unclear what to do when adding new files

Given the opportunity I'd prefer to optimise for making life easier for new contributors.

mgreter:
@xzyfer the order is not really important as long as all files are listed. If a file is missing it will fail in CI as functions will be missing in the executable. I don't see any danger at all here!

mgreter:
And it was as unclear where to add files before as it is now. The order is now even a bit more logic as I grouped the related stuff together where possible. I just needed to move the bigger files more to the top, so I don't end up with one as the last compile unit.

xzyfer:
Currently they're listed alphabetically so it was  no brainer where a new file belonged in the list.

I'm happy to comprise. Maintain alphabetical ordering, but move the problematic large files to the top, preferably with a comment for why it's out of order.

mgreter:
I'm ok with adding a comment, but the order is the best as it is now. It is a compromise between having the biggest compile units on top without generating a too heavy memory peak (it's distributed more evenly).

xzyfer:
This is good. Thank you.

Issue #1898, opened 2016-01-27
Line comments should be relative to cwd


xzyfer:
When displaying line comments in output Ruby Sass shows the file paths as relative to cwd, LibSass shows an absolute path.

Given
```
|- foo
|  |- bar.scss
|- test.scss
```
Running 
```
sassc -l foo/bar.scss
```
Ruby Sass
```
/* line 1, ../test.scss */
a {
  color: black; }

/* line 3, bar.scss */
b {
  color: red; }
```
LibSass 3.3.3
```
/* line 1, /Sass/libsass/test.scss */
a {
  color: black; }

/* line 3, /Sass/libsass/foo/bar.scss */
b {
  color: red; }
```

mgreter:
URLs by sass don't seem to be relative to CWD. That makes it harder to implement correctly ...

mgreter:
Retested again and it seems to be inline with ruby sass. Original test case here was not very clear where the import is as the contents of each file were not given in this issue.

mgreter:
@xzyfer Is it possible to switch on the line-comments flag via options.yml?

xzyfer:
@mgreter it is not. I would say it's out of scope for sass spec, much like sourcemaps and alternate output formats (except special compressed cases).

PR #2219, opened 2016-10-22
Make paths in source comments relative to CWD

mgreter:
Fixes https://github.com/sass/libsass/issues/1898


mgreter:
Paths by sass don't seem to be relative to CWD. That makes it harder to implement correctly ...

mgreter:
Retested again and it seems to be inline with ruby sass. Original test case here was not very clear where the import is as the contents of each file were not given.

PR #2241, opened 2016-12-11
Fix incorrect str-slice behaviour

### LINKED ISSUE #2240: Incorrect str-slice() behavior
LibSass replicates a Ruby Sass bug in the `str-slice()` function: if the `$end-at` parameter is negative and its absolute value is between the length of the string and two times the length of the string, the function should return an empty string, but it doesn't. For example, `str-slice("abcdef", 0, -8)` should return `""` but actually returns `"abcde"`.

I'm working on a fix for Ruby Sass. Once I land it, I'll mark specs affected by this as TODO for LibSass.

xzyfer:
When is `$end-at` is more negative than the length of the string
return an empty string.

Fixes #2240
Spec sass/sass-spec#997

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -1020,16 +1020,19 @@ namespace Sass {
         std::string str = unquote(s->value());
 
         size_t size = utf8::distance(str.begin(), str.end());
+
         if (end_at <= size * -1.0 && size > 1) {
-          end_at += size;
-          if (end_at == 0) end_at = 1;
+          end_at = 0;
         }
         if (end_at < 0) {
           end_at += size + 1;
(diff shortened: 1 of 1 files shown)


xzyfer:
@nex3 was this intentional or some github magic from https://github.com/sass/sass-spec/commit/b00da4024fa1c90ef5cd9ff1503bdcee3069ec33?

This PR was not ready.

nex3:
Intentional, sorry! I thought this was ready since its specs were passing.

xzyfer:
The specs were only passing because LibSass was still todo'd :) I'll finish this tonight so I wont bother reverting.

Issue #2246, opened 2016-12-14
"can't extend nested selectors" when using @at-root


esr360:
Thanks for all of your hard work getting Libsass up-to date with Node-Sass version 3.4 - I have been waiting a long time for this change, but I regret to say my app will still not compile using Libsass.

I am experiencing an issue when using @extend and @at-root, 2 things I know you guys have struggled with at Libsass, so I apologise if you are aware of this issue.

```scss
@mixin foo($option: 'foo') {
     // Create a unique, random placeholder to store styles
    $placeholder : $option + random(9999);

    // Store the styles in the placeholder
    @at-root %#{$placeholder} {
        content: 'foo';
    }

    @at-root {
        .bar {
            @extend %#{$placeholder};
        }
    }
}

@mixin bar($option) {
    @include foo($option);
}

.foo {
    @include bar('baz');
}
```

libsass 3.4.0
```css
Error: Can't extend %foo6595: can't extend nested selectors
```

ruby sass 3.4.21
```css
.bar {
  content: 'foo';
}
```

version info:
```cmd
node - 6.9.1
grunt-sass@2.0.0
```

mgreter:
Thanks for the report. This edge case was not yet covered by our spec tests.
This will be fixed with https://github.com/sass/libsass/pull/2247

esr360:
Amazing, thanks for the quick fix.

PR #2247, opened 2016-12-14
Fixes extensions with selector schema in at-root

mgreter:
Fixes https://github.com/sass/libsass/issues/2246


xzyfer:
Fixes master. This is green now. Ready to go when there are tests.

Issue #1037, opened 2015-04-02
[FR] Expose `ParserState` of currently invoked function [$100]


am11:
As per the discussion that transpired at: https://github.com/sass/node-sass/issues/646#issuecomment-87627653

As an implementer I want to propagate the contextual information, in particular, the file path, line and column numbers to the end-user so that they know from where the function was called.

mgreter:
Technically this translates to beeing able to access the `ParserState` of the currently invoked function.
This is pretty much the same as the API for accessing the import stack (just return a new sruct type).

chriseppstein:
Related: https://github.com/sass/libsass/pull/1227

mgreter:
This is pretty much possible now, beside fetching line and column numbers!

bdkjones:
Any updates on this? I'm looking to access the filename and line numbers in my custom function that handles @warn and @debug statements.

xzyfer:
We have no plans to further this until we reach Sass compatibility it 3.4. This increases the maintenance surface area and our hands are full atm.

mgreter:
Some ground work was already done and included in libsass, but not that we could yet craft a good and future proof API (we need to further refactor how we internally account for these states).

bdkjones:
Could it potentially be easier to simply have @warn and @debug handled by Libsass itself, rather than rely on the custom function approach?

99% of people are just going to write functions that log the message, line number and file of these warnings, right? Since there's no real customization to be done, what benefit is there in relying on the custom function approach? Does it maybe just make more sense to have Libsass log the output like it does for any other error?

mgreter:
Both warn and debug already take the same path as other custom functions. So no, it would rather be more difficult to handle warn etc. different that other custom functions.

Quoting from corresponding node-sass issue: https://github.com/sass/node-sass/issues/895#issuecomment-218315504
I don't think that sass_compiler_get_last_import will work reliably for custom functions. I stated before that the main problem is that libsass does not preserve the information needed ATM. The problem is that we read/parse/resolve the input/import files into an AST tree that will contain a function call op. Once we evaluate/execute that function call, the context of the actual import is no longer available (we only have one ast tree). That is what sass/libsass#1037 is all about. As I also wrote above IMO there is currently only one way to actually get that info via fb->pstate().path already. But that would be just a "crutch" for now, as the import stack should be available to custom functions (as other stuff). But nobody yet came around to actually solve/implement the missing pieces for libsass.

[...]

PR #2239, opened 2016-12-07
C-API: Add call stack queries (functions and mixins)

mgreter:
Fixes https://github.com/sass/libsass/issues/1037 and https://github.com/sass/libsass/issues/1950

## Corresponding commits in [perl-libsass][1]:

- [import stack queries][2]
- [call stack queries (functions and mixins)][3]
- [sass variable environment access][4]

[1]: https://github.com/sass/perl-libsass/commits/latest
[2]: https://github.com/sass/perl-libsass/commit/00ec13fb1877268c22d091af01bd6249d87508d1
[3]: https://github.com/sass/perl-libsass/commit/196fbd310dc28f9b0b6aca105c0ac68e243f9149
[4]: https://github.com/sass/perl-libsass/commit/2a3fa42d00746a59c8cf8a708267d1b6c6b0229c

### test.pl
```pl
use strict;
use warnings;

use CSS::Sass qw(get_callee_stack_size);
use CSS::Sass qw(get_callee_type);
use CSS::Sass qw(get_callee_name);
use CSS::Sass qw(get_callee_path);
use CSS::Sass qw(get_callee_line);
use CSS::Sass qw(get_callee_column);
use CSS::Sass qw(get_callee_var);
use CSS::Sass qw(set_callee_var);

my $sass = CSS::Sass->new(
  sass_functions => {
    'get-callee-stack-size()' => sub { get_callee_stack_size() },
    'get-callee-type($back : 0)' => sub { get_callee_type($_[0]) },
    'get-callee-name($back : 0)' => sub { get_callee_name($_[0]) },
    'get-callee-path($back : 0)' => sub { get_callee_path($_[0]) },
    'get-callee-line($back : 0)' => sub { get_callee_line($_[0]) },
    'get-callee-column($back : 0)' => sub { get_callee_column($_[0]) },
    'get-callee-var($name, $back : 0)' => sub { get_callee_var($_[0], $_[1]) },
    'set-callee-var($name, $val, $back : 0)' => sub { set_callee_var($_[0], $_[1], $_[2]) },
    'get-callee($back: 0)' => sub {
      return sprintf("(%d) %s at %s (%d/%d)",
        get_callee_type($_[0]),
        get_callee_name($_[0]),
        get_callee_path($_[0]),
        get_callee_line($_[0]),
        get_callee_column($_[0]),
      );
    }
  }
);

print $sass->compile_file("test.scss");
```

### test.scss
```scss
@function baz($back: 0) {
  @return get-callee($back);
}

[...]

PR #2251, opened 2016-12-18
C-API: Add call stack queries (functions and mixins)

mgreter:
Same as https://github.com/sass/libsass/pull/2239 but rebased and adjusted after memory refactoring

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -667,7 +667,7 @@ namespace Sass {
 
     // clean up by removing empty placeholders
     // ToDo: maybe we can do this somewhere else?
-    Remove_Placeholders remove_placeholders(*this);
+    Remove_Placeholders remove_placeholders;
     root->perform(&remove_placeholders);
     // return processed tree
     return root;
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -87,8 +87,9 @@ namespace Sass {
 
   {
 
-    // add cwd to include paths
-    include_paths.push_back(CWD);
+    // Sass 3.4: The current working directory will no longer be placed onto the Sass load path by default.
+    // If you need the current working directory to be available, set SASS_PATH=. in your shell's environment.
+    // include_paths.push_back(CWD);
 
     // collect more paths from different options
     collect_include_paths(c_options.include_path);
--- a/src/file.hpp
+++ b/src/file.hpp
@@ -4,6 +4,7 @@
 #include <string>
 #include <vector>
 
+#include "sass/context.h"
 #include "ast_fwd_decl.hpp"
 
 namespace Sass {
@@ -47,9 +48,16 @@ namespace Sass {
     std::string abs2rel(const std::string& path, const std::string& base = ".", const std::string& cwd = get_cwd());
 
     // helper function to resolve a filename
+    // searching without variations in all paths
+    std::string find_file(const std::string& file, struct Sass_Compiler* options);
     std::string find_file(const std::string& file, const std::vector<std::string> paths);
-    // inc paths can be directly passed from C code
-    std::string find_file(const std::string& file, const char** paths);
+
+    // helper function to resolve a include filename
+    // this has the original resolve logic for sass include
+    std::string find_include(const std::string& file, const std::vector<std::string> paths);
+
+    // split a path string delimited by semicolons or colons (OS dependent)
+    std::vector<std::string> split_path_list(const char* paths);
 
     // try to load the given filename
     // returned memory must be freed
@@ -113,7 +121,10 @@ namespace Sass {
 
   namespace File {
 
-    std::vector<Include> resolve_includes(const std::string& root, const std::string& file);
+    static std::vector<std::string> defaultExtensions = { ".scss", ".sass", ".css" };
+
+    std::vector<Include> resolve_includes(const std::string& root, const std::string& file,
+      const std::vector<std::string>& exts = defaultExtensions);
 
   }
 
--- a/src/sass.cpp
+++ b/src/sass.cpp
@@ -7,6 +7,23 @@
 #include "sass.h"
 #include "file.hpp"
 #include "util.hpp"
+#include "sass_context.hpp"
+#include "sass_functions.hpp"
+
+namespace Sass {
+
+  // helper to convert string list to vector
+  std::vector<std::string> list2vec(struct string_list* cur)
+  {
+    std::vector<std::string> list;
+    while (cur) {
+      list.push_back(cur->string);
+      cur = cur->next;
+    }
+    return list;
+  }
+
+}
 
 extern "C" {
   using namespace Sass;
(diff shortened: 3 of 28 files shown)


mgreter:
Fixes https://github.com/sass/libsass/issues/1037 and https://github.com/sass/libsass/issues/1950

## Corresponding commits in [perl-libsass][1]:

- [import stack queries][2]
- [call stack queries (functions and mixins)][3]
- [sass variable environment access][4]

[1]: https://github.com/sass/perl-libsass/commits/latest
[2]: https://github.com/sass/perl-libsass/commit/00ec13fb1877268c22d091af01bd6249d87508d1
[3]: https://github.com/sass/perl-libsass/commit/196fbd310dc28f9b0b6aca105c0ac68e243f9149
[4]: https://github.com/sass/perl-libsass/commit/2a3fa42d00746a59c8cf8a708267d1b6c6b0229c

## Sample perl program

### test.pl
```pl
use strict;
use warnings;

use CSS::Sass qw(get_callee_stack_size);
use CSS::Sass qw(get_callee_type);
use CSS::Sass qw(get_callee_name);
use CSS::Sass qw(get_callee_path);
use CSS::Sass qw(get_callee_line);
use CSS::Sass qw(get_callee_column);
use CSS::Sass qw(get_callee_var);
use CSS::Sass qw(set_callee_var);

my $sass = CSS::Sass->new(
  sass_functions => {
    'get-callee-stack-size()' => sub { get_callee_stack_size() },
    'get-callee-type($back : 0)' => sub { get_callee_type($_[0]) },
    'get-callee-name($back : 0)' => sub { get_callee_name($_[0]) },
    'get-callee-path($back : 0)' => sub { get_callee_path($_[0]) },
    'get-callee-line($back : 0)' => sub { get_callee_line($_[0]) },
    'get-callee-column($back : 0)' => sub { get_callee_column($_[0]) },
    'get-callee-var($name, $back : 0)' => sub { get_callee_var($_[0], $_[1]) },
    'set-callee-var($name, $val, $back : 0)' => sub { set_callee_var($_[0], $_[1], $_[2]) },
    'get-callee($back: 0)' => sub {
      return sprintf("(%d) %s at %s (%d/%d)",
        get_callee_type($_[0]),
        get_callee_name($_[0]),
        get_callee_path($_[0]),
        get_callee_line($_[0]),
        get_callee_column($_[0]),
      );
    }
  }
);

print $sass->compile_file("test.scss");
```

### test.scss
```scss
@function baz($back: 0) {
  @return get-callee($back);
}

@function bar($back: 0) {
  @return baz($back);
}

@function foo($back: 0) {
  @return bar($back);
}

@mixin foo($back: 0) {
  @content;
}

A {
  $foo: "foo";
  @include foo {
    callee-5: foo(5);
    callee-4: foo(4);
    callee-3: foo(3);
    callee-2: foo(2);
    callee-1: foo(1);
    callee-0: foo(0);
    foo: get-callee-var("$foo");
    foo: set-callee-var("$foo", "new");
    foo: get-callee-var("$foo");
  }
}
```

### result
```css
A {
  callee-5: (0) foo at D:/.../perl-libsass/test.scss (19/12);
  callee-4: (0) @content at D:/.../perl-libsass/test.scss (14/3);
  callee-3: (1) foo at D:/.../perl-libsass/test.scss (21/15);
  callee-2: (1) bar at D:/.../perl-libsass/test.scss (10/11);
  callee-1: (1) baz at D:/.../perl-libsass/test.scss (6/11);
  callee-0: (2) get-callee at D:/.../perl-libsass/test.scss (2/11);
  foo: foo;
  foo: new; }
```

src/context.cpp
```diff
@@ -87,8 +87,9 @@ namespace Sass {
 
   {
 
-    // add cwd to include paths
-    include_paths.push_back(CWD);
+    // Sass 3.4: The current working directory will no longer be placed onto the Sass load path by default.
```
nschonni: Is it work wrapping in a complier directive for the SASS_LANG = 3.4?

src/file.hpp
```diff
@@ -113,7 +121,10 @@ namespace Sass {
 
   namespace File {
 
-    std::vector<Include> resolve_includes(const std::string& root, const std::string& file);
+    static std::vector<std::string> defaultExtensions = { ".scss", ".sass", ".css" };
```
nschonni: Why include CSS files?

src/sass.cpp
```diff
@@ -49,11 +66,52 @@ extern "C" {
     return sass_copy_c_string(unquoted.c_str());
   }
 
+  char* ADDCALL sass_compiler_find_include (const char* file, struct Sass_Compiler* compiler)
+  {
+    // get the last import entry to get current base directory
+    // struct Sass_Options* options = sass_compiler_get_options(compiler);
```
nschonni: Dead code?

src/sass.cpp
```diff
@@ -49,11 +66,52 @@ extern "C" {
...
+    paths.insert( paths.end(), incs.begin(), incs.end() );
+    // now resolve the file path relative to lookup paths
+    std::string resolved(File::find_include(file, paths));
+    return sass_copy_c_string(resolved.c_str());
+  }
+
+  char* ADDCALL sass_compiler_find_file (const char* file, struct Sass_Compiler* compiler)
+  {
+    // get the last import entry to get current base directory
+    // struct Sass_Options* options = sass_compiler_get_options(compiler);
```
nschonni: Dead code?

src/context.cpp
mgreter: Not sure. I was under the impression current master is going for 3.5. Side note: we could use a runtime test for 3.4 vs 3.5 behavior, but a compiler directive is not simple, since we have a string in that constant (string compare is not really supported by `#if` preprocessor directives).

src/file.hpp
mgreter: Same behavor [as before][1], also see https://github.com/sass/libsass/pull/754
[1]: https://github.com/sass/libsass/blob/master/src/file.cpp#L308-L310

src/sass.cpp
mgreter: Yep, commented due to unused var compiler warning.

src/sass.cpp
mgreter: Yep, commented due to unused var compiler warning.

mgreter:
Needs a rebase after https://github.com/sass/libsass/pull/2267 is merged!

mgreter:
## A few C-API remarks 

The C-API has grown quite a bit and we now expose quite a few arrays. Implementation for them is often very different. Since the whole C-API is behind function calls and pointers now, we can use C++ behind those functions. Therefore we could just use `std::vector` instead of our own primitive implementations (as i.e. for `include_paths`, which is really a linked list in order to grow dynamically). But due to where we came from, most things on the C-API are still just structs (which are allocated via calloc and deleted via free). In order to support C++ containers, those structs need to be upgraded to real classes and allocated via new and freed via delete (otherwise we leak). So this task would basically boil down to design and implement a C++ API for LibSass (which would be a nice addon).

## More CI tests (plugins)

I've added my libsass plugins to the CI process (Travis only so far). This allows most of the C-API to be tested via CI. I created a specific [libsass-tests][1] plugin to expose most C-API features as custom functions (work in progress). I was hoping to include a small sass-spec suite for each plugin, but unfortunately ran against https://github.com/sass/sass-spec/issues/1025. So for now each plugin contains just one spec test that is run and compared via diff in libsass Travis CI.

For reference, these are the tested plugins:

- https://github.com/mgreter/libsass-glob
- https://github.com/mgreter/libsass-math
- https://github.com/mgreter/libsass-digest*
- https://github.com/mgreter/libsass-tests

*digest needs this PR to compile/work!

[1]: https://github.com/mgreter/libsass-tests

src/context.cpp
mgreter: Just realized that this change was done in ruby sass 3.4, not 3.5, therefore I think we should already have that removed. So to speak this is a bug in libsass 3.4. Maybe Or I did not get your comment?

nschonni: You're right that this is 3.5 now, but I was thinking for ease of backporting changes to 3.4 while they are still both being maintained. I guess it would need to be a plain definition of `SASS_LANGUAGE_3_4`

src/file.hpp
nschonni: You're right, but isn't it going away by 4 though? Might be another one to conditional out

src/context.cpp
mgreter: I created a seperate commit for it, so it's better visible in this PR (for cherry-pick). IMO implementers are still free to add this on their own if the want to keep compatibility.

src/file.hpp
mgreter: IMO it should move into C-API to be configurable. This is just a small step in that direction.

src/context.cpp
mgreter: This is actually a bugfix as ruby sass 3.4 removed this behavior and we did not yet ...

xzyfer: I understand, however it's a breaking change and 3.4 is stable. The ship has sailed on breaking changes in 3.4. Back ports are only happening for small Sass fixes.

typoworx-de:
How to use this in js using node-sass?

Issue #2056, opened 2016-04-27
Generic wrapped selector should only accept pseudo expressions


mgreter:
Continued from https://github.com/sass/libsass/issues/2051

``` scss
:foobar(.baz) {
    color: red;
}
```

libsass and ruby sass 3.3.14

``` css
:foobar(.baz) {
  color: red; }
```

ruby sass 3.4.21

``` css
Invalid CSS after ":foobar(": expected pseudo_expr, was ".baz)"
```

Expected pseudo expressions seem to be at least `identifier` and `2n+1`.
This is probably different for specific wrapped selector as `:not` and `:has`.
Fixing this will affect users that rely on the old ruby sass 3.3 behavior :exclamation: 

Spec added https://github.com/sass/sass-spec/pull/914
##

mgreter:
@chriseppstein hope it's ok to cc you here, but you might be able to come up with a few good spec tests to cover the parsing of pseudo selectors here!? Once we have that it should be quite simple to implement. It looks like an error we can directly raise in the parser.

xzyfer:
@mgreter the value for a pseudo selector is one or more of the following optionally separated by spaces/comments.

```
lex< one_plus< alternatives<
  exactly<'+'>, 
  exactly<'-*'>,  
  quoted_string, 
  identifier,
  sequence<
    unsigned_number,
    optional< sequence<
      alternatives< exactly<'e'>, exactly<'E'> >,
      optional<sign>,
      digits, 
    > >,
    optional<unit_identifier>
  >
> > >()
```

nex3:
This is no longer the intended behavior; see sass/sass#2120.

xzyfer:
@nex3 we'll need to figure a plan for how to handle sass-spec in these cases. It turns out @mgreter has incidentally "fixed" this issue as a side affect of other extends work #2253.

I suggest the relevant specs are updated to match the expected behaviour and ruby-sass todo'd.

nex3:
> I suggest the relevant specs are updated to match the expected behaviour and ruby-sass todo'd.

Done in sass/sass-spec@b92a53b19a5ac029f24cb156765fa9c6b1be00d6.

PR #2253, opened 2016-12-19
Fixes performance bottleneck in subset_map/extend

mgreter:
You can expect to see a 2x performance increase in bigger sass code bases. With more extend usage it can go as high as 10x performance increase.

Test with styles of a real corporate website showed a performance increase of 3x (from 750ms to 250ms). When I imported the main stylesheet 10x times, compile time went from 100s to 5s (~20x increase).

This indecently (according to probe-todo) also fixes https://github.com/sass/libsass/issues/2056 (false positive?)?

### DIFF
--- a/Makefile.conf
+++ b/Makefile.conf
@@ -43,6 +43,7 @@ SOURCES = \
 	to_c.cpp \
 	to_value.cpp \
 	source_map.cpp \
+	subset_map.cpp \
 	error_handling.cpp \
 	memory/SharedPtr.cpp \
 	utf8_string.cpp \
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -30,6 +30,23 @@ namespace Sass {
            SASS_MEMORY_CAST(Supports_Operator, cond);
   }
 
+  size_t HashExpression::operator() (Expression_Obj ex) const {
+    return ex ? ex->hash() : 0;
+  }
+
+  size_t HashSimpleSelector::operator() (Simple_Selector_Obj ex) const {
+    return ex ? ex->hash() : 0;
+  }
+
+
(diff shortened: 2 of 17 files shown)


xzyfer:
It's noting that the expected behaviour of #2056 has changed. The bug is now incorrect, see the final comments.

xzyfer:
Nice one. This approach is fine by me. It's worth nothing that the extends behaviour has changed slightly in dart-sass, whose behaviour is going to be the reference implementation.

mgreter:
As stated, this looks like a false positive (therefore meaning probe-todo is still broken) ...

xzyfer:
Yeah. Just wanted to make sure you were aware of the difference between Ruby Sass and Dart Sass since you're working on the extends code.

mgreter:
FWIW: made some performance assesments accross different implementations:

```scss
.clearfix {
  &:after {
    content: "";
    display: table;
    clear: both;
  }
}

@for $n from 1 through 100 {
	@for $i from 1 through 100 {
		.foo#{$i} { @extend .clearfix; }
	}
}
```

```cmd
$ bin/sassc.exe -v
sassc: 3.4.1
libsass: 3.4.0-19-g414b
sass2scss: 1.1.0
sass: 3.4
$ time sassc test.scss
real    0m0.279s
user    0m0.000s
sys     0m0.015s
```

```cmd
$ sass -v
Sass 3.4.23 (Selective Steve)
$ time sass test.scss
real    0m2.560s
user    0m1.984s
sys     0m0.561s
```

```cmd
$ dart-sass --version
1.0.0-alpha.6
$ time dart-sass test.scss
real    0m45.833s
user    0m0.000s
sys     0m0.000s
```

//CC @xzyfer @nex3

Issue #2261, opened 2016-12-23
Incorrect Comparison of Inline Interpolated String Values


epicmiller:
### LibSass incorrectly compares inline Interpolated string values

input.scss
```scss
$seven: 7;
$a: '#{7}';
$b: '#{7}';

.test {
  equal: ($a == $b);
  equal: ('#{7}' == '#{7}');
  equal: ('#{$seven}' == '#{$seven}');
}
```

[libsass 3.4.0][1]
```css
.test {
  equal: true;
  equal: false;
  equal: false;
}
```

[ruby sass 3.4.21][2]
```css
.test {
  equal: true;
  equal: true;
  equal: true;
}
```

version info:
```cmd
$ node-sass --version
node-sass       4.1.0   (Wrapper)       [JavaScript]
libsass         3.4.0   (Sass Compiler) [C/C++]
```

[1]: http://libsass.ocbnet.ch/srcmap/#JHNldmVuOiA3OwokYTogJyN7N30nOwokYjogJyN7N30nOwoKLnRlc3QgewogIGVxdWFsOiAoJGEgPT0gJGIpOwogIGVxdWFsOiAoJyN7N30nID09ICcjezd9Jyk7CiAgZXF1YWw6ICgnI3skc2V2ZW59JyA9PSAnI3skc2V2ZW59Jyk7Cn0=
[2]: http://www.sassmeister.com/gist/8b02108feae47783d125efd722a0a050

mgreter:
Thanks for your report. This will be fixed with https://github.com/sass/libsass/pull/2262.

PR #2262, opened 2016-12-24
Fixes equality compare evaluation issue

mgreter:
Fixes https://github.com/sass/libsass/issues/2261


xzyfer:
Spec https://github.com/sass/sass-spec/pull/1022

Issue #2243, opened 2016-12-12
Support url() with quotes in unknown directives


nex3:
Currently, libsass fails to parse a `url()`-style string with quotes when it appears in an unknown directive. For example:

```scss
@namespace url("http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul");
```

produces:

```
Error: Invalid CSS after '...pace url("foo");': expected "{", was ""
        on line 1 of stdin
>> @namespace url("foo");
   --------------------^
```

See sass/sass#2188

PR #2263, opened 2016-12-25
Fix parsing of url with quoted argument

mgreter:
Fixes https://github.com/sass/libsass/issues/2243


xzyfer:
Spec https://github.com/sass/sass-spec/pull/1027

xzyfer:
Backported to 3.4

Issue #2260, opened 2016-12-20
3.4: @content inside @at-root block throws an error


TxHawks:
### Title: 3.4: @content inside @at-root block throws an error

Including a `@content` statement inside an `@at-root` block (inside a mixin, of course) throws an error. This worked fine in `3.3.6`.

Here is a reduced test case:

input.scss
```scss
@mixin test() {
  @at-root {
    @content;
  }
}

@include test {
  .test {
    property: value;
   }
 }
```

libsass 3.4.0:
Throws an error:
```sh
Error: Base-level rules cannot contain the parent-selector-referencing character '&'.

  Backtrace:
    input.scss:8, in mixin `@content`
    input.scss:3, in mixin `test`
    input.scss:7
 on line 8 of input.scss:7
```
ruby sass 3.4.21
```css
.test {
  property: value;
}
```

version info:
```sh
$ node-sass --version
node-sass       4.1.1   (Wrapper)       [JavaScript]
libsass         3.4.0   (Sass Compiler) [C/C++]
```

xzyfer:
Spec https://github.com/sass/sass-spec/pull/1023

PR #2264, opened 2016-12-25
Fix error case for parent compound selectors check

mgreter:
Fixes https://github.com/sass/libsass/issues/2260


xzyfer:
https://github.com/sass/sass-spec/pull/1029

xzyfer:
Back ported to 3.4

PR #2265, opened 2016-12-28
Update test bootstrap script to use document environment vars

xzyfer:
We have documented `SASS_SPEC_PATH` and `SASS_SASSC_PATH` environment
variables for user defined sass-spec and sassc locations respectively.

The bootstrap script is currently ignoring these.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -2,9 +2,12 @@
 
 script/branding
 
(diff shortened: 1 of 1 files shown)


xzyfer:
Back ported to 3.4. in 606a4c2

Issue #2224, opened 2016-11-03
SourceMaps lines offset for interpolated selectors 


rodneyrehm:
The SourceMaps generation seems to trip over multi-line selectors if they contain interpolations (`#{$variable}`). In the demo SCSS below the generated CSS `.zulu-delta` is referenced to line 17, while it should be line 19. This does not happen for non-interpolated multi-line selectors. A single interpolation in the selector string causes the lines to collapse before the mutation could be tracked in SourceMaps. The offset is not recovered from, so all SCSS content *after* such an interpolated selector is affected by the shift in line number.

input.scss ([SourceMap](http://libsass.ocbnet.ch/srcmap/#JHByZWZpeDogenVsdTsKCi5hbHBoYSwKLmJyYXZvLAouY2hhcmxpZSB7CiAgY29sb3I6IHJlZDsKfQoKLmRlbHRhIHsKICBjb2xvcjogYmx1ZTsKfQoKLiN7JHByZWZpeH0tYWxwaGEsCi4jeyRwcmVmaXh9LWJyYXZvLAouI3skcHJlZml4fS1jaGFybGllIHsKICBjb2xvcjogcmVkOwp9CgouI3skcHJlZml4fS1kZWx0YSB7CiAgY29sb3I6IGJsdWU7Cn0KCi5lY2hvIHsKICBjb2xvcjogZ3JlZW47Cn0=))
```scss
$prefix: zulu;

.alpha,
.bravo,
.charlie {
  color: red;
}

.delta {
  color: blue;
}

.#{$prefix}-alpha,
.#{$prefix}-bravo,
.#{$prefix}-charlie {
  color: red;
}

.#{$prefix}-delta {
  color: blue;
}

.echo {
  color: green;
}
```

version info:
```cmd
$ node-sass --version
node-sass	3.11.0	(Wrapper)	[JavaScript]
libsass  	3.3.6	(Sass Compiler)	[C/C++]
```

mgreter:
Thanks for the thorough bug report. I can see the issue clearly in the sourcemap inspector.

mgreter:
@rodneyrehm can you tell on which versions you saw this error. Wondering if it got introduces with some latest sourcemap changes in 3.4-RC1. I specially would like to know if 3.3.6 already had this issue!? Thx!

rodneyrehm:
>  can you tell on which versions you saw this error.

[...]

PR #2268, opened 2016-12-29
Fix source mappings for selector schema

mgreter:
Addresses https://github.com/sass/libsass/issues/2224, but does not fully solve it.

Underlying problem is that the selector schema interpolation is first evaluated into a string and then passed to the parser again to produce the final AST tree. This patch should fix the most urgent error that following lines can be offset when the selector schema has more lines than the evaluated string.

We still output pretty much garbage mappings for the parts in the selector schema itself. I think it should theoretically be possible to re-map the mappings after the second parsing, but I wasn't able to get this working so far. But with this patch following lines should have the correct mappings.

Not perfect, but still much better than before!

### DIFF
--- a/src/debugger.hpp
+++ b/src/debugger.hpp
@@ -663,6 +663,7 @@ inline void debug_ast(AST_Node_Ptr node, std::string ind, Env* env)
     std::cerr << ind << "String_Schema " << expression;
     std::cerr << " (" << pstate_source_position(expression) << ")";
     std::cerr << " " << expression->concrete_type();
+    std::cerr << " (" << pstate_source_position(node) << ")";
     if (expression->is_delayed()) std::cerr << " [delayed]";
     if (expression->is_interpolant()) std::cerr << " [is interpolant]";
     if (expression->has_interpolant()) std::cerr << " [has interpolant]";
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -145,7 +145,9 @@ namespace Sass {
     }
 
     // update for end position
-    block->update_pstate(pstate);
+    // this seems to be done somewhere else
+    // but that fixed selector schema issue
+    // block->update_pstate(pstate);
 
     // parse comments after block
     // lex < optional_css_comments >();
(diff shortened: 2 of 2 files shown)


xzyfer:
An alternative ideal solution, and the one I recommend delay the selector parsing to the eval stage all together. This is what Ruby Sass does. During parsing they parse the selector a single thunk, then do the actual selector parsing during eval.

This approach also solves a couple other issues with at-root and parent selectors as discussed briefly in #2199

mgreter:
That would not have helped in any way here, as the (delayed) selector schema parsing was exactly the one that was broken in regard to source-maps. We may need to loose the inner mappings for selectors that are re-parsed (as ruby sass does), but I have not yet come to a final conclusion here. FWIW: IMO the core issue with at-root was that the following was never anticipated:

```scss
foo {
  @at-root {
    bar {
	  baz: baz;
	}
  }
}

foo {
  @at-root {
    & bar {
	  baz: baz;
	}
  }
}
```

Results in:
```css
bar {
  baz: baz; }
foo bar {
  baz: baz; }
```

As one can see, the explicit parent selector (`&`) seems to reach accross at-root rules. This pretty much explains all the headaches we had with parent selectors.

xzyfer:
> As one can see, the explicit parent selector (&) seems to reach accross at-root rules

As I'm sure you've figured out, what's happening here is that `&` is being resolved in eval.

```css
foo {
  @at-root {
    foo bar {
      baz: baz;
    }
  }
}
```

Then the at-root is resolved in cssize. The `&` isn't reaching across the at-root. The root of problems we've had with `&` is that we've been resolving it at the wrong time which is what #2199 was addressing.

mgreter:
@xzyfer I did not get your point? Your sample does not contain any `&`?

Please see this example closely:
```scss
root {
  @at-root {
    & foo bar {
      baz: baz;
    }
  }
}
```

This results in ruby sass:
```css
root foo bar {
  baz: baz; }
```

Which is what I mean by "reaching across at-root"! This exactly explains why we needed the differentiation of explicit and implicit parent selectors!

Issue #2095, opened 2016-05-26
@mixin inside @media (compiles with Sass v3.4.21)


glen-84:
``` sass
@media all {
  @mixin foo() {}
}
```

Ruby Sass

``` css
```

LibSass

```
Error: Mixins may not be defined within control directives or other mixins.
        on line 75 of test.scss
>>   @mixin foo() {}
   --^
```

Spec added https://github.com/sass/sass-spec/pull/911

xzyfer:
Thanks @glen-84 I've confirmed this is a bug with defining mixins within @media. The work around for now is moving the `@mixin` outside the `@media`.

Minimal test case
```sass
@media all {
  @mixin foo() {}
}
```

PR #2269, opened 2016-12-29
Fix wrong error on valid mixin in media rule

mgreter:
Fixes https://github.com/sass/libsass/issues/2095

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -364,7 +364,7 @@ namespace Sass {
   Definition_Obj Parser::parse_definition(Definition::Type which_type)
   {
     Scope parent = stack.empty() ? Scope::Rules : stack.back();
-    if (parent != Scope::Root && parent != Scope::Rules && parent != Scope::Function) {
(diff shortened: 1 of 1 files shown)


xzyfer:
This check belongs in the check nesting visitor. This entire check should be removed from the parser.

src/parser.cpp
```diff
@@ -364,7 +364,7 @@ namespace Sass {
   Definition_Obj Parser::parse_definition(Definition::Type which_type)
   {
     Scope parent = stack.empty() ? Scope::Rules : stack.back();
-    if (parent != Scope::Root && parent != Scope::Rules && parent != Scope::Function) {
+    if (parent != Scope::Root && parent != Scope::Rules && parent != Scope::Function && parent != Scope::Media) {
```
xzyfer: This belongs in the check nesting visitor for now. The parser should avoid validating the AST tree.

xzyfer: I'm looking into why the check_nesting visitor is bailing out early here. It's a direct port from Ruby Sass so tracking down the issue should be fast.

xzyfer:
Superseded by #2270

PR #2270, opened 2016-12-29
Fix incorrect validation of mixin definition

### LINKED ISSUE #2095: @mixin inside @media (compiles with Sass v3.4.21)

xzyfer:
Removed the duplicate validation check from the parser. The validation
in the check nesting visitor was being skipped.

Looks like there is a bunch of left of AST validation in the parser,
which now belongs in the check nesting visitor. With the current
approach the parser should only ever error when it encounters invalid
tokens. It cannot accurately validate the AST because things like
`at-root` change the structure in important ways.

Closes #2269
Fixes #2095
Spec sass/sass-spec#1031

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -11,13 +11,8 @@ namespace Sass {
     current_mixin_definition(0)
   { }
 
-  Statement_Ptr CheckNesting::before(Statement_Ptr s) {
-      if (this->should_visit(s)) return s;
-      return NULL;
-  }
-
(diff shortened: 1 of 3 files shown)


xzyfer:
Back ported to 3.4 in 1b7b8dd7c797fc51cff96a6895fcaeb02d08a417

PR #2271, opened 2016-12-29
Remove left over AST validation from the parser

xzyfer:
Continuing on from #2270.

This PR removes some AST validation in the parser left over
from implementing the check nesting visitor in #2062. This
also address a bug that caused some validations to be
skipped.

There's one final AST validation left in the parser which is
for `@import`. Unfortunately due to how we handle 
`@import` nodes there isn't a straight forward way to fix
the validation issues. I'm leaving that alone until I overhaul
`@import` for the module system.

### DIFF
--- a/src/check_nesting.cpp
+++ b/src/check_nesting.cpp
@@ -82,7 +82,10 @@ namespace Sass {
   Statement_Ptr CheckNesting::operator()(Definition_Ptr n)
   {
     if (!this->should_visit(n)) return NULL;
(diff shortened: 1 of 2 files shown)


xzyfer:
Back ported to 3.4 in d301570e6536f251b9a2d226b2c0ae7ccda7cbf1 and b5fdde7747c5db3a9bb0ba4e8b32bce56d2c7505

PR #2276, opened 2016-12-31
Add deprecation notice for executing call() with a string

### LINKED ISSUE #2275: Calling call() with a string should produce a deprecation notice
This is part of sass/sass#2061. Starting in 4.0 `call` will require a function reference obtained via `get-function()`.

Specs are in https://github.com/sass/sass-spec/pull/867/commits/506e3352c1ff5c196032c118f2cc1065a71431e3 as part of https://github.com/sass/sass-spec/pull/867

xzyfer:
This is part of sass/sass/pull/2061. Starting in 4.0 `call` will
require a function reference obtained via `get-function()`.

Fixes #2275
Spec sass/sass-spec#1044

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -1652,9 +1652,17 @@ namespace Sass {
     Signature call_sig = "call($name, $args...)";
     BUILT_IN(call)
     {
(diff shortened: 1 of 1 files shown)


xzyfer:
Merging into #2373 since all the specs are in the same folder.

PR #2282, opened 2017-01-02
Add list is_bracketed support to the C API

xzyfer:
Lists gained an `is_bracketed` attribute in #2284. The primary 
semantic difference is that the`is_bracketed` changes list equality 
(#2281) which can matter in custom functions.

This is a breaking change so I'd like to get it into the first 3.5
beta.

/cc 
@nex3 to sanity check this actually matters for custom functions
@mgreter because you most about the C API
@chriseppstein because of the implication for eyeglass

### DIFF
--- a/docs/api-value.md
+++ b/docs/api-value.md
@@ -58,7 +58,7 @@ union Sass_Value* sass_make_string  (const char* val);
 union Sass_Value* sass_make_qstring (const char* val);
 union Sass_Value* sass_make_number  (double val, const char* unit);
 union Sass_Value* sass_make_color   (double r, double g, double b, double a);
-union Sass_Value* sass_make_list    (size_t len, enum Sass_Separator sep);
+union Sass_Value* sass_make_list    (size_t len, enum Sass_Separator sep, bool is_bracketed);
 union Sass_Value* sass_make_map     (size_t len);
 union Sass_Value* sass_make_error   (const char* msg);
 union Sass_Value* sass_make_warning (const char* msg);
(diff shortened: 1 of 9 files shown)


xzyfer:
I've thought about bracketed lists a bit more. I'm going to drop to
delimiter all together and just tagged the list as bracketed or not. In
which case maybe having a different make function for bracketed lists make
sense. It'll still return a Sass_List just with the bracketed bool set.

I have Wip passing tests for dropping the delimiter I'll push this week.

xzyfer:
>I'm not too excited about the term "delimiter" here

This has been addressed in #2284. The delimiter `attribute` was dropped in favour of the more semantic `is_bracketed` property which is what the delimiter `attribute` was being used to infer.

>There would be a way to make this a "non" breaking change.

I'm fine with making this a breaking a change as long it's document the change accordingly, which they will be. API breakages were expected in 3.5 and 4.0. 

>I guess bracketed lists will be a seldomly used feature?

I disagree. Given the sheer amount of grid "frameworks" that exist, and the number of new ones that came out for flexbox, I expect this feature to get a decent workout. We shouldn't get in the way of experimentation.

>IMO there are a few other flags that are not exposed via the constructor directly.

Separator, and bracket are core semantics of a list. The same isn't true for most other list attributes which are largely implementation details of LibSass.

>Alternatively there could also be an explicit constructor for bracketed lists. I.e. sass_make_array;

I don't think arrays are right here. These are lists. They share all the same semantics as lists.

mgreter:
> I'm fine with making this a breaking a change as long it's document the change accordingly, which they will be. API breakages were expected in 3.5 and 4.0.

Then you should also increase ABI version number as AFAIK this is the first time we break ABI.

I guess you'll add the documentation also to this PR?

xzyfer:
Update the documentation and bumped the ABI version to 1:0:0

am11:
@xzyfer responded with a PR https://github.com/sass/libsass/pull/2288 😎

Issue #2081, opened 2016-05-11
Should error on unquote ; inside interpolation


xzyfer:
Discovered from #2080

Sass

``` scss
$foo: #{bar();};
```

LibSass 3.3.6

``` css
```

Ruby Sass 3.4.21

``` css
Invalid CSS after "$foo: #{bar()": expected "}", was ";};"
```

Spec added https://github.com/sass/sass-spec/pull/913
##

PR #2285, opened 2017-01-04
Fix error case with superfluous data in interpolates

mgreter:
Fixes https://github.com/sass/libsass/issues/2081


xzyfer:
Back ported to 3.4 in 1e3c700

xzyfer:
Specs https://github.com/sass/sass-spec/pull/1046

PR #2286, opened 2017-01-05
 Fix thread-safety for mixin recursions check

mgreter:
Keeping the counter in a static exposed some weird behavior in https://github.com/sass/node-sass/issues/1669#issuecomment-270020822. This PR should fix that particular issue. I believe this is due to thread safety. Not sure how they do it, but the behavior we see is the following:

```c++
static unsigned int recursions = 0;
mixin() {
  if (recursions > 500) {
    throw "stack level too deep";
  }
  recursions ++;
  ... do more work
  ... may call mixin() again
  recursions --;
}
```

At the decrement, `recursions` at some point contains a `0`, therefore the decrement will overflow the unsigned value to 4294967295, which will then trigger the error check. For me this can only mean that the static variable got initialized again on the same thread. Moving the variable to the class will make sure that every context uses their own memory for this counter. Still, LibSass is not guaranteed or considered to be 100% thread safe (mostly due to the lack of anybody really verifying it).

### DIFF
--- a/src/expand.cpp
+++ b/src/expand.cpp
@@ -14,21 +14,21 @@
 namespace Sass {
 
   // simple endless recursion protection
-  const unsigned int maxRecursion = 500;
-  static unsigned int recursions = 0;
+  const size_t maxRecursion = 500;
 
   Expand::Expand(Context& ctx, Env* env, Backtrace* bt, std::vector<Selector_List_Obj>* stack)
   : ctx(ctx),
     eval(Eval(*this)),
+    recursions(0),
+    in_keyframes(false),
+    at_root_without_rule(false),
+    old_at_root_without_rule(false),
     env_stack(std::vector<Env*>()),
     block_stack(std::vector<Block_Ptr>()),
     call_stack(std::vector<AST_Node_Obj>()),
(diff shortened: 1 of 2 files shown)


mgreter:
A simpler reproduction of the error:

```js
var sass = require('node-sass');
function render() {
  sass.render({
    data: [
      "@mixin test2() {",
      "  foo: bar;",
      "}",
      "@mixin test() {",
      "  @for $i from 1 to 100000 {",
      "  }",
      "}",
      "foo {",
      "  @include test2;",
      "  @include test;",
      "  @include test2;",
      "}",
    ].join("\n"),
  }, function(error, result) {
    if (!result) console.log('error ', error);
  });
}

for (var i = 0; i < 100; i ++) {
  setTimeout(render, Math.random() * 1000);
}
```

FWIW: I always aim to be thread safe in regard of contexts. Multiple contexts/compilers should be safe to use even on the same thread, which is what node-sass obviousely does. Since we don't share anything between them, this should be quite easy to achieve, but nonetheless there is no real guarantee that no side effects hide in the source code! This one was really just badly implemented!

xzyfer:
Back ported to 3.4 in a7ef887

PR #2288, opened 2017-01-06
Update VS build facade

am11:
@xzyfer, @mgreter,

I am not well familiar with the syntax of resource files. Just hacked it using Visual Studio's RC editor and updated only few parts (note that I manually updated the .rc file using sublime as editing + saving with VS results in multiple files noise such as .db, .aps and whatnot..)

Added a `.SolutionItems` folder to VS solution with some repo files including the resource.rc, so it gets visible in Solution Explorer of VS.

Added .db and .aps to .gitignore.

### DIFF
--- a/res/resource.rc
+++ b/res/resource.rc
@@ -18,14 +18,14 @@ BEGIN
   BEGIN
     BLOCK "080904b0"
     BEGIN
-      VALUE "CompanyName", "Libsass Organization"
+      VALUE "CompanyName", "Sass Open Source Foundation"
       VALUE "FileDescription", "A C/C++ implementation of a Sass compiler"
-      VALUE "FileVersion", "0.9.0.0"
+      VALUE "FileVersion", "1.0.0.0"
       VALUE "InternalName", "libsass"
-      VALUE "LegalCopyright", "©2014 libsass.org"
+      VALUE "LegalCopyright", "\251 2017 libsass.org"
       VALUE "OriginalFilename", "libsass.dll"
-      VALUE "ProductName", "Libsass Library"
-      VALUE "ProductVersion", "0.9.0.0"
+      VALUE "ProductName", "LibSass Library"
+      VALUE "ProductVersion", "1.0.0.0"
     END
   END
   BLOCK "VarFileInfo"
--- a/win/libsass.sln
+++ b/win/libsass.sln
@@ -1,9 +1,20 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libsass", "libsass.vcxproj", "{E4030474-AFC9-4CC6-BEB6-D846F631502B}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".SolutionItems", ".SolutionItems", "{33318C77-2391-4399-8118-C109155A4A75}"
+	ProjectSection(SolutionItems) = preProject
+		..\.editorconfig = ..\.editorconfig
+		..\.gitattributes = ..\.gitattributes
+		..\.gitignore = ..\.gitignore
+		..\.travis.yml = ..\.travis.yml
+		..\appveyor.yml = ..\appveyor.yml
+		..\Readme.md = ..\Readme.md
+		..\res\resource.rc = ..\res\resource.rc
+	EndProjectSection
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ VERSION
 .cproject
 .project
 .settings/
+*.db
+*.aps
 
 # Configuration stuff
 


res/resource.rc
```diff
@@ -18,14 +18,14 @@ BEGIN
...
       VALUE "FileDescription", "A C/C++ implementation of a Sass compiler"
-      VALUE "FileVersion", "0.9.0.0"
+      VALUE "FileVersion", "1.0.0.0"
       VALUE "InternalName", "libsass"
-      VALUE "LegalCopyright", "©2014 libsass.org"
+      VALUE "LegalCopyright", "\251 2017 libsass.org"
       VALUE "OriginalFilename", "libsass.dll"
-      VALUE "ProductName", "Libsass Library"
-      VALUE "ProductVersion", "0.9.0.0"
+      VALUE "ProductName", "LibSass Library"
```
xzyfer: This has bugged me since I discovered this file :)

xzyfer:
@am11 you're my hero 🎉 

Are there any good resource on the recommended versioning schema for Windows, or is it fine to just mirror our libtool versioning?

am11:
@xzyfer, I just mirrored the version numbers. As far as I can tell, the version conventions are not set in stone and there is quite a lot of variety of versions in win32 libs. For instance, instead of *four-sequence identifier*, we can even use three sequences (semver style).

The RC file can read from other .h defines. I was thinking, if libtools and resource.rc share common .h file, that would make our life slightly easier. :)

Alternatively, we can use a bumpversion.sh or [extend git command](https://coderwall.com/p/bt93ia/extend-git-with-custom-commands), something like: `git sass-tag M.m.p` which will update these files, create tag and also create a commit -- exactly how `npm version` command works.

res/resource.rc
```diff
@@ -18,14 +18,14 @@ BEGIN
   BEGIN
     BLOCK "080904b0"
     BEGIN
-      VALUE "CompanyName", "Libsass Organization"
+      VALUE "CompanyName", "LibSass Organization"
```
nschonni: "Sass Open Source Foundation" from the license?

win/libsass.sln
```diff
@@ -1,10 +1,21 @@
...
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libsass", "libsass.vcxproj", "{E4030474-AFC9-4CC6-BEB6-D846F631502B}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".SolutionItems", ".SolutionItems", "{33318C77-2391-4399-8118-C109155A4A75}"
+	ProjectSection(SolutionItems) = preProject
+		..\.editorconfig = ..\.editorconfig
```
nschonni: Nice, does this actually pick up the setting when you include it here? I was having problems when editing in VS the other day because it wasn't respecting it

win/libsass.sln
```diff
@@ -1,10 +1,21 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+# Visual Studio 14
```
nschonni: Are we dropping 2013 as the base version?

res/resource.rc
```diff
@@ -18,14 +18,14 @@ BEGIN
...
       VALUE "FileDescription", "A C/C++ implementation of a Sass compiler"
-      VALUE "FileVersion", "0.9.0.0"
+      VALUE "FileVersion", "1.0.0.0"
       VALUE "InternalName", "libsass"
-      VALUE "LegalCopyright", "©2014 libsass.org"
+      VALUE "LegalCopyright", "\251 2017 libsass.org"
```
nschonni: libsass.org just redirects to sass-lang.com now

xzyfer: It redirects to the LibSass landing page. It's worth keeping for now.

res/resource.rc
am11: Will fix. Thanks.

win/libsass.sln
am11: I *think* 2015 sln file works with 2013 and 2012, but not the lower ones. Besides, from command line perspective, `msbuild` adapts to the installed toolchain.

Now that things are moving fast with the advent of VS Code for windOS and Unices, the previewed full VS for Mac and previewed VS2017, I honestly see no point in sticking with 2013 when we still can provide *better* support from XP-10 and all related server editions with latest CRT/CPP toolchains (brining dozens of bugfixes in STL and standard conformance). If there is a bug refraining us to moving forward, like any other piece of software, lets report the issue upstream, track it in our repo and keep the dice rolling forward when we are no backward compatibility trade-offs. Officially VC team hasn't even dropped XP support yet (released: 24.04.2001).

My €0.2 :)

win/libsass.sln
am11: It seems to be working. I installed [editorconfig for VS](https://github.com/editorconfig/editorconfig-visualstudio) extension, opened `libsass/win/libsass.sln`, set `indent_size` to 12, opened `libsass/src/file.cpp` and Ctrl+K,D to format the document,  the intended 12 space indentation applied.

Note starting with VS2017, editorconfig and cmake have built-in support [[ref](https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes)].

win/libsass.sln
nschonni: I'm pretty sure you're correct, since it's only the project toolsversion stuff that breaks backwards compat. Maybe AppVeyor should be updated too (separately) https://github.com/sass/libsass/blob/2847edc45d1f887277fba5cd6873435e50da6f9b/appveyor.yml#L1

xzyfer: Due to issues with Node Sass we need to keep CI at VS2013.

xzyfer:
Thanks for looking into this @am11. I don't think we need to complicate this any further. This is the first time we've bumped the C API ABI, so the process is new to me, and undocumented.

am11:
Rebased the branch against master.

xzyfer:
Thanks. I'm in bed now, will take care of this in the morning.

PR #2292, opened 2017-01-11
Fix issue with parent selector evaluation

mgreter:
This took me quite a few hours to figure out. In the end the fix is rather small, but it seems to be exactly the one needed. It looks like this was previousely "wrongly" fixed here: https://github.com/sass/libsass/pull/2004 (that probably introduced this regression).


xzyfer:
Back ported the bug fix to 3.4.4.

PR #2297, opened 2017-01-14
Remove `image-url` function signature

mgreter:
This was deprecated a long time ago!


nschonni:
Probably good to throw the "Release - Breaking Change" tag on this one so it gets called out in the release notes

xzyfer:
Done

PR #2305, opened 2017-01-21
Ci and clang fixes

mgreter:
- Use more explicit matrix
- Fix issue with clang 3.5+
- Add more compiler variations
- Speed up coverage builds a bit
- Fix coverage on mac (path issue)
- Disable coverage for clang on linux
- Error out correctly on plugin tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,6 @@
 language: cpp
 sudo: false
 
-os:
-  - linux
-  - osx
-
-compiler:
-  - gcc
-  - clang
 
 # don't create redundant code coverage reports
 # - AUTOTOOLS=yes COVERAGE=yes BUILD=static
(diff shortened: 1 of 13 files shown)


mgreter:
Took me nearly four days on and off to finally find the culprits. In the end I actually was able to reproduce some issues localy, but testing against travis CI is really no fun nowadays, specially when it involves the mac osx builds 👎 I've tried to speed them up a bit so they hopefully will no longer timeout, if they even start in a timely fashion.

This PR includes one fix that was needed to satisfy the clang-3.7 address-sanitizer (which indeed was a very minor memory leak). This PR also brings a bit more variety of compilers, since I've learned how to configure them properly with travis. Seems like a nice addition. We are also back to 3 ci runs that produce coverage reports. With the latest cleanups and additional spec-tests, plus the tests via plugins, the coverage went from 80% to 85%, which is quite a big step given our codebase!

mgreter:
Waiting for osx builds to start. Will 🚢 when 💚

mgreter:
Now appveyor seems stuck. @xzyfer I can no longer cancel appveyor jobs. Might be because my collaborator role is just user for the sass account on appveyor. Pretty sure I was able to cancel them before. Looks like I have to let them time out until 💚

nschonni:
@andrew is the other admin left (besides @xzyfer), I think I nuked myself when I handed over the AV org a long time ago. If you have a better GH team to sync as admins it would be good to get fixed going forward.

mgreter:
Thanks @nschonni. I still do have access to the "sass" account, but I'm only user. Not sure how the permissions work for AV, I'm just pretty sure I had the rights to cancel or restart jobs. Anyway, the one job that got stuck timed out and this PR should get green in a few minutes :)

xzyfer:
@mgreter you AppVeryor role should be fixed. I couldn't fix it until you had signed in.

mgreter:
@xzyfer thx, confirmed, I see the options now again!

PR #2310, opened 2017-01-31
Reformatted and improved Readme.md

mgreter:
Our main readme was getting pretty dusty and was missing a few new documentation links.

Decided to reformatted the whole thing, fixed quite a few typos and also added a few paragraphs here and there. Feel free to nitpick my english and to add more corrections. I also added two new badges from http://isitmaintained.com/. IMO it [looks and reads a lot better already][1].

[1]: https://github.com/mgreter/libsass/blob/readme/Readme.md

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,95 +1,98 @@
-LibSass
-=======
+LibSass - Sass compiler written in C++
+======================================
 
-by Aaron Leung ([@akhleung]), Hampton Catlin ([@hcatlin]), Marcel Greter ([@mgreter]) and Michael Mifsud ([@xzyfer])
+Currently maintained by Marcel Greter ([@mgreter]) and Michael Mifsud ([@xzyfer])  
+Originally created by Aaron Leung ([@akhleung]) and Hampton Catlin ([@hcatlin])
 
-[![Unix CI](https://travis-ci.org/sass/libsass.svg?branch=master)](https://travis-ci.org/sass/libsass)
-[![Windows CI](https://ci.appveyor.com/api/projects/status/github/sass/libsass?svg=true)](https://ci.appveyor.com/project/sass/libsass/branch/master)
-[![Bountysource](https://www.bountysource.com/badge/tracker?tracker_id=283068)](https://www.bountysource.com/trackers/283068-libsass?utm_source=283068&utm_medium=shield&utm_campaign=TRACKER_BADGE)
-[![Coverage Status](https://img.shields.io/coveralls/sass/libsass.svg)](https://coveralls.io/r/sass/libsass?branch=feature%2Ftest-travis-ci-3)
-[![Join us](https://libsass-slack.herokuapp.com/badge.svg)](https://libsass-slack.herokuapp.com/)
+[![Unix CI](https://travis-ci.org/sass/libsass.svg?branch=master)](https://travis-ci.org/sass/libsass "Travis CI")
+[![Windows CI](https://ci.appveyor.com/api/projects/status/github/sass/libsass?svg=true)](https://ci.appveyor.com/project/sass/libsass/branch/master "Appveyor CI")
+[![Coverage Status](https://img.shields.io/coveralls/sass/libsass.svg)](https://coveralls.io/r/sass/libsass?branch=feature%2Ftest-travis-ci-3 "Code coverage of spec tests")
+[![Percentage of issues still open](http://isitmaintained.com/badge/open/sass/libsass.svg)](http://isitmaintained.com/project/sass/libsass "Percentage of issues still open")
+[![Average time to resolve an issue](http://isitmaintained.com/badge/resolution/sass/libsass.svg)](http://isitmaintained.com/project/sass/libsass "Average time to resolve an issue")
+[![Bountysource](https://www.bountysource.com/badge/tracker?tracker_id=283068)](https://www.bountysource.com/trackers/283068-libsass?utm_source=283068&utm_medium=shield&utm_campaign=TRACKER_BADGE "Bountysource")
+[![Join us](https://libsass-slack.herokuapp.com/badge.svg)](https://libsass-slack.herokuapp.com/ "Slack communication channels")
 
-https://github.com/sass/libsass
 
-LibSass is just a library, but if you want to RUN LibSass,
-then go to https://github.com/sass/sassc or
-https://github.com/sass/sassc-ruby or
-[find your local implementer](docs/implementations.md).
+[LibSass](https://github.com/sass/libsass "LibSass GitHub Project") is just a library!
+If you want to use LibSass to compile Sass, you need an implementer. Some
+implementations are only bindings into other programming languages. But most also
+ship with a command line interface (CLI) you can use directly. There is also
+[SassC](https://github.com/sass/sassc), which is the official lightweight
+CLI tool built by the same people as LibSass.
 
-LibSass requires GCC 4.6+ or Clang/LLVM. If your OS is older, this version may not compile.
+### Excerpt of "sanctioned" implementations:
 
-On Windows, you need MinGW with GCC 4.6+ or VS 2013 Update 4+. It is also possible to build LibSass with Clang/LLVM on Windows.
+- https://github.com/sass/node-sass (Node.js)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -1,95 +1,97 @@
...
-While LibSass is a library primarily written in C++, it provides a simple
-C interface which should be used by most implementers. LibSass does not do
-much on its own without an implementer. This can be a command line tool, like
-[sassc](https://github.com/sass/sassc) or a [binding](docs/implementations.md)
-to your favorite programing language.
-
-If you want to build or interface with LibSass, we recommend to check out the
-documentation pages about [building LibSass](docs/build.md) and
-the [C-API documentation](docs/api-doc.md).
+Current LibSass 3.4 should be compatible with Sass 3.4. Please refer to the [sass compatibility
```
nschonni: "3.4 branch" with a link?

Readme.md
```diff
@@ -1,95 +1,97 @@
...
-and tests there.
-
-We even run Travis tests for SassC!
-
-Tests
+Testing
 -------
 
-Since LibSass is a pure library, tests are run through the [SassSpec](https://github.com/sass/sass-spec) project using the [SassC](http://github.com/sass/sassc) driver.
+Since LibSass is a pure library, tests are run through the [Sass-Spec](https://github.com/sass/sass-spec)
```
nschonni: Think "sass-spec" is always lowercase

Readme.md
```diff
@@ -1,95 +1,97 @@
...
+developing, you can run `./script/spec`. This will clone SassC and Sass-Spec under the project folder and
+then run the Sass-Spec test suite. You may want to update the clones to ensure you have the latest version.
+Note that the scripts in the `./script` folder are mainly intended for our CI needs.
 
-To run tests against LibSass while developing, you can run `./script/spec`. This will clone SassC and Sass-Spec under the project folder and then run the Sass-Spec test suite. You may want to update the clones to ensure you have the latest version.
+Building
+--------
 
-### DEBUG builds
+To build LibSass you need GCC 4.6+ or Clang/LLVM. If your OS is older, you may need to upgrade
```
nschonni: Is there a minimum Clang/LLVM version too to call out?

Readme.md
```diff
@@ -1,95 +1,97 @@
...
 About
 -----
 
-LibSass is a C/C++ port of the Sass CSS precompiler. The original version was written in Ruby, but this version is meant for efficiency and portability.
-
-This library strives to be light, simple, and easy to build and integrate with a variety of platforms and languages.
+LibSass is a C++ port of the original Ruby Sass CSS compiler with a [C API](docs/api-doc.md).
+We coded LibSass with portability and efficiency in mind. You can expect LibSass to be magnitudes
+faster than Ruby Sass. We would even go as far and claim that it is the fastest CSS compiler
+around, which should be the case given it's written in C++.
```
xzyfer: >You can expect LibSass to be magnitudes faster than Ruby Sass. We would even go as far and claim that it is the fastest CSS compiler around, which should be the case given it's written in C++.

This bragging is unnecessary.

Readme.md
```diff
@@ -1,95 +1,97 @@
...
 
 Developing
 ----------
 
-As you may have noticed, the LibSass repo itself has
-no executables and no tests. Oh noes! How can you develop???
+As noted above, the LibSass repository does not contain any binaries or other way to execute
```
xzyfer: I'm struggling to follow this paragraph. IMHO the original was good.

Readme.md
xzyfer: With clang it's hard to know. If you're on OSX (which is where clang is prevalent) it'll just work.

Readme.md
xzyfer: The 3.4 branch isn't an official thing. It was an implementation detail because I wasn't ready to drop some large changes into Node Sass. Node Sass now has the latest 3.5 betas on master.

xzyfer: Additionally I believe http://sass-compatibility.github.io/  is no longer actively maintained. I did the last round of updates a long time ago.

We should just say that we aim to be compatible the latest Ruby Sass stable release. All known incompatibilities are in our issue tracker.

Readme.md
```diff
@@ -1,95 +1,97 @@
-LibSass
-=======
+LibSass - Sass compiler written in C++
+======================================
 
 by Aaron Leung ([@akhleung]), Hampton Catlin ([@hcatlin]), Marcel Greter ([@mgreter]) and Michael Mifsud ([@xzyfer])
```
xzyfer: It's worth distinguishing current maintainers. People still contact Hampton directly who then proxies messages on to me.

Readme.md
xzyfer: Sass Spec is fine. The "Sass" capitalization is part of the Sass branding.

xzyfer:
Nice work on the update. I found the language a little hard to follow in some places. I've left comments.

Readme.md
```diff
@@ -1,95 +1,97 @@
...
+[SassC](https://github.com/sass/sassc), which is the official lightweight
+CLI tool built by the same people as LibSass.
+
+### Excerpt of "sanctioned" implementations:
+
+- https://github.com/sass/node-sass (Node.js)
+- https://github.com/sass/perl-libsass (Perl)
+- https://github.com/wellington/go-libsass (Go)
+- https://github.com/sass/sassc-ruby (Ruby)
+- https://github.com/medialize/sass.js (JS)
```
xzyfer: https://github.com/sass/libsass-net is now under the Sass org also.

Readme.md
mgreter: Can you provide an alternative paragraph?

Readme.md
mgreter: Can you provide an alternative paragraph?

Readme.md
mgreter: Not sure what you mean exactly? IMO it adds a bit more information than the old wording. And I don't think the old wording was any less confusing? IMO the whole paragraph does not make much sense, but I've left it in since it already was there.

Readme.md
mgreter: Can you provide an alternative paragraph?

Readme.md
mgreter: I included the url

Readme.md
mgreter: I separated it into two lines with "current mainters" and "original creators"

Readme.md
mgreter: Tried the rewrite it a little and be more "humble". I really would like to keep the emphasize on the speed, because that is IMO one of the core and key assets of libsass.

mgreter:
I've updated some parts a bit. Please provide alternative wording or paragraphs for the open things. Otherwise it's difficult to improve it without specific feedback. Thanks!

xzyfer:
I'll provide some copy this week.

mgreter:
@xzyfer still not heard back from you on this one!

xzyfer:
Sorry I've been preoccupied with tracking that Alpine issues since it's blocked 3.5.0.beta.3. Feel free to ship this if you want and I'll PR fix later. My priority atm is figuring out what's causing the memory issues with Alpine.

mgreter:
OK. Btw. I suspect a compiler (buggy toolchain) issue on alpine linux. I debugged it down to the heisenbug on the assembler level, but couldn't see why accessing a function jmp fails after the same jmp/function was called successfully before many times at the same memory location. I verified that the memory location is not overwritten. So for me this is a "wont fix" until we see similar issues on other platforms!

mgreter:
FWIW. The only difference between node-sass and sassc is that node-sass seems to flag some memory pages as non-executable (as seen when running with strace), which could be a reason why the jmp could fail. But I wasn't able to really make a connection (the raw memory addressess don't seem to overlap).

mgreter:
And in the end musl libc is a very young libc implementation (initial release is less than 5 years old, so chances are that it is simply a bug in said musl libc implementation) ...

xzyfer:
I suspect you're correct. However if there is something we can do to mitigate the issue I'd like to try. Alpine is becoming a very important tool within the containerisation realm.

PR #2315, opened 2017-02-04
Replace `to_string` compare with `==` object compare

mgreter:
This a follow up for https://github.com/sass/sass/issues/2229. This breaks a spec test which should hopefully be the same when https://github.com/sass/sass/pull/2232 is released. There might be more cases where we use `to_string` to compare objects. This could lead to some regressions, as I think the compare functions do lack some edge cases. But we want them to surface, as we already use them in our hash map implementation. It is also faster than converting to a string before.


mgreter:
This has some unexpected failures in https://ci.appveyor.com/project/sass/libsass/build/1.0.3068/job/ih3by0c12b0fc0jv! Needs further investigations ...

mgreter:
Seems this is a bit too early. The failing spec test is `is-superselector(':-moz-any(.foo, .bar)', ':-s-any(.foo, .bar)');` which crashes with a segfault, seems to indicate that we store some kind of selector inside those wrapped selectors. The compare operator doesn't seem to like it. Might be that we have some crude static cast lingering somewhere. The solution should be the same as in https://github.com/sass/libsass/pull/2300

mgreter:
It seems I was able to fix the issue. Not it is failing only the expected spec test.

mgreter:
Up on further review this needs some more changes as it is not equivalent to ruby sass now. There seems to be some path in `unify_with` that does not properly use the `==` operator.

PR #2322, opened 2017-02-07
Fix SIGSEGV from bad cast

### LINKED ISSUE #2321: SIGSEGV on invalid if function call
Originally reported by @mrtuxracer
```sass
a {
  b: if(a,b,c...);
}
```
Ruby Sass
```css
a {
  b: b; }
```
LibSass
```
[1]    33262 segmentation fault
```
Spec https://github.com/sass/sass-spec/pull/1080

xzyfer:
- [x] don't segfault
- [ ] don't error
- [ ] produce correct output

Fixes #2321
Spec https://github.com/sass/sass-spec/pull/1081

### DIFF
--- a/src/bind.cpp
+++ b/src/bind.cpp
@@ -168,7 +168,7 @@ namespace Sass {
         // normal param and rest arg
         List_Obj arglist = Cast<List>(a->value());
         // empty rest arg - treat all args as default values
-        if (!arglist->length()) {
+        if (!arglist || !arglist->length()) {
           break;
         } else {
           if (arglist->length() > LP - ip && !ps->has_rest_parameter()) {


mgreter:
I see you already started fixing this. I did the same and came up with this hotfix:
```diff
@@ -167,8 +167,15 @@ namespace Sass {
       else if (a->is_rest_argument()) {
         // normal param and rest arg
         List_Obj arglist = Cast<List>(a->value());
+        if (!arglist) {
+          if (Expression_Obj arg = Cast<Expression>(a->value())) {
+            arglist = SASS_MEMORY_NEW(List, a->pstate(), 1);
+            arglist->append(arg);
+          }
+        }
+
         // empty rest arg - treat all args as default values
-        if (!arglist->length()) {
+        if (!arglist || !arglist->length()) {
           break;
         } else {
           if (arglist->length() > LP - ip && !ps->has_rest_parameter()) {
```

Arglists is one of those classes that need an urgent refactoring.
But there are many like these on the todo list (ie. all vectorized classes)!

xzyfer:
@mgreter I'm about to go to bed. If want to take this over whilst I'm sleeping feel free.

I got distracted by making a reduced test case for https://github.com/sass/node-sass/issues/1879

mgreter:
OK, will make a PR. Note that I believe the other segfault in Alpine to be a heisenbug. I don't think you will be able to reduce the test case while still exposing the same bug. So far it looks like we are overwriting memory where we should not. Only explanation I could come up so far is that we rely on some undefined behavior somewhere, which only breaks in that certain case. It somewhat looks like stack memory is overwritten, which leads to invalid jmp address.

To be clear, the segfault in my case happens when the executable tries to jump into the "parse_expression" function. But interestingly it is called from "parse_arguments", so this might be related. I know that we do store strange variations of AST nodes in these containers. But I haven't finished my assessments, so this could also be the wrong assumption.

xzyfer:
I managed to come up with a relatively small reproducible test case see https://github.com/sass/node-sass/issues/1879#issuecomment-277993203.

mgreter:
Fixed in https://github.com/sass/libsass/pull/2323

PR #2323, opened 2017-02-07
Fix another null pointer access

### LINKED ISSUE #2321: SIGSEGV on invalid if function call

mgreter:
Hotfix 
Fixes #2321
Spec sass/sass-spec#1081

### DIFF
--- a/src/bind.cpp
+++ b/src/bind.cpp
@@ -167,8 +167,15 @@ namespace Sass {
       else if (a->is_rest_argument()) {
         // normal param and rest arg
         List_Obj arglist = Cast<List>(a->value());
+        if (!arglist) {
+          if (Expression_Obj arg = Cast<Expression>(a->value())) {
+            arglist = SASS_MEMORY_NEW(List, a->pstate(), 1);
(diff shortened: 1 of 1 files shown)


src/bind.cpp
```diff
@@ -167,8 +167,15 @@ namespace Sass {
...
+            arglist->append(arg);
+          }
+        }
+
         // empty rest arg - treat all args as default values
-        if (!arglist->length()) {
+        if (!arglist || !arglist->length()) {
           break;
         } else {
           if (arglist->length() > LP - ip && !ps->has_rest_parameter()) {
```
nschonni: This would have the same issue wouldn't it?

nschonni: Nevermind, read it wrong

PR #2329, opened 2017-02-15
Coverity scan

nschonni:
First test of the setup.
Currently failing on my Travis, but that might be because it's looking for the right repo https://travis-ci.org/nschonni/libsass/jobs/201763546#L387
Alternately the secure string is generated incorrectly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,12 @@ sudo: false
 # built with coverage, clang and autotools
 # - AUTOTOOLS=yes COVERAGE=yes BUILD=shared
 
+env:
+  global:
+   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
+   #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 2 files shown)


nschonni:
@xzyfer I think you'll need to merge this to see if it actually works since i'm not seeing the addon appearing in the PR config payload https://travis-ci.org/sass/libsass/jobs/201764378/config

xzyfer:
Travis secrets won't work for PRs from forked repos. It's a security feature.

nschonni:
True about the secrets, but i think the addon might also just be ignored for PRs since it doesn't show up in the config.

xzyfer:
I found a hacky way to make you a contributor. Use your powers for good.

nschonni:
Looks like it really wants an email https://travis-ci.org/sass/libsass/jobs/201768367#L390
I'll try mine for now, and just use a dummy in the end

mgreter:
Thanks @nschonni for this. It actually pointed out more than I expected! I looked through the list and it catched 2 memory leaks (due to the new memory handling) and one possible out of bound access. The rest were mostly undefined behaviors due to possibly unitialized values. Most of them were unproblematic with the given behavior of gcc/clang/msvc. None the less I fixed most of them with #2335. The rest seem to be false positives, but I have to see the updated report after said PR is merged (I pushed these changes also to the coverity branch). The rest was mostly cosmetic (i.e. "logical dead code")!

Issue #2330, opened 2017-02-15
index variable reference issues in for loops within functions


meodai:
### index variable reference issues in for loops within functions

Looks like index variables within loops, within functions are references. Since all the map entries contain the last $index.

http://www.sassmeister.com/gist/965a577f60fd8f8e45d21235edb7cc9a

input.scss
```scss
@function test () {
  $m: ();
  $abc: a b c d e f g h i j k;
  
  @for $index from 1 through length($abc) {;
    $m: map-merge($m, (nth($abc, $index):$index) );
  }
  
  @return $m;
}

test {
  content: #{test()};
}
```

libsass 3.5.0.beta.2
```css
(a: 11, b: 11, c: 11, d: 11, e: 11, f: 11, g: 11, h: 11, i: 11, j: 11, k: 11)
```

ruby sass 3.4.21
```css
(a: 1, b: 2, c: 3, d: 4, e: 5, f: 6, g: 7, h: 8, i: 9, j: 10, k: 11)
```

mgreter:
Thanks for your bug report! I can confirm this issue and your observation with "referenced iteration var" seems to be correct. But I'm not sure if the proposed fix is accurate. The open question is if `map-merge` does a deep clone or not. I guess it probably should. I believe we lack tests when merging two existing arrays and then change a sub-item. Should this change also be reflected on the other assignment, or are all deep structures independent once assigned?

The initial issue is fixed by #2332 but I know that this adds a pretty big overhead for tight loops (since we need to allocate a new "number" object for every iteration)! Maybe somebody could come up with more spec tests to cover all cases? That would greatly help to improve this area!

meodai:
wow that was quick. Thanks @mgreter

xzyfer:
This should be in node-sass in a day or so.

jacksleight:
According to the release notes this fix was included in 3.5.0.beta.3, however the latest node-sass is still bundling libsass 3.5.0.beta.2:

```
./node_modules/node-sass/bin/node-sass --version
node-sass	4.7.2	(Wrapper)	[JavaScript]
libsass  	3.5.0.beta.2	(Sass Compiler)	[C/C++]
```

Any idea when node-sass will be updated to 3.5.0.beta.3? This bug breaks our build currently so we're stuck on a very old version.

PR #2332, opened 2017-02-16
Fix for loop variable to be referenced (force a copy)

mgreter:
Fixes https://github.com/sass/libsass/issues/2330


xzyfer:
Activated the spec in https://github.com/sass/sass-spec/pull/1086

PR #2337, opened 2017-02-18
Speed up tight for loops

mgreter:
Avoids unnecessary copies of the iteration value
if it is only used in the loop control.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -178,6 +178,11 @@ namespace Sass {
       WRAPPED_SEL,
     };
   private:
+    // Force copy in eval if needed. This is only recognized by
+    // numbers at the moment. We want to avoid to make a separate
+    // ast node for every iteration. This effectively delayes that
+    // copy to the eval stage (where numbers are normaly static).
+    ADD_PROPERTY(bool, force_copy);
     // expressions in some contexts shouldn't be evaluated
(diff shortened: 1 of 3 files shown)


xzyfer:
IMO copying the iterator is the right thing to do here. Trying to get clever about the copy looks like may get messy, and potentially result in a long tail of bugs related to not checking the `force_copy` flag.

There's bigger perf fish to fry.

xzyfer:
Nevermind my previous comment. I think I was getting tripped up on the missing copy in the the number visitor.

mgreter:
IMO the unessecary AST Node copies are one of the main remeaining bottlenecks. But I'm curious to hear what you had in mind where the bigger perf fishes are?

I started to optimize these copies a long time ago. I.e. the number eval visitor was always returning copies in previous versions. This was adding up to 25% unenessary runtime overhead in certain cases. I also believe that the right place to make copies is on assignment and not when we "prepare" the ast variable. The flag would probably be better placed at the Number Class directly, but it is actually more efficient to add it to the existing booleans in the base class. Most dynamic languages would probably implement this with a copy on write (COW) mechanism, but we are not there yet.

mgreter:
Added some comments to make it clear that this is a very special flag for now!

PR #2340, opened 2017-02-20
Add libsass-python to Readme

asottile:
Perhaps a bit selfish but I definitely try and keep up to date with the project as much as possible with:
- an occasional fix here and there
- documentation help
- weird windows bugs
- release testing

libsass-python is an actively maintained and mature implementation which supports:
- custom functions
- custom imports

We're also very timely with releases -- usually releasing within a day of sass/libsass releases.


xzyfer:
Thanks @asottile

Issue #2277, opened 2016-12-31
New Feature in Sass 3.5: First class functions


xzyfer:
See sass/sass#2061

Specs added sass/sass-spec@506e335

PR #2373, opened 2017-05-09
Implement 3.5 first class functions

xzyfer:
Implements `get-function` and a `Function` to represent function references.

Fixes https://github.com/sass/libsass/issues/2277
Spec https://github.com/sass/sass-spec/pull/1108


xzyfer:
Down to one last failing test.

PR #2385, opened 2017-05-20
Improve C++ compiler compatibility

mgreter:
std::to_string is not available in GCC 4.6.3


xzyfer:
All the variations of the `rgb` function use `to_string`. Any reason only this call site was changed?

mgreter:
It was the only one failing on my compiler

mgreter:
And `std::to_string` is not the same as `AST::to_string`, or did I miss something?

xzyfer:
No, you're correct. My mistake

PR #2408, opened 2017-06-02
Docs: Add initial doxygen setup

nschonni:
Took the idea/template from the N-API project.
Currently the generated docs are excluded from the project, but I pushed up and example here http://nschonni.github.io/libsass/html/ afer turning on the `docs` folder in the Settings


xzyfer:
Whoa is that generating docs from parsing the C++ source?

My primary concern here is that having the internals documented implies a stability contract that we don't offer.

xzyfer:
This is hella cool though - https://nschonni.github.io/libsass/html/class_sass_1_1_a_s_t___node.html

nschonni:
> My primary concern here is that having the internals documented implies a stability contract that we don't offer.

Right now it doesn't even get pushed into the repo since it's gitignored. I just figured it was a good start to get it in place. Then the javadoc style blocks can be added to start indicating the stablity of parts before maybe adding it to a CI build publish

xzyfer:
Aside from the C API it's all unstable, maybe indefinitely. If we could
restrict the parsing to just those file this could be nifty though.

docs/Doxyfile
```diff
@@ -0,0 +1,2451 @@
...
+# need to set EXTENSION_MAPPING for the extension otherwise the files are not
+# read by doxygen.
+#
+# If left blank the following patterns are tested:*.c, *.cc, *.cxx, *.cpp,
+# *.c++, *.java, *.ii, *.ixx, *.ipp, *.i++, *.inl, *.idl, *.ddl, *.odl, *.h,
+# *.hh, *.hxx, *.hpp, *.h++, *.cs, *.d, *.php, *.php4, *.php5, *.phtml, *.inc,
+# *.m, *.markdown, *.md, *.mm, *.dox, *.py, *.pyw, *.f90, *.f95, *.f03, *.f08,
+# *.f, *.for, *.tcl, *.vhd, *.vhdl, *.ucf and *.qsf.
+
+FILE_PATTERNS          =
```
nschonni: @xzyfer I think by changing this to the `sass.h`, and mabye the docs, it would only expose those parts

PR #2428, opened 2017-06-27
expose list-delim option for simplifying huge css changes

jbremer:
Please read the commit message for more information.

I'll be upfront that, although it builds, I haven't actually tested the code, as there's more to it than just a change in the backend library. In our case I believe we're going to need changes to `gulp-sass` as well.

Is this a change that you'd consider finishing off, provided the use-case?
Some example commits where this would be useful: https://github.com/cuckoosandbox/cuckoo/commit/cb7fda22c38e244e1e6bccc1ee2f9d4e8008c21c https://github.com/cuckoosandbox/cuckoo/commit/cc89f90443a3c3e6337bf8d4043f88456a3b0886 https://github.com/cuckoosandbox/cuckoo/commit/02b95ca2a70ec59a8ecbc2a61123ab85f5aedc4e.

### DIFF
--- a/include/sass/context.h
+++ b/include/sass/context.h
@@ -79,6 +79,7 @@ ADDAPI bool ADDCALL sass_option_get_omit_source_map_url (struct Sass_Options* op
 ADDAPI bool ADDCALL sass_option_get_is_indented_syntax_src (struct Sass_Options* options);
 ADDAPI const char* ADDCALL sass_option_get_indent (struct Sass_Options* options);
 ADDAPI const char* ADDCALL sass_option_get_linefeed (struct Sass_Options* options);
+ADDAPI char ADDCALL sass_option_get_list_delim (struct Sass_Options* options);
 ADDAPI const char* ADDCALL sass_option_get_input_path (struct Sass_Options* options);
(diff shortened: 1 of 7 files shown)


xzyfer:
Thanks for taking to time to contribute. 

However I do not believe LibSass is the right place for this change. If anything we may be reducing output options. There are just way too many edge cases to handle when dealing with output generation. 

Additionally due to how sass-spec works this isn't a feature we can reasonably test, and such guarantee to maintain.

Formatting operations are much easier to do as a post-process on the generated CSS. Since you're already using gulp I recommended using a post-css backed formatter.

jbremer:
Fair enough. Would you by any chance know of a starting point for such CSS formatter in Gulp? I just ran into `postcss`, but didn't see anything in this direction in there. I'm personally not all that familiar with the tools out there in this area.

xzyfer:
PostCSS is a generic CSS parser and AST. There's an huge ecosystem of tools built upon it. [CSSComb][1]  or [stylelint][2] might be a good starting points.

[1]: http://csscomb.com/config
[2]: https://stylelint.io/

jbremer:
Oh, sweet, thanks! https://cuckoo.sh/ebf1b499512a098a/screenshot.png
Best of luck with the rest of the project.

PR #2438, opened 2017-07-11
Implement nesting guard to avoid "out of stack space"

mgreter:
Note that this limit is not an exact science
it depends on various factors, which some are
not under our control (compile time or even OS
dependent settings on the available stack size)
It should fix most common segfault cases though.

Fixing this correctly would need a major refactoring
of the parser to avoid any recursing function calls.
IMO the limit should be pretty safe and it at least
allows somebody to finetune libsass if they need.

Fixes 5+ issues as reported on hackerone

### DIFF
--- a/src/ast_def_macros.hpp
+++ b/src/ast_def_macros.hpp
@@ -29,6 +29,11 @@ class LocalOption {
 };
 
 #define LOCAL_FLAG(name,opt) LocalOption<bool> flag_##name(name, opt)
+#define LOCAL_COUNT(name,opt) LocalOption<size_t> cnt_##name(name, opt)
+
+#define NESTING_GUARD(name) \
+  LocalOption<size_t> cnt_##name(name, name + 1); \
+  if (nestings > MAX_NESTING) throw Exception::NestingLimitError(pstate); \
 
 #define ATTACH_OPERATIONS()\
 virtual void perform(Operation<void>* op) { (*op)(this); }\
--- a/src/error_handling.cpp
+++ b/src/error_handling.cpp
@@ -59,6 +59,10 @@ namespace Sass {
     : Base(pstate, msg, import_stack)
     { }
 
+    NestingLimitError::NestingLimitError(ParserState pstate, std::string msg, std::vector<Sass_Import_Entry>* import_stack)
+    : Base(pstate, msg, import_stack)
+    { }
+
     UndefinedOperation::UndefinedOperation(Expression_Ptr_Const lhs, Expression_Ptr_Const rhs, const std::string& op)
     : lhs(lhs), rhs(rhs), op(op)
     {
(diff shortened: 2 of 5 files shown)


src/ast_def_macros.hpp
```diff
@@ -29,6 +29,11 @@ class LocalOption {
 };
 
 #define LOCAL_FLAG(name,opt) LocalOption<bool> flag_##name(name, opt)
+#define LOCAL_COUNT(name,opt) LocalOption<size_t> cnt_##name(name, opt)
+
+#define NESTING_GUARD(name) \
+  LocalOption<size_t> cnt_##name(name, name + 1); \
+  if (nestings > MAX_NESTING) throw Exception::NestingLimitError(pstate); \
```
xzyfer: Should `nestings` be `name` here?

xzyfer:
This commit is failing on the 3.4-stable branch. Not exactly sure why but I've track it down to the default argument assignment of `msg` here

```
NestingLimitError(ParserState pstate, std::string msg = def_nesting_limit, std::vector<Sass_Import_Entry>* import_stack = 0);
```

mgreter:
Cannot reproduce locally against https://github.com/sass/sass-spec/tree/b744c1036a47681c45b3a2b0eee915ae24166e5e (3.4 seems to be removed in latest). Also pretty unlikely to be the cause for the failing at-root-alone-itpl/basic-alone-itpl tests. We also use the same default argument assignment in https://github.com/sass/libsass/pull/2438/files#diff-3f483a2c484893970bcb41dc1c2dac79L29 for quite some time ...

src/ast_def_macros.hpp
mgreter: No, the Macro is called as `NESTING_GUARD(nestings)`, therefore it will in turn insert a call to `LocalOption<size_t> cnt_nestings(nestings, nestings + 1)`. Meaning it will store the previous value and then increase it by one (reverting once local variable `cnt_nesting` goes out of scope).

mgreter:
Issue seems related to the length of the sassc invocation.

```
works: sassc\bin\sassc.exe sass-spec\spec\libsass\base-level-parent\root\basic-alone-itpl\input.scss
crashes: sassc\bin\sassc.exe sass-spec\..\sass-spec\spec\libsass\base-level-parent\root\basic-alone-itpl\input.scss
crashes: sassc\bin\sassc.exe  sass-spec\spec\..\spec\libsass\base-level-parent\root\basic-alone-itpl\input.scss
works: ..\sassc\bin\sassc.exe  spec\..\spec\libsass\base-level-parent\root\basic-alone-itpl\input.scss
```

mgreter:
Debugged it to the main memory corruption: https://github.com/sass/libsass/pull/2484

xzyfer:
Nice find

PR #2441, opened 2017-07-12
chore: add shebang for Shellcheck SC2148

nschonni:
https://github.com/koalaman/shellcheck/wiki/SC2148

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 if test "x$LIBSASS_VERSION" = "x"; then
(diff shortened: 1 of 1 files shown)


mgreter:
I hope this also works under MSYS/CYGWIN and windows cmd.com cmake/dmake/mingw32-make etc.

nschonni:
looks like it's only called by the Autotools make, but I'll see if I can test that on my machine

nschonni:
Actually, since this is Autotools only, I'm not going to bother

PR #2466, opened 2017-08-10
Update tap-driver

naroga:
Adds the recommended shebang syntax -- the previous way breaks on sample AWS Elastic Beanstalk PHP environments.


xzyfer:
Thanks @naroga.

I've fixed the CI issues. I have no real opinion on this change but I believe @saper might?

saper:
That's okay, thank you!

Issue #2477, opened 2017-10-01
Windows 10 MAX_PATH Problem in @import


Noxmox6:
Hi,
i have a Problem with libsass on Windows 10. I use node-sass and an Import with a long relative Path:

## input.scss
`@import "../../lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/lorem-ipsum-dolor-sit-ame/ExampleFile";`

For example in absolute Path:
`C:\MyProjects\parent\frontend\src\main\webapp\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\lorem-ipsum-dolor-sit-ame\_ExampleFile.scss`

The absolute path is greater than 256 Characters und i get the error: "File to import not found or unreadable". If I move the file to a shorter path, the build is working. But in our project we can’t use short paths and we need to run it on windows.

I already have set the new „greater than 256 Chars max_path setting“ in windows and in different other use cases this setting is working. But it is not working on sass compiling. I have test the project on Linux too and there are no problems.

Is there any workaround or fix to run @Imports with path length greater than 256 Chars on Windows 10?

mgreter:
This is a pretty well known limitation of the underlying OS functions we use:
https://msdn.microsoft.com/en-us/library/aa365247(v=vs.85).aspx#maxpath

None the less it is possible to use longer paths if one uses an alternative Windows API.
Unfortunately it is hard to predict if this change causes other regressions.
Could you check (build node-sass from source) if https://github.com/sass/libsass/pull/2478 would fix it for you?

Thanks

Noxmox6:
Hey,
thanks for your quick reply. :)

I have test it and now i can use a Path with 302 Characters. But if i increase the length to 369, it is also not working. The path length has been changed, but not to much

mgreter:
Hmmm, that doesnt really make sense ... can you try with sassc?

[...]

PR #2478, opened 2017-10-01
Implement long file path support for Windows

mgreter:
Fixes https://github.com/sass/libsass/issues/2477
May cause regressions in other cases. Hard to predict.
Lets see if it passes our CI tests ...

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -49,9 +49,10 @@ namespace Sass {
 
     // return the current directory
     // always with forward slashes
+    // always with trailing slash
     std::string get_cwd()
     {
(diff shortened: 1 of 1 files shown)


nschonni:
I wonder if sniffing the registry key from https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx with https://stackoverflow.com/questions/34065/how-to-read-a-value-from-the-windows-registry would make sense. Although is seems like the `\\` is pretty universal

mgreter:
Lets merge and see if we get any regression reported ...

Issue #2513, opened 2017-12-02
3.4.7: segmentation fault when cwd goes missing


asottile:
The backtrace points at `getcwd` returning `NULL` here:

https://github.com/sass/libsass/blob/5b9572e3fdfe2701352acf0621b888a15bf51280/src/file.cpp#L58

The trace I have is:

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
============================= test session starts ==============================
platform linux -- Python 3.6.3, pytest-3.3.0, py-1.5.2, pluggy-0.6.0
rootdir: /tmp/libsass-python, inifile:
collected 91 items                                                             

sasstests.py ........................................................... [ 64%]
...............................
Program received signal SIGABRT, Aborted.
0x00007ffff6cd2428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) where
#0  0x00007ffff6cd2428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff6cd402a in __GI_abort () at abort.c:89
#2  0x00007ffff2c7184d in __gnu_cxx::__verbose_terminate_handler() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff2c6f6b6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff2c6f701 in std::terminate() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff2c6f919 in __cxa_throw ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff2c9814f in std::__throw_logic_error(char const*) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff30d3044 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*> (
    this=this@entry=0xfd3a68, __beg=__beg@entry=0x0, __end=<optimized out>)
    at /usr/include/c++/5/bits/basic_string.tcc:216
[...]

PR #2529, opened 2018-01-11
Throw error if cwd goes missing

mgreter:
Addresses https://github.com/sass/libsass/issues/2513. Did not test windows implementation, but I trust our CI ;)


asottile:
sweet thanks!

mgreter:
You're welcome. Btw. thanks for your nice python integration and constant good bug reports!
LibSass could probably handle this error more gracefully, but I think a hard error is ok in this case 😃

Issue #2522, opened 2018-01-04
Emit a warning for &&


nex3:
Dart and Ruby Sass are adding a warning for `&&`, since it looks like a logical operator but doesn't behave like one. The text we're using is `In Sass, "&&" means two copies of the parent selector. You probably want to use "and" instead.`. LibSass should add the same warning.

See https://github.com/sass/sass/issues/2429 and https://github.com/sass/sass-spec/pull/1192.

mgreter:
FWIW. the following doesn't emit any warning at all in ruby sass 3.5.5
```
.and-and {
  value: true &&&& false;
}
```

mgreter:
Closed via @2530

PR #2530, opened 2018-01-11
Emit warning for double parent selectors

mgreter:
Addresses https://github.com/sass/libsass/issues/2522

Seems this needs to introduce yet another type of warning ...


nex3:
Don't forget to remove the TODOs from the corresponding specs!

xzyfer:
Removed todo in https://github.com/sass/sass-spec/pull/1200

Issue #2366, opened 2017-05-02
Segmentation fault with @extend and :global(> .sth)


tleunen:
### Title: Be as meaningful as possible in 60 chars if possible

input.scss
```scss
.item {
    display: inline-block;

    :global(> .ext-link) {
        background: #000;
    }
}

.link {
    color: red;
}

.textLink {
    @extend .link;
    padding: 0 10px;
}
```

[libsass 3.5.5] [1]
```css
Segmantation fault
```

ruby sass 3.4.21
```css
Segmantation fault
```

version info:
See https://github.com/sass/node-sass/issues/1961

mgreter:
Segfault has been fixed in libsass, but output does not match ruby sass.

mgreter:
This is basically also a duplicate of #2055 (missing specific pseudo selector behavior)

mgreter:
Closed via #2531

PR #2531, opened 2018-01-11
Fix whitespace issue for wrapped selectors

mgreter:
Finally addresses https://github.com/sass/libsass/issues/2366


xzyfer:
Nice one

PR #2534, opened 2018-01-13
Add support for custom property syntax

### LINKED ISSUE #2076: New Feature in Sass 3.5: Custom Properties
The [CSS custom properties spec](https://www.w3.org/TR/css-variables-1/) is starting to be supported by browsers. Like libsass, Sass 3.4 supported custom properties as basically the same as normal properties. Unfortunately, this is incompatible with plain CSS; custom property values can contain almost any text, including many Sass constructs that are currently parsed as Sass.

Sass 3.5 changes the behavior to hew much more closely to CSS. Custom property values are now parsed exactly according to the CSS spec, with the sole exception of interpolation, which is provides a way for users to inject Sass objects into their custom properties.

See sass/sass#1128 for a more thorough discussion. Related specs are in 
- [spec/css/custom_properties](https://github.com/sass/sass-spec/tree/master/spec/css/custom_properties)
- [spec/core_functions/feature_exists](https://github.com/sass/sass-spec/tree/master/spec/core_functions/feature_exists)

xzyfer:
I've been working on this on-and-off for the last couple months. 
It's mostly a 1-1 port of the Ruby Sass implementation.

The only thing of interest is that the whitespace after the `:`
need to be captured as written for custom properties.

This PR does not handle some less significant features
- [re-indentation][1]
- [sass syntax][2] (will open a bug downstream in sass2scss)

[1]: https://github.com/sass/sass-spec/tree/master/spec/css/custom_properties/indentation
[2]: https://github.com/sass/sass-spec/tree/master/spec/css/custom_properties/error/no_nesting

Spec sass/sass-spec#1199
Fixes #2076

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -2394,6 +2394,13 @@ namespace Sass {
     return quote(value_, '*');
   }
 
+  bool Declaration::is_invisible() const
+  {
+    if (is_custom_property()) return false;
+
+    return !(value_ && value_->concrete_type() != Expression::NULL_VAL);
+  }
+
   //////////////////////////////////////////////////////////////////////////////////////////
   // Additional method on Lists to retrieve values directly or from an encompassed Argument.
   //////////////////////////////////////////////////////////////////////////////////////////
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -660,19 +660,22 @@ namespace Sass {
     ADD_PROPERTY(String_Obj, property)
     ADD_PROPERTY(Expression_Obj, value)
     ADD_PROPERTY(bool, is_important)
+    ADD_PROPERTY(bool, is_custom_property)
     ADD_PROPERTY(bool, is_indented)
   public:
     Declaration(ParserState pstate,
-                String_Obj prop, Expression_Obj val, bool i = false, Block_Obj b = 0)
-    : Has_Block(pstate, b), property_(prop), value_(val), is_important_(i), is_indented_(false)
+                String_Obj prop, Expression_Obj val, bool i = false, bool c = false, Block_Obj b = 0)
+    : Has_Block(pstate, b), property_(prop), value_(val), is_important_(i), is_custom_property_(c), is_indented_(false)
     { statement_type(DECLARATION); }
     Declaration(const Declaration* ptr)
     : Has_Block(ptr),
       property_(ptr->property_),
       value_(ptr->value_),
       is_important_(ptr->is_important_),
+      is_custom_property_(ptr->is_custom_property_),
       is_indented_(ptr->is_indented_)
     { statement_type(DECLARATION); }
+    virtual bool is_invisible() const;
     ATTACH_AST_OPERATIONS(Declaration)
     ATTACH_OPERATIONS()
   };
--- a/src/inspect.cpp
+++ b/src/inspect.cpp
@@ -123,6 +123,8 @@ namespace Sass {
     if (dec->value()->concrete_type() == Expression::NULL_VAL) return;
     bool was_decl = in_declaration;
     in_declaration = true;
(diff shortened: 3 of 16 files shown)


xzyfer:
There are still a couple bugs to chase down, and some clean up but this is ready for some :eyes:

src/ast.hpp
```diff
@@ -673,6 +674,7 @@ namespace Sass {
       is_important_(ptr->is_important_),
```
mgreter: The new property is not copied when cloning here.

src/inspect.cpp
```diff
@@ -123,6 +123,8 @@ namespace Sass {
     if (dec->value()->concrete_type() == Expression::NULL_VAL) return;
     bool was_decl = in_declaration;
     in_declaration = true;
+    if (dec->is_custom_property()) in_custom_property = true;
```
mgreter: Maybe use LOCAL_FLAG macro to ensure it is reset automatically when scope is left (e.g. due to thrown error).

src/prelexer.cpp
```diff
@@ -1578,6 +1589,50 @@ namespace Sass {
...
 
+    extern const char css_variable_url_negates[] = "()[]{}\"'#/";
+    const char* css_variable_value(const char* src) {
+      return sequence<
+        alternatives<
+          sequence<
+            negate< sequence<
+              exactly<'u'>,
+              exactly<'r'>,
+              exactly<'l'>,
```
mgreter: Use `url_kwd` or maybe introduce a `urlfn_kwd` for reuse (seems we need both).

src/prelexer.cpp
```diff
@@ -1578,6 +1589,50 @@ namespace Sass {
...
+    extern const char css_variable_url_top_level_negates[] = "()[]{}\"'#/;!";
+    const char* css_variable_top_level_value(const char* src) {
+      return sequence<
+        alternatives<
+          sequence<
+            negate< sequence<
+              exactly<'u'>,
+              exactly<'r'>,
+              exactly<'l'>,
+              exactly<'('>
```
mgreter: Use `url_kwd` or maybe introduce a `urlfn_kwd` for reuse (seems we need both).

src/prelexer.cpp
```diff
@@ -1578,6 +1589,50 @@ namespace Sass {
...
+
+    extern const char css_variable_url_top_level_negates[] = "()[]{}\"'#/;!";
+    const char* css_variable_top_level_value(const char* src) {
+      return sequence<
+        alternatives<
+          sequence<
+            negate< sequence<
+              exactly<'u'>,
+              exactly<'r'>,
+              exactly<'l'>,
```
mgreter: Use `url_kwd` or maybe introduce a `urlfn_kwd` for reuse (seems we need both).

src/ast.cpp
```diff
@@ -2394,6 +2394,15 @@ namespace Sass {
     return quote(value_, '*');
   }
 
+  bool Declaration::is_invisible() const
+  {
+    return !(
+      is_custom_property() ||
+      value_ ||
```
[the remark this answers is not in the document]
xzyfer: Totally right, this was a last minute change.

src/ast.hpp
xzyfer: Nice catch

src/prelexer.cpp
xzyfer: Updated

src/inspect.cpp
xzyfer: Updated

src/prelexer.cpp
xzyfer: Updated

xzyfer:
CI is green

mgreter:
👍

Issue #2312, opened 2017-02-03
Mixins with compound "&" selectors make sourcemaps invalid


ceph3us:
Under a specific set of conditions, compound selectors using `&`, such as `&.mod-enabled` or `&:after`, cause the sourcemap generator to produce sourcemaps with negative column indices.

The conditions are:

- The rule must be a compound selector using the parent selector as above
- The rule must be in the root of a mixin
- The mixin must be included in the document at least 6 times 

This popped up for me when `gulp-sass` switched to version 4.0 of `node-sass`, which pushed LibSass up to 3.4. Smallest reproducible test case I could make is [available here](https://github.com/ceph3us/break-gulp-sass).

Sample error:
```
Error: Invalid mapping: {"generated":{"line":1,"column":6},"source":"mixins.scss","original":{"line":3,"column":-4},"name":null}
    at SourceMapGenerator_validateMapping [as _validateMapping] (/Users/cepheus/break-gulp-sass/node_modules/source-map/lib/source-map-generator.js:277:13)
    at SourceMapGenerator_addMapping [as addMapping] (/Users/cepheus/break-gulp-sass/node_modules/source-map/lib/source-map-generator.js:101:12)
    at /Users/cepheus/break-gulp-sass/node_modules/concat-with-sourcemaps/index.js:60:28
    at Array.forEach (native)
    at SourceMapConsumer_eachMapping [as eachMapping] (/Users/cepheus/break-gulp-sass/node_modules/source-map/lib/source-map-consumer.js:155:14)
    at Concat.add (/Users/cepheus/break-gulp-sass/node_modules/concat-with-sourcemaps/index.js:58:18)
    at DestroyableTransform.bufferContents [as _transform] (/Users/cepheus/break-gulp-sass/node_modules/gulp-concat/index.js:68:12)
    at DestroyableTransform.Transform._read (/Users/cepheus/break-gulp-sass/node_modules/readable-stream/lib/_stream_transform.js:159:10)
    at DestroyableTransform.Transform._write (/Users/cepheus/break-gulp-sass/node_modules/readable-stream/lib/_stream_transform.js:147:83)
    at doWrite (/Users/cepheus/break-gulp-sass/node_modules/readable-stream/lib/_stream_writable.js:347:64)
```

A bisection of the test case revealed this issue was introduced in commit de83d30.

ceph3us:
CCing in dlmanning/gulp-sass#577 so this issue is linked to that one.

mgreter:
[...]

PR #2540, opened 2018-01-18
Fix sourcemap crutch once again

mgreter:
Fixes https://github.com/sass/libsass/issues/2312


mgreter:
I'd like to keep it in for now in case we get any regressions due to this. I'm pretty sure I've added this for a reason, but due to lacking of any tests regarding source maps I can't really remember why.

entr:
Finally, thanks!

Issue #2404, opened 2017-05-30
Error indicator not unicode aware


mgreter:
With [latest perl-libsass we discovered two bugs][2] in how libsass creates the error indicator (the excerpt of error line with arrow indicator). First `e.pstate.src` does not seems to reference the right content when parsing interpolated selectors. Second the algorithm to create the excerpt is not unicode aware, as it only counts bytes, which can lead to unicode errors if implementors have strict rules for their strings (as perl does).

[I.e. the following line should use utf8::advance and similar helpers:][3]
```c
msg_stream << ">> " << std::string(line_beg + move_in, line_end - shorten) << "\n";
```

Will post a hotfix to solve the most urgent issue, since it is currently [failing most of perl-libsass][1] installs.

[1]:  http://matrix.cpantesters.org/?dist=CSS-Sass%203.4.6
[2]: https://github.com/sass/perl-libsass/issues/31
[3]: https://github.com/sass/libsass/blob/master/src/sass_context.cpp#L74-L93

PR #2543, opened 2018-01-18
Make error indicator Unicode aware

mgreter:
Should fix https://github.com/sass/libsass/issues/2404
Spec PR https://github.com/sass/sass-spec/pull/1202


mgreter:
Improved parser to check for unicode issues first to report a better and sanitized error message

PR #2546, opened 2018-01-19
Update libsass-python url

asottile:
Thanks for welcoming us to the org! 🎉

CC @dahlia


nschonni:
Great! Might want to take a look at https://github.com/sass/sass-site to see if there is anything to update in there as well

PR #2556, opened 2018-02-06
Fix preserving quoted strings on C-API

mgreter:
Related to https://github.com/sass/node-sass/issues/2228, since I did not understand how static parsing of strings, string constants and quoted strings do work before.


mgreter:
Obsolete with upcoming refactoring.

PR #2557, opened 2018-02-06
[WIP] Refactor node weave

mgreter:
This is a early WIP branch trying to fix the very old ast<->node implementation nobody understands. Subweave is already pretty much working ok, but still some more work to do. One can see that the initial architecture for complex_selectors is not very optimal, since we need to copy/clone the whole linked list chain (tails) once we want to change connections. For that I have introduced `Selector_Groups` for now. Putting this here since it was laying around for quite some time in my local repo.


glebm:
The `unify_with` implementation here should be updated to incorporate the changes from https://github.com/sass/libsass/pull/2737.

glebm:
@mgreter I have a branch where I've rewritten unification to match Dart Sass. This fixes a ton of bugs but unfortunately breaks `Complex_Selector` unification.

Perhaps you can make a PR that *just* implements subweave for AST nodes, then I can try updating `Complex_Selector` unification impl to also match Dart Sass.

glebm:
Update: My unification rewrite has been merged, all good.

src/ast_fwd_decl.hpp
```diff
@@ -385,6 +393,21 @@ namespace Sass {
...
+  template <class T>
+  bool SetsAreEqual(const T& lhs, const T& rhs) {
+    if (lhs.size() != rhs.size()) return false;
+    auto lit = lhs.begin(), lend = lhs.end();
+    auto rit = rhs.begin() /* , rend = rhs.end() */;
+    return std::equal(lit, lend, rit, CompareNodes());
+  }
+
+  template <class T>
+  bool SetContains(const T& lhs, const T& rhs) {
```
glebm: How about using hash maps/sets throughout (`std::unordered_map`)? They faster and easier to get right

PR #2564, opened 2018-02-20
Support var() in rbga short-hand syntax

### LINKED ISSUE #2561: CSS custom properties with rgba cause an error
Using CSS custom properties with rgba in scss files cause an error when trying to compile it.

Example code:

```css
.foo {
  color: rgba(var(--colorvar), 1)
}
```

> Error: argument `$color` of `rgba($color, $alpha)` must be a color

$ node-sass --version
node-sass	4.7.2	(Wrapper)	[JavaScript]
libsass 3.5.0.beta.2 (Sass Compiler) [C/C++]

xzyfer:
Fixes #2561
Spec https://github.com/sass/sass-spec/pull/1213

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -320,8 +320,8 @@ namespace Sass {
 
     inline bool special_number(String_Constant_Ptr s) {
       if (s) {
-        std::string calc("calc");
-        std::string var("var");
+        std::string calc("calc(");
+        std::string var("var(");
         std::string ss(s->value());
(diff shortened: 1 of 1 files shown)


mgreter:
Did you check that the same issue is not present in hsla and/or similar functions?

xzyfer:
Yes. This function was the only one missed the first time around according
to the reference implementations and sass-spec.

PR #2574, opened 2018-03-09
Add sass.cr to implementations list

straight-shoota:
This adds the [Crystal](https://crystal-lang.org) bindings `sass.cr` to the list of implementations.


xzyfer:
Thanks for supporting LibSass :)

Issue #2588, opened 2018-03-13
Regression in include paths


xzyfer:
There have been some reports of include paths behaviour changing in 3.5.0. I haven't had chance to investigate yet.

https://github.com/sass/node-sass/issues/2287
https://github.com/dlmanning/gulp-sass/issues/668

mgreter:
Works ok with `sassc.exe index.scss -I include`!??

mgreter:
`node_modules\.bin\node-sass index.scss --include-path include` works too??

xzyfer:
The gulp sass issue is interesting. Suggests an issue with the format of
the path maybe.

Still requires some digging on my end. Just wanted to flag it because there
have multiple reports for latest node sass beta.

mgreter:
OK, found the issue, seems windows UNC paths do not support `./` ...

xzyfer:
Not sure if this information makes any difference.
https://github.com/dlmanning/gulp-sass/issues/668#issuecomment-372963863

PR #2589, opened 2018-03-14
Fix win UNC path handling for dot and dotdot directories

mgreter:
This ~~implements~~ _uses_ a naive path cleanup for windows
to resolve dot and dotdot directories in otherwise already
absolute paths, since UNC file paths must be fully qualified
and do not support dot and dotdot references.

_Note that this does collapse x/../y sections into y.
This is by design and wrong, e.g. if /foo on your system is
a symlink to /bar/baz, then /foo/../cd is actually /bar/cd,
not /cd as a naive ../ removal will give you._

If someone knows of a MIT licensed function that does
UNC path resolving with proper symlink handling, please
speak up, I'm really getting tired of cumbersome win API.

Fixes https://github.com/sass/libsass/issues/2588

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -56,6 +56,8 @@ namespace Sass {
       #ifndef _WIN32
         char wd[wd_len];
         char* pwd = getcwd(wd, wd_len);
+        // we should check error for more detailed info (e.g. ENOENT)
+        // http://man7.org/linux/man-pages/man2/getcwd.2.html#ERRORS
         if (pwd == NULL) throw Exception::OperationError("cwd gone missing");
         std::string cwd = pwd;
       #else
@@ -74,11 +76,15 @@ namespace Sass {
     bool file_exists(const std::string& path)
     {
       #ifdef _WIN32
+        wchar_t resolved[32768];
         // windows unicode filepaths are encoded in utf16
         std::string abspath(join_paths(get_cwd(), path));
         std::wstring wpath(UTF_8::convert_to_utf16("\\\\?\\" + abspath));
         std::replace(wpath.begin(), wpath.end(), '/', '\\');
-        DWORD dwAttrib = GetFileAttributesW(wpath.c_str());
+        DWORD rv = GetFullPathNameW(wpath.c_str(), 32767, resolved, NULL);
+        if (rv > 32767) throw Exception::OperationError("Path is too long");
+        if (rv == 0) throw Exception::OperationError("Path could not be resolved");
+        DWORD dwAttrib = GetFileAttributesW(resolved);
         return (dwAttrib != INVALID_FILE_ATTRIBUTES &&
                (!(dwAttrib & FILE_ATTRIBUTE_DIRECTORY)));
       #else
@@ -200,6 +206,13 @@ namespace Sass {
       if (is_absolute_path(r)) return r;
       if (l[l.length()-1] != '/') l += '/';
 
+      // this does a logical cleanup of the right hand path
+      // Note that this does collapse x/../y sections into y.
+      // This is by design. If /foo on your system is a symlink
+      // to /bar/baz, then /foo/../cd is actually /bar/cd,
+      // not /cd as a naive ../ removal would give you.
+      // will only work on leading double dot dirs on rhs
+      // therefore it is safe if lhs is already resolved cwd
       while ((r.length() > 3) && ((r.substr(0, 3) == "../") || (r.substr(0, 3)) == "..\\")) {
         size_t L = l.length(), pos = find_last_folder_separator(l, L - 2);
         bool is_slash = pos + 2 == L && (l[pos+1] == '/' || l[pos+1] == '\\');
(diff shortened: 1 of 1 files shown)


xzyfer:
@mgreter if you could compose a tweet about what exactly you'd like help, or reference to, with the relevant information, then I can try to get some folks to boost the signal.

mgreter:
After some more API doc reading and try and error I figured out, that contrary to the MSDN docs some functions do support dot and dotdot references in UNC paths, namely GetFullPathNameW. There are many open questionmarks on my side, e.g. thread safety, why don't they just allow it with CreateFileW or GetFileAttributesW, how does it handle symbolic links in the path ...

xzyfer:
@am11 is my Goto windows expert. Maybe he can shed some light?

xzyfer:
The documentation suggest [GetFullPathName might not be thread-safe][1]

>Note  The GetFullPathName function is not recommended for multithreaded applications or shared library code. For more information, see the Remarks section.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa364963(v=vs.85).aspx

am11:
> No suprise why so few programs (even windows Explorer) fail to implement all this properly ...

I agree and it is indeed a complex problem to support all kinds of path formats in Windows. Not many applications, including Microsoft's own do not support them all.

I will quote https://github.com/dotnet/corefx/issues/16094#issuecomment-280105166:

> On Windows should it give back the "canonical" volume, which is Guid based `\\?\Volume{572FE591-B2EF-4617-AFEC-8DE83A32A7C5}`? <ins>The same volume can be given any number of aliases</ins>. While this seems niche at first I expect it to become more common. There are only 26 possible DOS path drive letters, I can see wanting to use volumes without "mounting" them or using something other than `A-Z:`.

Another strange-looking path can be of form `\??\dir1\dir2` (note how it differs from `\\?\...`), which is alias to the device aka NT Object name: https://en.wikipedia.org/wiki/Object_Manager_(Windows)#Object_structure. Some applications do support it when reading the registry values.

Then in PowerShell, ~/ resolves to `$HOME`/`$env:USERPROFILE` (note `~/` is legal, but `~\` is illegal). Also PowerShell can handle `\`, `/` and mixture of these, because at some point these were implemented to make PowerShell POSIX (1.0) complaint. I am not sure if NTFS/Win32 APIs can handle `/` the same way PowerShell does?

Thus, path normalization can also be a pain, which in our case is required for source-map generation in addition to reading the contents of file.

IMO, we should confine the scenarios we want to support, and call it out in docs explicitly what are those supported (advanced) UNC path formats on Windows. For example, if this PR is fixing certain scenarios reported by user, without breaking the existing ones, it's probably enough until the next issue is reported. :)

xzyfer:
I stumbled across this which looked promising.

http://pdh11.blogspot.com.au/2009/05/pathcanonicalize-versus-what-it-says-on.html

xzyfer:
Had a quick look at the boost `absolute` implementation and it appears to use `GetFullPathNameW` in some places.

xzyfer:
As @am11 says I'm happy to proceed with this approach and instruct folks to not use '.' or '..' if they run into issues.

mgreter:
I believe they mean it's unsafe to use if something is changing the current directory, anything else would surprise me, as the implementation would IMO pretty suck otherwise, although e.g. [realpath on Darwin][1] seems to do (has done?) exactly that. It should IMO anyway only be relevant if we feed it relative paths, which we dont. And BTW. since we already rely on CWD nothing really changes.

Additionally there are more confusing [remarks for _getdcwd and _wgetcwd][2]:

> This function is thread-safe even though it depends on GetFullPathName, which is itself not thread-safe. However, you can violate thread safety if your multithreaded application calls both this function and GetFullPathName. For more information, go to MSDN Library and then search for GetFullPathName.

The link provided by @xzyfer seems to support my assumption that GetFullPathName only does a naive logical path cleanup, therefore violating how symlinks and dotdot references (should) work, but since symlinks and windows are very seldomly used, I'm fine with that and it's propably no different to my initial approach. It's also unclear if win APIs even support this in the regular case.

Further Note: We do not support UNC paths, we only use them to get around the path length limitations and my main beef being that MS does not offer some simple API to do that.

All in all I agree with @am11: ... it's probably enough until the next issue is reported. We could opt to only use UNC paths if the given path exceeds a certain limit, but it would only shadow errors in certain use cases and make them hard to reproduce.

[1]: https://sourceforge.net/p/tcl/bugs/2274/
[2]: https://msdn.microsoft.com/en-us/library/7t2zk3s4.aspx#Remarks

mgreter:
On a side note, the link by @xzyfer makes one wrong assumption:
> on Unix, /home/peter/foo and /home/peter/src/../foo are the same file

This is not true when `/home/peter/src` is a symlink to e.g. `/opt/src`, then `/home/peter/src/../foo` would actually point to `/opt/foo` and not `/home/peter/foo`.

xzyfer:
I agree on all points :+1:

PR #2594, opened 2018-03-17
Roundhouse cleanup nr 1

mgreter:
Base refactoring #1 to bring function references to the C-API. First batch that AFAICT should not change any behavior, a lot more to come. Let's see if CI agrees (specially if I didn't mess up the windows build). Mainly moving stuff to separate compile units to better decouple implementations.

one adjustment needed in sass-spec: https://github.com/sass/sass-spec/pull/1231


mgreter:
MSCV and mac osx had some errors, therefore I included the fixes also here. I can split the PRs into smaller chunks if really needed, but it doesn't really change the commit history, so in order for me to safe some time I'll keep going here.

xzyfer:
No particular objects from me. I only ask we hold of merging significant changes like this until we have a stable 3.5.x for node-sass. We've only release 3.5.x to small fraction of the node-sass ecosystem as we work out any bugs/regressions.

mgreter:
Feel free to branch off a 3.5 stable branch. IMHO we can target this refactor for a libsass 3.6+ release, since sass has no version between 3.5 and 4.0, and the C-API will probably increment too. I expect that 3.5 is pretty stable already! Given that it took node-sass nearly a year to adopt 3.5 and LibSass waiting for that, I want to move this forward a bit faster this time. We need to clean out a lot of legacy stuff for that and naturally this means that we might break stuff that isn't covered with spec tests.

mgreter:
FWIW this PR should contain all improvements from my first WIP branch. I'll try to do the same with the next WIP branch tomorrow, as I have a few more ready on my local repo.

mgreter:
Yay, CI says 🚢 IT

mgreter:
FWIW I forked of https://github.com/sass/libsass/tree/3.5-stable from current master.

PR #2601, opened 2018-03-18
Fix automake build if sassc is missing

mgreter:
Improves and fixes `make check` target too


xzyfer:
Back ported to 3.5-stable

Issue #2360, opened 2017-04-20
CSS-Modules hexadecimal @value properties are undefined


monstersintokyo:
While migrating to webpack 2 I stumbled upon CSS-Modules `@value`s being undefined. Interestingly enough values like e.g. `1px` or rgb values are working fine. Hexadecimal definitions simply seem to vanish. Moreover I am not sure, why there is a `{}` after the second definition.

test.sass
```sass
@value test_1 #555
@value test_2 rgb(0,255,0)
```

`node node_modules\node-sass\bin\node-sass test.sass > test.css` produces:

test.css
```css
@value test_1;
@value test_2 rgb(0,255,0) {}
```

version info:
```cmd
$ node-sass --version
node-sass       4.5.2   (Wrapper)       [JavaScript]
libsass         3.5.0.beta.2    (Sass Compiler) [C/C++]

Windows 10
$ node -v
v4.8.2
```

Sassmeister output:
```css
@value test_1 #555
@value test_2 rgb(0,255,0);
```
Ruby output on win xp host:

```css
@value test_1 #555;
@value test_2 rgb(0,255,0);

/*# sourceMappingURL=test.css.map */
```

version info:
```cmd
$ ruby -v
ruby 2.0.0p648 (2015-12-16) [i386-mingw32]

$ sass -v
Sass 3.4.23 (Selective Steve)
```

I could reproduced the said behaviour since node-sass 3.6.0 with libsass 3.3.6.

Thanks!

xzyfer:
Thanks for the report. I am able to confirm this issue. We don't currently account for `@` rules with more than one argument.

monstersintokyo:
Any idea, if this will get fixed? Thanks!

mgreter:
Closed via https://github.com/sass/libsass/pull/2602

mgreter:
Seems the scss part is still missing, thought it passed locally ...

mgreter:
Fixed this specific edge case but there are probably more with `@at` directives.

chriseppstein:
@mgreter yes, I don't think sass should have any specific handling for `@value` like I see in https://github.com/sass/libsass/pull/2602/files#diff-950f6260135ab0fef6f27e24ebd6e863R605

[...]

PR #2603, opened 2018-03-19
Allow colors in `@at` directives

mgreter:
Fixes https://github.com/sass/libsass/issues/2360


xzyfer:
Back ported to 3.5-stable

Issue #2604, opened 2018-03-19
Applying darken function gives different result in libsass 3.5.0 and sass 3.4.21


40in:
[todo]: # (This is only a template!)
[todo]: # (remove unneeded bits)
[todo]: # (use github preview!)

Here is the full Gist of the range colors `#FFFFFF - #EEEEEE`: [libsas](https://gist.github.com/40in/6c58a95745b9e1ae5d3b12f7834a0742) and [sass](https://gist.github.com/40in/199cd182142ec969eecb7cf8905fa198) versions

The same example without `darken` function gives all white, thus expected result here is the same output color. In fact, we have 2 different values going back and forth.

## input.scss
```scss
test {
  border-top-color: darken(lighten(desaturate(#ffffff, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#fefefe, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#fdfdfd, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#fcfcfc, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#fbfbfb, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#fafafa, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f9f9f9, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f8f8f8, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f7f7f7, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f6f6f6, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f5f5f5, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f4f4f4, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f3f3f3, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f2f2f2, 15.7%), 33.5%), 14.625%);
  border-top-color: darken(lighten(desaturate(#f1f1f1, 15.7%), 33.5%), 14.625%);
[...]

PR #2606, opened 2018-03-19
Fix edge case converting achromatic colors to HSL

mgreter:
Fixes https://github.com/sass/libsass/issues/2604


xzyfer:
Back ported to 3.5-stable

Issue #2203, opened 2016-10-08
Incorrect evaluation of arithmetic inside interpolation


xzyfer:
The behaviour of this changed LibSass 3.4. Interestingly in this case Ruby Sass' output is also incorrect. A bug has been filed with Ruby Sass also https://github.com/sass/sass/issues/2160.

``` sass
foo {
  a: #{5% / 2};
  b: #{+5% / 2};
  c: #{-5% / 2};
}
```

Expected output

``` css
foo {
  a: 5%/2;
  b: 5%/2;
  c: -5%/2;
}
```

LibSass 3.4 RC

``` css
foo {
  a: 2.5%;
  b: 2.5%;
  c: -2.5%;
}
```

Ruby Sass 3.4.21

``` css
foo {
  a: 5%/2;
  b: 5%/2;
  c: -2.5%;
}
```

Spec https://github.com/sass/sass-spec/pull/940
##

xzyfer:
The behaviour exhibited by LibSass here is different from Ruby Sass, but it is no more or less inconsistent. I don't believe this should block 3.4 unless it arises as an issue during RC.

We'll address this for good when https://github.com/sass/sass/issues/2160 is resolved.

xzyfer:
Rescoping this to post 3.4.

mgreter:
@nex3 @chriseppstein is there a reason why ruby sass 3.5.5 still has this "unexpected" behavior? I believe this stems from static value parsing in ruby sass? Any plans if this will change in future versions?

nex3:
It just hasn't been a high priority to fix—as far as I know, it hasn't come up in any real code. `-5%/2` is the correct output, so that's what libsass should target.

PR #2607, opened 2018-03-19
Delay interpolated expression as in regular CSS

mgreter:
Fixes https://github.com/sass/libsass/issues/2203


xzyfer:
Back ported to 3.5-stable

Issue #2611, opened 2018-03-23
LibSass should not import CSS files, let alone as SCSS


nex3:
If I have two files:

```scss
// file1.scss
@import "file2";
```

```css
// file2.css
$var: 12;
a {b: $var}
```

...LibSass will import "file2" *and* interpret it as SCSS rather than plain CSS. Neither of these behaviors are correct. `.css` files cannot be imported by Sass files without an importer, and if they could they should be interpreted as plain CSS without allowing any Sass features.

I recommend adding a deprecation warning for this as soon as possible, possibly along with a link to an example importer that will import `.css` files for users for whom migrating will be difficult.

xzyfer:
This has been a long running issue, and one of the biggest node sass pain points. Thean was to removed this feature/bug with the module system landed to reduce disruption.

See #754
See #1656
See #1963

nex3:
I'd like to at least add a warning for this as soon as possible. Dart Sass will *not* be implementing this behavior, so I'd like to strongly warn people away from it.

xzyfer:
PR to add a deprecation warning #2613

tony:
> I recommend adding a deprecation warning for this as soon as possible, possibly along with a link to an example importer that will import .css files for users for whom migrating will be difficult.

Please provide some examples of how to handle importing raw CSS files from third party libraries. This would break a lot of behavior for me. It was an important feature.

xzyfer:
It was a bug not a feature. If you want to do this your code will not be
portable. If you use sassc you can use the new import extensions flag. That
feature will be supported in node-sass ever.. alternatively you can use a
custom importer.

liamkeily:
Like @tony I was using this to import css files from an npm package, whats the best way to go about this now?

xzyfer:
See my previous comment

qodesmith:
@bayucandra If you're using Webpack, you can import your css files at the top of your entry JavaScript file. That's where I was importing my top-level scss file anyway. And so long as you're using `css-loader` in you're chain of loaders, everything should be good to go.

tcchau:
[...]

PR #2613, opened 2018-03-24
Add a deprecation warning for @import's that resolved .css files

xzyfer:
See #2611

This is the first to removing native CSS imports.

Like Ruby Sass, Dart Sass will not implement this behaviour. This is
blocking compatibility between Node Sass and Dart Sass.

This is a non-standard behaviour which is constant source of confusion.

We had originally planned to removed this behaviour when the module
system was ready. However that has feature has been delayed and
people are continuing to be tripped up by this behaviour.

```
DEPRECATION WARNING on line 1, column 8 of /libsass/test.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.
```

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -373,6 +373,14 @@ namespace Sass {
     // process the resolved entry
     else if (resolved.size() == 1) {
       bool use_cache = c_importers.size() == 0;
+      if (resolved[0].deprecated) {
+        // emit deprecation warning when import resolves to a .css file
+        deprecated(
+          "Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.",
+          "Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.",
+          true, pstate
+        );
+      }
       // use cache for the resource loading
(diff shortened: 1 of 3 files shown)


xzyfer:
@nex3 feel free to word smith the deprecation warning. 

Linking to a specific custom importer implementation/documentation is tricky because LibSass is mostly used embedded into other projects.

xzyfer:
Back ported to 3.5-stable

xzyfer:
Rebased this off of master since it only belongs in 3.5.
The deprecated raw css imports are being removed in 3.6. See #2615.

xzyfer:
>There needs to be something somewhere that tells you what to do about this warning

The resolution is to stop importing .css files. This is a non-standard feature that goes against the Sass language spec.

>There appears to be zero documentation about sass_option_push_import_extension anywhere

It's a temporary work around in the C API for integrators. It's use is highly discouraged! It is only offered a temporary polyfill to enable  each implementation to migrate their users off of the invalid behaviour

PR #2619, opened 2018-03-27
Remove duplicate definition of out_of_memory macro

thatguystone:
When combining the project into a single file (as go-libsass does), gcc complains that `out_of_memory` is redefined.

Ref: https://github.com/wellington/go-libsass/issues/57

### DIFF
--- a/src/sass.cpp
+++ b/src/sass.cpp
@@ -34,8 +34,10 @@ extern "C" {
   void* ADDCALL sass_alloc_memory(size_t size)
   {
     void* ptr = malloc(size);
-    if (ptr == NULL)
-      out_of_memory();
+    if (ptr == NULL) {
+      std::cerr << "Out of memory.\n";
+      exit(EXIT_FAILURE);
+    }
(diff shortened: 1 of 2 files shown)


xzyfer:
Is this enough? The out_of_memory macro defined in util.cpp is under the Sass namespace

nschonni:
This file is a copy/paste from another library, so editiing might not be the best solution. Just thinking of if we ever try to sync it again (although I think it's been static since the 2011 copyright header) ❤️

thatguystone:
It looks like the file came from [json.c](https://github.com/rustyrussell/ccan/blob/master/ccan/json/json.c), which, as @nschonni noted, hasn't been updated since 2011.

Apart from the usage in json.cpp, `out_of_memory` is only used once, so it might be better just inline it. Thoughts?

xzyfer:
I'm ok with inlininh, or even just renaming the Sass namespaced version.

xzyfer:
Thanks

PR #2622, opened 2018-03-28
Emit transparent colors as rgba(0, 0, 0, 0)

### LINKED ISSUE #2298: Emit transparent colors as `rgba(0, 0, 0, 0)`
>[..] we've decided to officially drop support for versions of IE prior to IE9, which was the first to support `rgba()`. As such, we'll start emitting all colors with no alpha component as `rgba(0, 0, 0, 0)` across the board.

Sass issue: https://github.com/sass/sass/issues/1782

This is a significant BC break so this should be the last thing to land before tagging 3.5 stable. This way implementors can make releases with the CSS compat stuff in 3.5 without breaking their BC.

xzyfer:
This change breaks browser compat for IE < 10 so it is scheduled
for 3.6. It will not be back ported to the 3.5 line.

See sass/sass#1782
Fixes #2298
Spec https://github.com/sass/sass-spec/pull/1242

### DIFF
--- a/src/inspect.cpp
+++ b/src/inspect.cpp
@@ -671,9 +671,6 @@ namespace Sass {
     if (name != "") {
       ss << name;
     }
-    else if (r == 0 && g == 0 && b == 0 && a == 0) {
-      ss << "transparent";
-    }
     else if (a >= 1) {
       if (res_name != "") {
         if (compressed && hexlet.str().size() < res_name.size()) {


xzyfer:
Travis OSX boxes are backed up. Cancelled the build once it passed on Travis linux.

xzyfer:
After further clarifying the scope of the change, and the associated browser compat breakage we've decided to back port this to 3.5.

PR #2623, opened 2018-03-28
Mentioned SharpScss and LibSassHost in implementations.md

tompazourek:
Just added these two links to implementations in documentation:

* [SharpScss](https://github.com/xoofx/SharpScss)
* [LibSassHost](https://github.com/Taritsyn/LibSassHost)

### DIFF
--- a/docs/implementations.md
+++ b/docs/implementations.md
@@ -32,6 +32,8 @@ There are several implementations of `libsass` for a variety of languages. Here 
 * [libsass-net](https://github.com/darrenkopp/libsass-net)
 * [NSass](https://github.com/TBAPI-0KA/NSass)
 * [Sass.Net](https://github.com/andyalm/Sass.Net)
+* [SharpScss](https://github.com/xoofx/SharpScss)
+* [LibSassHost](https://github.com/Taritsyn/LibSassHost)
 
 ### Nim
 * [nim-sass](https://github.com/zacharycarter/nim-sass)


tompazourek:
I thought it might be useful to keep them all listed them even when they haven't been updated recently. They might still work just fine. I haven't tested SharpScss or LibSassHost myself, really, so I cannot tell if they are better than the other options on the list. I just noticed them as alternatives for somebody researching how to use libsass on the .NET Platform.

I think those links are still useful information, and I can imagine it can help someone. Many other links on the page are dead projects, I think it's alright.

**SharpScss** has been updated 7 months ago, which isn't that bad, really. I wouldn't think it's a dead project. A difference is that it runs on .NET Core, comparing to LibSassHost only running on full .NET Framework. That might be an important factor to some developers.

And as for **libsass-net**, it's one of the "sanctioned" implementations mentioned in README.md as well, and it falls under https://github.com/sass. I'm not sure if it makes sense to remove that one...

Taritsyn:
> …, comparing to LibSassHost only running on full .NET Framework. That might be an important factor to some developers.

@tompazourek LibSassHost supports a .NET Core (.NET Standard) since [version 1.0.0](https://github.com/Taritsyn/LibSassHost/releases/tag/v1.0.0).

xzyfer:
Thanks for this. I'll prune the abandoned projects separately.

PR #2646, opened 2018-05-08
Include <stdexcept> in operation.hpp

yboulkaid:
Because we are throwing `std::runtime_error` on line 192 (since aa7386f2fc5f9c3681fbe861220c40b5a7a910b5) , we need to include `<stdexcept>`.
This was causing building to fail with `runtime_error is not a member of std` (gcc version 4.6.3).


xzyfer:
Thanks!

PR #2648, opened 2018-05-10
Update vcxproj for VS2017

Geobert:
Hi, 

I'm trying to build a Rust project that uses `sass-rs` which uses `libsass` on appveyor, using VS2017 image and it fails: https://ci.appveyor.com/project/tak1n/cobalt-rs/build/1.0.781#L272
VS2010 is indeed not installed on VS2017 image: https://www.appveyor.com/docs/build-environment/#visual-studio-2010

Is it possible to upgrade vcxproj to use VS2015 build tools?

### DIFF
--- a/win/libsass.vcxproj
+++ b/win/libsass.vcxproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="GitVersion;Main" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <LIBSASS_VERSION>[NA]</LIBSASS_VERSION>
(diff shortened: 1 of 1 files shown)


Geobert:
Locally, to make it works, I had to open the sln with VS2017 and let it upgrade. Then I've compare what VS2017 added to the project and adding these platforms did the trick on my computer.

saper:
We are not requiring Visual Studio 2010 anywhere, in fact that one will it work because we depend on C++11 features.    Since sometimes it is needed to set the toolkit version externally I would prefer not to hardcore it in the son file.
Where is th is VS2010 default is coming from? Maybe it is set somewhere as a part of Rust scaffolding?

Geobert:
> Where is th is VS2010 default is coming from? Maybe it is set somewhere as a part of Rust scaffolding?

I've no clue :-/ If you think it's not a good idea discard this PR, I've manage a workaround in the build script from Rust.

PR #2684, opened 2018-07-02
Undeprecate css imports

xzyfer:
This reverts #2615 as it landed in 3.5. It will also be removed in master.

See sass/node-sass#2362 for additional context on this back flip.

****

This was added as a first step to removing support for raw CSS
`@imports`. However it appears that CSS @imports are a highly valued
feature among the current Sass community. Without a viable way
to resolve the deprecation warning AND keep the desired behaviour
we've decided to remove the blanket deprecation warning and restore
the old behaviour.

In the near future we will be introducing a new deprecation warning
when an `@import`'ed CSS file uses Sass language features.

Additionally this removes the C-API `sass_option_push_import_extension`
which was added a work around to ease the transition for implementors.

cc @asottile as an early heads up (very sorry for the churn!).


asottile:
Thanks for the heads up :)

PR #2708, opened 2018-08-06
Update make CC and CXX defaults for wider compat

### LINKED ISSUE #2707: Building with make does not work if g++ is not installed
I only have clang, no g++ on my system. However, when building this with `make`, it tries to invoke `g++` and fails (sassc-ruby builds with make during install).

This is because of:
https://github.com/sass/libsass/blob/edd29a2d9cb25a69dd1da84f485a8debf542ec80/Makefile#L2-L3

Perhaps these should instead default to:

```
CC       ?= cc
CXX      ?= c++
```

These are aliases on Debian and Ubuntu for the default c/cpp compilers, but I'm not sure about other systems. What do you think?

Worked around by setting these environment variables before installing.

glebm:
Fixes #2707

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 OS       ?= $(shell uname -s)
-CC       ?= gcc
-CXX      ?= g++
+CC       ?= cc
+CXX      ?= c++
 RM       ?= rm -f
 CP       ?= cp -a
 MKDIR    ?= mkdir


xzyfer:
Rebuilding CI

deivid-rodriguez:
Just run into this too. Is the build failure a real failure due to this PR?

glebm:
The build failure seems unrelated, because the failing pipeline doesn't even use this makefile.

deivid-rodriguez:
That's good! :)

xzyfer:
Kicked off a rebuild

PR #2728, opened 2018-11-12
Makefile: Use order-only deps for directories

### LINKED ISSUE #1433: Makefile `mkdir` needs `-p` flag

### LINKED ISSUE #1992: Don't mkdir existing directories when installing
Currently the install-* commands run mkdir directly when install-* commands are run, to ensure that the directory exists. mkdir raises an error if the directory already exists and this causes an abort of the build as more often than not, `/usr/local/lib" exists.

Some sort "ensure_exists" routine should be run instead.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32896125-don-t-mkdir-existing-directories-when-installing?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283068&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2727: facepalm: install fails because $(PREFIX)/lib already exists
# install fails because $(PREFIX)/lib already exists

```
[stephan@host:~/cvs/libsass]$ m install PREFIX=$HOME
mkdir /home/stephan/lib
mkdir: cannot create directory ‘/home/stephan/lib’: File exists
Makefile:263: recipe for target '/home/stephan/lib' failed
make: *** [/home/stephan/lib] Error 1
```

The fix is trivial: use `mkdir -p` instead of `mkdir`.

This is the same as #1992, but that ticket is inexplicably closed.

PS: when trying to use the autotools-generated build, by running `autoconf` to convert `configure.ac` to `configure`, the resulting `./configure` fails because:

```
[stephan@host:~/cvs/libsass]$ ./configure --prefix=$HOME
configure: error: cannot find install-sh, install.sh, or shtool in script "."/script
```

Thus i'm using the makefile which is checked in to the tree, rather than the one generated by autotools (a.k.a. GNU "Auto, my ass!" Tools).

glebm:
See https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

With this PR the targets only depend on directories' existence, but not on their timestamps.

Fixes #2727, fixes #1992, fixes #1433.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,8 +269,7 @@ $(DESTDIR)$(PREFIX)/include/sass: | $(DESTDIR)$(PREFIX)/include
 	$(MKDIR) $(DESTDIR)$(PREFIX)/include/sass
 
 $(DESTDIR)$(PREFIX)/include/%.h: include/%.h \
-                                 | $(DESTDIR)$(PREFIX)/include \
-                                 $(DESTDIR)$(PREFIX)/include/sass
+                                 | $(DESTDIR)$(PREFIX)/include/sass
 	$(INSTALL) -v -m0644 "$<" "$@"
 
 install-headers: $(DESTDIR)$(PREFIX)/include/sass.h \
--- a/Makefile
+++ b/Makefile
@@ -230,13 +230,13 @@ debug-shared: shared
 lib:
 	$(MKDIR) lib
 
-lib/libsass.a: lib $(COBJECTS) $(OBJECTS)
+lib/libsass.a: $(COBJECTS) $(OBJECTS) | lib
 	$(AR) rcvs $@ $(COBJECTS) $(OBJECTS)
 
-lib/libsass.so: lib $(COBJECTS) $(OBJECTS)
+lib/libsass.so: $(COBJECTS) $(OBJECTS) | lib
 	$(CXX) -shared $(LDFLAGS) -o $@ $(COBJECTS) $(OBJECTS) $(LDLIBS)
 
-lib/libsass.dll: lib $(COBJECTS) $(OBJECTS) $(RCOBJECTS)
+lib/libsass.dll: $(COBJECTS) $(OBJECTS) $(RCOBJECTS) | lib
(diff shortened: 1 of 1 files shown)


glebm:
@sgbeal I think this may also solve the problem described in #2727. Does it?

sgbeal:
@glebm if i knew how to use git/github properly, and could figure out how to pull this for a local test, i could probably tell you that :/. (i'm a Fossil SCM guy, not a git guy :/) i have no personal experience with order-only prerequisites, so i can't say with certainty whether it resolves the problem.

sgbeal:
@glebm 

     [stephan@host:~/cvs/libsass]$ git checkout patch-2
     error: pathspec 'patch-2' did not match any file(s) known to git.

That's after doing a fresh clone.

mgreter:
P.s. `git clone https://github.com/glebm/libsass.git --branch patch-2`

sgbeal:
@glebm Unfortunately:

```
$ git checkout origin/patch-2
error: pathspec 'origin/patch-2' did not match any file(s) known to git.
$ git checkout origin patch-2
error: pathspec 'patch-2' did not match any file(s) known to git.
```

@mgreter that does the trick.

This is exactly why i'm not a git guy - even the most banal operations are cryptic :/.

So, after building...

```
[stephan@host:~/cvs/libsass]$ m install PREFIX=$HOME
'lib/libsass.a' -> '/home/stephan/lib/libsass.a'
```

(noting that `~/lib` existed beforehand)

Success!

xzyfer:
Thank you all for seeing this through :heart:

PR #2729, opened 2018-11-12
Makefile: Compiler-specific LDFLAGS

glebm:
Detects compiler family by compiling and running a tiny C file on every invocation of make (30ms on my machine).

Alternative to #2720

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ GNUmakefile
 /script/install-sh
 /script/ltmain.sh
 /script/missing
+/script/print-compiler-family
 /script/test-driver
 /src/stamp-h1
 /src/Makefile.in
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,14 @@ else
   CXXFLAGS += -O1 -fno-omit-frame-pointer
   LDFLAGS  += -O1 -fno-omit-frame-pointer
 endif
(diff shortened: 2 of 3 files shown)


mgreter:
Seems coverage reporting by the target `(AUTOTOOLS=no COVERAGE=yes BUILD=static)` has broken. Maybe we should set coverage for this target to no (we still have to valid results). @glebm maybe you could adjust `.travis.yml`?

glebm:
> Seems coverage reporting by the target (AUTOTOOLS=no COVERAGE=yes BUILD=static) has broken

It's been broken for a while, not by this PR

mgreter:
Yep, I know, sorry, didn't want to blame your commit :) But maybe you'd have time to clean it up :)

mgreter:
Ok, we just need to change COVERAGE=YES to NO in travis.yml. I'll see if I have time to make a PR.

alyssais:
On second thoughts, I’m not actually sure this is going to solve my problem. Solaris generally uses the GCC _compiler_, but not its _linker_. And this is checking the compiler, right?

glebm:
@alyssais This is checking the compiler but these flags are passed to `$(CC)`. I'm not sure how the linker gets invoked. Can you try this on Solaris?

glebm:
@alyssais Oh, you're absolutely right. :man_facepalming:

PR #2732, opened 2018-11-13
Makefile cleanup

glebm:
1. Flattens nested if-else chains.
2. Replaces undefined-or-empty variable checks (`ifeq (,$(SOMEVAR))`) with `ifdef SOMEVAR` (equivalent, see https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html).
3. Merges two mostly duplicate sections of Windows-specific overrides.
4. Removes is-empty checks before `+= $(VAR)` (no-op if `VAR` is empty).


xzyfer:
Thanks!

PR #2735, opened 2018-11-14
SharedPtr: More cleanup

glebm:
1. Use `size_t` instead of `long` for `refcounter`.
2. Use parent impls for `SharedImpl` where possible.
3. Remove some unused methods.
4. Set `node` to `NULL` after calling `delete`.


xzyfer:
Thanks!

PR #2736, opened 2018-11-14
Fix ignored :not(multiple, selectors)

### LINKED ISSUE #2697: `:not()` clause ignored completely
This has started failing once we updated `gulp-sass` which indirectly updated `libsass` to 3.5.4.

## input.scss
```scss
.Card {
  &:not(.is-open, .is-static) {
    .CardContents {
      display: none;
    }
  }
}
```

## Actual results

[libsass 3.5.3][1]
```css
.Card .CardContents {
    display: none;
}
```

## Expected result

E.g. [DartSass 1.6.2, Sass 3.5.6](https://www.sassmeister.com/):
```css
.Card:not(.is-open, .is-static) .CardContents {
  display: none;
}
```

version info: I don't have the CLI versions installed to output the version, as we're using gulp-sass. But the problem is easily reproducible with libsass.ocbnet.ch or sassmeister.com.

[1]: http://libsass.ocbnet.ch/srcmap/#LkNhcmQgewogICY6bm90KC5pcy1vcGVuLCAuaXMtc3RhdGljKSB7CiAgICAuQ2FyZENvbnRlbnRzIHsKICAgICAgZGlzcGxheTogbm9uZTsKICAgIH0KICB9Cn0=

glebm:
I have no idea what this code does or what my change does but it does seem to fix #2697.

I do not propose submitting this PR, but only that people who understand what's going on here better than me please have a look.


xzyfer:
I've confirmed this addresses the reported issue and passes all other tests. I've attached the spec metadata to the PR and kicked off a CI build to confirm.

xzyfer:
This was originally reported as a regression however I couldn't find a point in the last 2 years where this behaviour worked correctly.

The changes seem reasonable enough to progress. Thanks again!

src/eval.cpp
```diff
@@ -1622,10 +1622,8 @@ namespace Sass {
   // be fixed by implement superselector correctly for `:not`
   // first use of "find" (ATM only implemented for selectors)
   bool hasNotSelector(AST_Node_Obj obj) {
-    if (Wrapped_Selector_Ptr w = Cast<Wrapped_Selector>(obj)) {
-      return w->name() == ":not";
-    }
-    return false;
+    Wrapped_Selector_Ptr w = Cast<Wrapped_Selector>(obj);
+    return w && w->name() == ":not";
```
xzyfer: This appears to be a style only change. If so can you please restore the previous style.

glebm: Yes, reverted.

xzyfer: Thanks

PR #2737, opened 2018-11-18
Fix selector unification ordering

### LINKED ISSUE #2681: Type selectors should always print before psuedo-classes
When extending a placeholder from a `[type]` selector, if the placeholder uses pseudo-classes, the selectors are printed in the wrong order.

## input.scss

```scss
%button-styles {
  color: red;
  
  &:focus {
    color: blue;
  }
}

[type="button"] {
  @extend %button-styles;
}
```

## Actual results

[libsass 3.5.4][1]
```css
[type="button"] {
  color: red;
}

:focus[type="button"] {
  color: blue;
}
```

## Expected result

[ruby sass 3.5.6](https://www.sassmeister.com/gist/ddecf6a0f29bca917d05c8ee46fd3c16)
```css
[type="button"] {
  color: red;
}

[type="button"]:focus {
  color: blue;
}
```

version info:
```cmd
$ node-sass --version
node-sass	4.9.0	(Wrapper)	[JavaScript]
libsass  	3.5.4	(Sass Compiler)	[C/C++]
```

[1]: http://libsass.ocbnet.ch/srcmap/#JWJ1dHRvbi1zdHlsZXMgewogIGNvbG9yOiByZWQ7CiAgCiAgJjpmb2N1cyB7CiAgICBjb2xvcjogYmx1ZTsKICB9Cn0KClt0eXBlPSJidXR0b24iXSB7CiAgQGV4dGVuZCAlYnV0dG9uLXN0eWxlczsKfQ==

I also ran a couple of debug statements with `selector-unify` and noticed if I used an element selector it unified as expected regardless of order, but the type selector didn't.

```
@debug(selector-unify('button', ':focus'));
// button:focus

@debug(selector-unify(':focus', 'button'));
// button:focus

@debug(selector-unify('[type="button"]', ':focus'));
// [type="button"]:focus

@debug(selector-unify(':focus', '[type="button"]'));
// :focus[type="button"]
```

Let me know if there's any more information I can provide!

glebm:
Introduces explicit order for selector unification and simplifies the implementation.

Spec https://github.com/sass/sass-spec/pull/1311
Fixes #2681

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -505,36 +505,28 @@ namespace Sass {
     return false;
   }
 
+  namespace {
+
+  int SelectorOrder(Simple_Selector_Ptr sel) {
+    if (Cast<Element_Selector>(sel)) return 1;
+    if (Cast<Id_Selector>(sel) || Cast<Class_Selector>(sel)) return 2;
+    if (Cast<Attribute_Selector>(sel)) return 3;
+    if (Cast<Pseudo_Selector>(sel)) return Cast<Pseudo_Selector>(sel)->is_pseudo_element() ? 6 : 4;
+    if (Cast<Wrapped_Selector>(sel)) return 5;
+    return 7;
+  }
+
+  }  // namespace
+
   Compound_Selector_Ptr Simple_Selector::unify_with(Compound_Selector_Ptr rhs)
   {
-    for (size_t i = 0, L = rhs->length(); i < L; ++i)
+    const size_t rsize = rhs->length();
+    for (size_t i = 0; i < rsize; ++i)
     { if (to_string() == rhs->at(i)->to_string()) return rhs; }
-
-    // check for pseudo elements because they are always last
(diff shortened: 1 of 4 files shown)


glebm:
There is two possible implementations here:

First commit (https://github.com/sass/libsass/pull/2737/commits/880cd261e434e6e4eeec7463695136ba48ab1b94) uses a functions that returns the unification order by matching on `typeid`.

Second commit (https://github.com/sass/libsass/pull/2737/commits/474e9a7483751960a7d986e777c48f98a518e801) adds a `unification_order` method to the `Selector` class. 

I prefer the second one, although the class hierarchy is not ideal for this, because `unification_order` is not defined for a few selector types but I don't want to introduce an intermediate type just for this.

xzyfer:
Added tests, test metadata, and kicked off CI

xzyfer:
I agree the final result is preferable.

PR #2738, opened 2018-11-18
Selector unification: use == for comparison

glebm:
Extracted from #2315 
@mgreter

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -509,7 +509,7 @@ namespace Sass {
   {
     const size_t rsize = rhs->length();
     for (size_t i = 0; i < rsize; ++i)
-    { if (to_string() == rhs->at(i)->to_string()) return rhs; }
+    { if (*this == *rhs->at(i)) return rhs; }
     const int lhs_order = this->unification_order();
     size_t i = rsize;
     while (i > 0 && lhs_order < rhs->at(i - 1)->unification_order()) --i;
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -2228,6 +2228,8 @@ namespace Sass {
     // dispatch to correct handlers
     virtual bool operator<(const Selector& rhs) const = 0;
     virtual bool operator==(const Selector& rhs) const = 0;
+    bool operator>(const Selector& rhs) const { return rhs < *this; };
+    bool operator!=(const Selector& rhs) const { return !(rhs == *this); };
     ATTACH_VIRTUAL_AST_OPERATIONS(Selector);
   };
   inline Selector::~Selector() { }
@@ -2314,6 +2316,7 @@ namespace Sass {
     {
       if (hash_ == 0) {
         hash_combine(hash_, std::hash<int>()(SELECTOR));
+        hash_combine(hash_, std::hash<int>()(simple_type()));
         hash_combine(hash_, std::hash<std::string>()(ns()));
         hash_combine(hash_, std::hash<std::string>()(name()));
       }


xzyfer:
I'm heading to bed for the night. Looks like a previous change is causing
CI to fail after landing of master. It's causing all subsequent PRs to
fail. I'll try to investigate further tomorrow unless you beat me to it.

glebm:
@xzyfer The first failing commit seems to be this one:

https://github.com/sass/libsass/commit/d09d09d9607d7593b92ba5daf3d8d947a4267f6b

glebm:
This is what I use for investigating by the way:

```bash
enable_asan() {
  export EXTRA_CXXFLAGS='-fsanitize=address -g -O0 -fno-omit-frame-pointer'
  export EXTRA_CFLAGS="$EXTRA_CXXFLAGS"
  export EXTRA_LDFLAGS="-fsanitize=address -lstdc++"
  export ASAN_OPTIONS="verbosity=1:symbolize=1"
  export CC=clang
  export CXX=clang++
}

enable_asan
cd sassc && make clean && cd - && make clean && make -j8 test
```

glebm:
>  in the small window where the configuration was incorrect,

Maybe not that small, it fails on d09d09d9607d7593b92ba5daf3d8d947a4267f6b as well.

glebm:
Also fails on https://github.com/sass/libsass/commit/38f4c3699d06b64128bebc7cf1e8b3125be74dc4 (July 2018). Passes on [@3.5.2](https://github.com/sass/libsass/tree/3.5.2) (March 2018).

glebm:
According to the bisection results, the culprit is https://github.com/sass/libsass/commit/205dc65d0f1af1f5f715080432192195d64256be

xzyfer:
Sorry I'm not following why your finding explain why master only began failing recently.

>Maybe not that small, it fails on d09d09d as well.

The change in d09d09d essentially reverts ab03dc0. This why I suggested the issue crept in between those two commits - making d225a09a obvious candidate.

It's worth noting that all following PRs to address leaks continue to fail in the same 
- #2738 (https://travis-ci.org/sass/libsass/jobs/456898779)
- #2739 (https://travis-ci.org/sass/libsass/jobs/456900757)
- #2742 (https://travis-ci.org/sass/libsass/jobs/457121156)
- #2744 (https://travis-ci.org/sass/libsass/jobs/457211061)
- #2745 (https://travis-ci.org/sass/libsass/jobs/457215304)
- #2747 (https://travis-ci.org/sass/libsass/jobs/457376054)

This suggest to me that the problem is more recent. I'm started debugging this by branching from f477cd6c and replaying the subsequent commits except my changes to `.travis.yml`.

xzyfer:
>The first failing commit seems to be the one that disables coverage for clang: d09d09d

Aha apologies, I missed this comment.

glebm:
My PRs each address individual leaks. With all 3 there are no more leaks.

xzyfer:
I don't get notified of edits via email. It's easier for me to respond in a timely manner if you can avoid editing comments :)

glebm:
OK, will do so in the future. CI has only been passing because it wasn't running ASAN (until you disable coverage). Memory leaks existed since March 17.

xzyfer:
Thanks for looking into this. I agree it is weird. I'm catching up on the backlog before going to bed.

glebm:
Looks like it doesn't work in a docker container without `--cap-add ptrace`. See https://github.com/google/sanitizers/issues/764.

Looks like there is no way to add container capabilities on Travis. However, we can simply set `sudo: required` in .travis.yml. See https://github.com/travis-ci/travis-ci/issues/9033

PR #2742, opened 2018-11-19
AST: Fix clang warnings, add const, remove virtual

glebm:
Mechanical changes:
* Fixes all clang warnings in ast.hpp and removes warnings suppression.
* Adds `const` to a bunch of methods.
* Removes `virtual` where unnecessary.
* Adds `override` where necessary (as part of fixing clang warnings).

Non-mechanical changes:
* Removes unused `Thunk` class.
* Makes `hash()` const by marking `hash_` as `mutable` (implementation detail).
* Makes `first()` and `last()` const, and adds `mutable_first()` and `mutable_first()` for the rare cases where mutability is necessary.
* `Complex_Selector` internally now uses `tail_` and `head_` directly instead of going through accessors (these accessors create a `SharedImpl` on every call).


glebm:
Rebased.

glebm:
This passes ASAN locally. Please merge this so that the Travis output becomes readable.

xzyfer:
Thanks, merged

PR #2752, opened 2018-11-20
Do not use `0` for initializing SharedImpl

glebm:
This is more readable and avoids creating a temporary `SharedImpl(nullptr)`

Later on this will allow us to initialized `SharedImpl` directly from `T*`, avoiding a redundant temporary `SharedImpl` construction and counter increment+decrement.

Also changes `is_superselector_of` signature to accept a const pointer, avoiding an unnecessary `SharedImpl` there.

ASAN passes locally.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -807,7 +807,7 @@ namespace Sass {
     return ns() < rhs.ns();
   }
 
-  bool Wrapped_Selector::is_superselector_of(Wrapped_Selector_Obj sub) const
+  bool Wrapped_Selector::is_superselector_of(Wrapped_Selector_Ptr_Const sub) const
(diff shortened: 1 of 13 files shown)


xzyfer:
I can confirm with this the `ASAN` are now gone :tada:

However it does appear that somewhere along the way some test regressions have crept in.

xzyfer:
I'm kicking off a master build to see if the test failures are present there. https://travis-ci.org/sass/libsass/builds/457712914

glebm:
The ASAN build on Travis is not fixed yet but I'm looking into it.

xzyfer:
Master is failing on the same tests so let's keep moving this forward

PR #2758, opened 2018-11-23
Update AppVeyor CI to Visual Studio 2015

### LINKED ISSUE #2757: msbuild CI is not using C++ 11
Got a syntax error from Windows CI when trying to use `constexpr` for:

```c++
constexpr char[] calc = "calc(";
```

https://ci.appveyor.com/project/sass/libsass/builds/20499178/job/bg4pm6n5jeu9eoql:

```
2>..\..\src\fn_colors.cpp(13): error C2065: 'constexpr' : undeclared identifier [C:\projects\libsass\sassc\win\sassc.vcxproj]
   2>..\..\src\fn_colors.cpp(13): error C2144: syntax error : 'char' should be preceded by ';' [C:\projects\libsass\sassc\win\sassc.vcxproj]
```

Looks like it isn't configured for C++11.

Asking Windows experts for help with updating the Windows build configuration.

glebm:
Fixes #2757 

Looks like sassc has been using VS2015 since the beginning.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2013
+os: Visual Studio 2015
 
 environment:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -9,22 +9,11 @@
 
 #include <cmath>
 #include <stdint.h>
-#if defined(_MSC_VER) && _MSC_VER >= 1800 && _MSC_VER < 1900 && defined(_M_X64)
-#include <mutex>
-#endif
 
 namespace Sass {
 
   double round(double val, size_t precision)
   {
-    // Disable FMA3-optimized implementation when compiling with VS2013 for x64 targets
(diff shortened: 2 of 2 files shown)


nschonni:
think we kept 2013 because of old node-sass/node compatibility, but if we're close to a new version we can probably do this @xzyfer

mgreter:
I don't really care about the bump, but I would keep the compiler workaround `ifdef` just in case somebody needs them, since IMO @am11 took quite some effort to figure it out and it doesn't really hurt to keep it in.

mgreter:
Btw. https://github.com/sass/libsass/pull/2648 seems closely related.

glebm:
> For now I'm in favour of keeping VS2013 as the default.

For how long? My main concern is that VS2013 doesn't have full C++ 11 support.

xzyfer:
Node Sass still needs to compile on VS2013 for some older versions of Node. We'll be dropping support for those versions in the next major version. I'd like to see that released Dec-Jan latest, but hopefully earlier depending on my availability.

glebm:
OK! That's not too bad.

Issue #2663, opened 2018-06-03
AddressSanitizer: null pointer dereference (SEGV) in Sass::Functions::selector_append (libsass/src/functions.cpp:2088:13)


glen-mac:
Hey there, I have discovered a null pointer dereference in libsass at: [functions.cpp:2088:13](https://github.com/sass/libsass/blob/60f839158eff1f1bb84cb44f0f0c4bce3ed86a6d/src/functions.cpp#L2088)

Found when fuzzing commit 60f839158eff1f1bb84cb44f0f0c4bce3ed86a6d of libsass, using commit [aa6d5c6](https://github.com/sass/sassc/commit/aa6d5c635ea8faf44d542a23aaf85d27e5777d48) of sassc as a harness.

Compile flags to reproduce: 
```
CC=clang CXX=clang++ CFLAGS='-fsanitize=address -g -O2 -fno-omit-frame-pointer' CXXFLAGS=$CFLAGS make -C sassc -j8
```

System information:
```
$ uname -a
Linux s127422 3.13.0-137-generic #186-Ubuntu SMP Mon Dec 4 19:09:19 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

This bug was found to be in libsass releases from [3.3.6](https://github.com/sass/libsass/tree/3.3.6) until the commit listed above.

You can find a collection of PoC files that trigger the bug [here](https://github.com/sass/libsass/files/2065572/functions_crashes.zip).

The full ASAN report is shown below:
```
↳ ./sassc.bin < crash.file
ASAN:DEADLYSIGNAL
=================================================================
==16225==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000098 (pc 0x0000005b8758 bp 0x7ffd5ab26780 sp 0x7ffd5ab25b80 T0)
    #0 0x5b8757 in Sass::Functions::selector_append(Sass::Environment<Sass::SharedImpl<Sass::AST_Node> >&, Sass::Environment<Sass::SharedImpl<Sass::AST_Node> >&, Sass::Context&, char const*, Sass::ParserState, Sass::Backtrace*, std::vector<Sass::SharedImpl<Sass::Selector_List>, std::allocator<Sass::SharedImpl<Sass::Selector_List> > >) /home/glenn/temp/libsass/src/functions.cpp:2088:13
    #1 0x6f6a04 in Sass::Eval::operator()(Sass::Function_Call*) /home/glenn/temp/libsass/src/eval.cpp:1030:18
[...]

PR #2760, opened 2018-11-24
Disallow parent selectors in `selector-append`

glebm:
Now correctly errors on inputs such as this one instead of crashing (see #2663):

```scss
div {
  a: selector-append('.menu', 'li', '&');
}
```

```
Error: Parent selectors aren't allowed here.
        on line 3:37 of ../../../tmp/test.scss, in function `selector-append`
        from line 3:6 of ../../../tmp/test.scss
>>   e: selector-append('.menu', 'li', '&');

   ------------------------------------^
```

### DIFF
--- a/src/fn_selectors.cpp
+++ b/src/fn_selectors.cpp
@@ -83,7 +83,10 @@ namespace Sass {
           str->quote_mark(0);
         }
         std::string exp_src = exp->to_string();
-        Selector_List_Obj sel = Parser::parse_selector(exp_src.c_str(), ctx, traces);
+        Selector_List_Obj sel = Parser::parse_selector(exp_src.c_str(), ctx, traces,
+                                                       exp->pstate(), pstate.src,
+                                                       /*allow_parent=*/false);
+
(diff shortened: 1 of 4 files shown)


nschonni:
Might need a PR to sass-spec for this one. I'm guessing there are already tests that are just disabled if this error wasn't supported before that need to be enabled

glebm:
No disabled tests get fixed by this so a test should be added to sass-spec.

xzyfer:
There isn't a sass-spec because there wasn't a reported issue. We were producing an error in this case, but not the right error

```
Error: $selectors: "&" is not a valid selector: Invalid CSS after "": expected selector, was "&" for `selector-append'
        on line 2 of test.scss
  Use --trace for backtrace
```

The SEGV from #2663 is a known issue with reparsing selector during eval. I've previously created the https://github.com/sass/libsass/labels/selector-validation tag for related issues. The fix for that is a fairly significant refactor.

glebm:
We're producing the same error as dart-sass

Issue #2667, opened 2018-06-03
AddressSanitizer: stack-overflow at Sass::Compound_Selector::operator== (ast.cpp:1058)


HongxuChen:
We found with our fuzzer some stack over flow errors when creating temp selector objects at Line 1508 in ast.cpp (45f50873962b7d1c66bd115ba6e644bdaaf6cac1) when compiled with Address Sanitizer (using sassc as the driver).

```
ASAN:SIGSEGV
=================================================================
==20277==ERROR: AddressSanitizer: stack-overflow on address 0x7ffdc834dfd8 (pc 0x7f37e2a4d80b bp 0x7ffdc834e8e0 sp 0x7ffdc834dfd0 T0)
    #0 0x7f37e2a4d80a  (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x2280a)
    #1 0x7f37e2ac44fe in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x994fe)
    #2 0x7f37e23f6790 in __gnu_cxx::new_allocator<Sass::SharedImpl<Sass::Simple_Selector> >::allocate(unsigned long, void const*) /usr/include/c++/5/ext/new_allocator.h:104
    #3 0x7f37e23f6790 in std::allocator_traits<std::allocator<Sass::SharedImpl<Sass::Simple_Selector> > >::allocate(std::allocator<Sass::SharedImpl<Sass::Simple_Selector> >&, unsigned long) /usr/include/c++/5/bits/alloc_traits.h:491
    #4 0x7f37e23f6790 in std::_Vector_base<Sass::SharedImpl<Sass::Simple_Selector>, std::allocator<Sass::SharedImpl<Sass::Simple_Selector> > >::_M_allocate(unsigned long) /usr/include/c++/5/bits/stl_vector.h:170
    #5 0x7f37e23f6790 in std::_Vector_base<Sass::SharedImpl<Sass::Simple_Selector>, std::allocator<Sass::SharedImpl<Sass::Simple_Selector> > >::_M_create_storage(unsigned long) /usr/include/c++/5/bits/stl_vector.h:185
[...]

PR #2764, opened 2018-11-25
Clean up `Compound_Selector::operator==`

glebm:
Implements equality for `Compound_Selector` in a cleaner and more efficient manner.
Similar change for `Selector_List`: #2739

This does not fix the stack overflow reported in #2667


xzyfer:
Thanks

PR #2769, opened 2018-11-29
improve compare operators

mgreter:
Next PR with some fixes on the heels of #2768

- Improvements for compare operators (important for selector unifying)
- Rename element to type selector and type enums (in line with CSS specs)


src/ast_selectors.cpp
```diff
@@ -154,19 +155,17 @@ namespace Sass {
...
     {
-      if (lbase->to_string() == rbase->to_string()) {
-        for (size_t i = 1, L = length(); i < L; ++i)
-        { lset.insert((*this)[i]->to_string()); }
-        for (size_t i = 1, L = rhs->length(); i < L; ++i)
-        { rset.insert((*rhs)[i]->to_string()); }
-        return includes(rset.begin(), rset.end(), lset.begin(), lset.end());
+      if (*lbase == *rbase) {
+        // create ordered sets for includes query
+        lset.insert(this->begin(), this->end());
```
glebm: `std::set` can be constructed directly from a range:

```c++
SelectorSet lset(lset.begin(), lset.end());
SelectorSet rset(rset.begin(), rset.end());
```

mgreter: Thx for the top, I'll see that I can incorporate it in a later commit!

src/ast_selectors.cpp
```diff
@@ -154,19 +155,17 @@ namespace Sass {
...
+        rset.insert(rhs->begin(), rhs->end());
+        return std::includes(rset.begin(), rset.end(), lset.begin(), lset.end(), OrderSelectors);
```
glebm: Consider `SelectorSet::key_compare` instead of `OrderSelector` (it might communicate the meaning better here)

mgreter: Can you elaborate a bit more? BTW. I struggled quite a bit here since I tried to reuse `struct OrderNodes` but failed miserably. Not sure if the way one defines the order for a `std::set` is incompatible to pass to `std::includes`, it feels like it should be possible to use the same way somehow??

src/ast_selectors.cpp
mgreter: Need to check if it is possible though, not sure we "always" want to construct it with those items.

src/ast_selectors.cpp
mgreter: Found it, one can pass `OrderNodes()` directly. Looks like I already found this one a few months ago for `std::sort` but forgot about it. Reverted it in my other PR.

mgreter:
Rebased on parent PR, kinda starting to loose the overview ... Ci is our friend ...

mgreter:
OK, I probably squashed it unexpectedly with a previous commit. History seems now OK again!

mgreter:
Rebased on master after roundhouse-cleanup-02 merge.

mgreter:
Rebased to fix another clang warning.

PR #2770, opened 2018-11-29
Roundhouse cleanup 03

mgreter:
On the heels of https://github.com/sass/libsass/pull/2769

Further AST split up and one small compare operator cleanup.

This would be all for today. I hope I did not regress anything due to rebasing.

### DIFF
--- a/src/ast_sel_cmp.cpp
+++ b/src/ast_sel_cmp.cpp
@@ -147,34 +147,17 @@ namespace Sass {
 
   bool Selector_List::operator== (const Selector_List& rhs) const
   {
-    // for array access
-    size_t i = 0, n = 0;
-    size_t iL = length();
-    size_t nL = rhs.length();
-    // create temporary vectors to sort them for compare
-    std::vector<Complex_Selector_Obj> l_lst = this->elements();
-    std::vector<Complex_Selector_Obj> r_lst = rhs.elements();
-    std::sort(l_lst.begin(), l_lst.end(), OrderNodes());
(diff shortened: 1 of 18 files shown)


mgreter:
Rebased after merging last PR.

src/ast_sel_cmp.cpp
```diff
@@ -147,34 +147,17 @@ namespace Sass {
...
+    lhs_set.reserve(length());
+    for (const Complex_Selector_Obj &element : elements()) {
+      lhs_set.insert(element.ptr());
     }
-    // not equal
-    return false;
+    for (const Complex_Selector_Obj &element : rhs.elements()) {
+        if (lhs_set.find(element.ptr()) == lhs_set.end()) return false;
+    }
+    return true;
```
mgreter: @glebm looks like it got reverted again with the previous commit and added here again (as I thought, due to the rebasing). Please check again that it is included in the final result. Sorry about the mess. I will also keep an eye out that the spec you mentioned will pass after all the refactoring!

mgreter:
Rebased since I forgot to update MSVC project files ...

mgreter:
Yay, no clang warnings and CI seems to get green ... still waiting for appveyor though ...

src/ast_sel_cmp.cpp
mgreter: CI still passing sass-spec/spec/core_functions/color/hsla/error/one_arg ...

PR #2771, opened 2018-11-29
Fix for remote win unc locations

mgreter:
Tested successfully locally on my windows PC.
Did not rebase since I expect the other PR to be merged ...


src/file.cpp
```diff
@@ -81,7 +81,10 @@ namespace Sass {
         wchar_t resolved[32768];
         // windows unicode filepaths are encoded in utf16
         std::string abspath(join_paths(get_cwd(), path));
-        std::wstring wpath(UTF_8::convert_to_utf16("\\\\?\\" + abspath));
+        if (abspath[0] != '/' && abspath[1] != '/') {
```
mgreter: After reviewing again, this check could regress if abspath = '\0`

PR #2773, opened 2018-11-29
appveyor CI: Reduce parallelism to 2

glebm:
Should make appveyor builds faster
Pretty sure appveyor OSS doesn't have more than 2 cores (maybe it's just
one), and ~~looks like it doesn't have enough RAM to build in parallel~~.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,9 +41,9 @@ install:
 build_script:
   - ps: |
       if ($env:Compiler -eq "mingw") {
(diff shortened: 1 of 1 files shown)


xzyfer:
The plan is call Pro FOSS (2 jobs). I don't see the details for that
specific plan listed anymore

glebm:
The build doesn't seem to be any slower, so I guess it has <2 cores.

Issue #1216, opened 2015-05-15
Integer truncation -- regression between 3.1 and 3.2 (and different from ruby sass)


asottile:
**EDIT** This post has been edited to reflect the actual worldstate and not the erroneous state when I originally created the ticket.  I had some wackiness in my environment which mislead me :(

Input:
```scss
a { width: 3.0px }
```

Output (libsass 3.1):
```css
a {
  width: 3px; }
```

Output (libsass 3.2.4):
```css
a {
  width: 3.0px; }
```

Ruby sass 3.4.12:
```css
a {
  width: 3.0px; }
```

asottile:
I've bisected it to this commit: 925220c57cfc9e3bc44c317eba473733194069c4

Looking at the commit it's not obvious to me what changed...

@xzyfer any ideas?

Using this (yes it leaks memory):

```c++
// entries/main1.cpp
#include "sass_context.h"
#include <cstring>
#include <iostream>
#include <string>


int main() {
    char* src = (char*)malloc(sizeof(char) * 50);
    strcpy(src, "a { width: 3.0px }");
    struct Sass_Data_Context* context = sass_make_data_context(src);
    sass_compile_data_context(context);
    struct Sass_Context* ctx = sass_data_context_get_context(context);
    std::string output = sass_context_get_output_string(ctx);
    std::cout << output << std::endl;
    return static_cast<int>(output != "a {\n  width: 3px; }\n");
}
```

```Makefile
C_SOURCES := $(wildcard *.c)
C_OBJECTS = $(patsubst %.c,build2/c/%.o,$(C_SOURCES))
CPP_SOURCES := $(wildcard *.cpp)
CPP_OBJECTS = $(patsubst %.cpp,build2/cpp/%.o,$(CPP_SOURCES))
MAIN_SOURCES := $(wildcard entries/*.cpp)
OUTS = $(patsubst entries/%.cpp,%.out,$(MAIN_SOURCES))

all: main1.out

build2/c/%.o: %.c
	@mkdir -p build2/c/
	gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I./ -c $^ -o $@ -c -O2 -fPIC -std=c++0x -Wall -Wno-parentheses

build2/cpp/%.o: %.cpp
	@mkdir -p build2/cpp/
[...]

Issue #550, opened 2014-10-17
Fractional numbers such as ".01" get displayed as "0.01"


Lekensteyn:
A minor issue, but it differs from ruby-sass. Consider this test case:

``` css
#foo {
  d: .0001px;
  f: .01px;
  d: -.0001px;
  f: -.01px;
}
```

The expected result after passing through sassc is the the identity. I actually got this instead:

``` css
#foo {
  d: 0.0001px;
  f: 0.01px;
  d: -0.0001px;
  f: -0.01px;
}
```

According to [CSS 3, 4.2. Numbers: the number type](http://www.w3.org/TR/css3-values/#numbers), a number is either an integer or zero or more integers followed by a dot, followed by one or more digits.

PR #2777, opened 2018-11-30
Add prelexer for decimal value with trailing zeroes

maximus8891:
Resolves: #1216

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1071,7 +1071,7 @@ namespace Sass {
       return SASS_MEMORY_NEW(Declaration, prop->pstate(), prop, parse_css_variable_value(), false, true);
     }
     lex < css_comments >(false);
-    if (peek_css< static_value >()) {
+    if (!peek_css< static_value_decimal_trailing_zeros >() && peek_css< static_value >()) {
       return SASS_MEMORY_NEW(Declaration, prop->pstate(), prop, parse_static_value()/*, lex<kwd_important>()*/);
     }
     else {
--- a/src/prelexer.cpp
+++ b/src/prelexer.cpp
@@ -1010,6 +1010,12 @@ namespace Sass {
           >
         >(src);
     }
+    const char* static_value_decimal_trailing_zeros(const char* src) {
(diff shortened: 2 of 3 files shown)


xzyfer:
Given that the referenced issue is closed, can you please provide an
example of exactly what this is fixing?

maximus8891:
There was a spec created for #1216 and it is still in the todo list. I wanted to close it.

I can updated the spec to match the current output and close this PR. However, first I wanted to check that the behaviour is correct. Given `3.0px`, the output is `3.0px`. Would not contracting `3.0px` to `3px` save space?

In #550 it worked the other way round. `.001px` was extended to `0.001px` adding a character.

nschonni:
@nex3 Dart seems to be pruning trailing zeros. Is that a new feature or a bug?

nex3:
@nschonni The answer to that question is multi-layered! Strap in!

How an implementation chooses to serialize its final AST to CSS is up to that implementation, as long as the CSS is semantically identical to the specified CSS. As far as CSS is concerned, `3.0px` and `3px` are the same value, so implementations are free to switch between them as they choose.

Ruby Sass does this too! For example, Ruby Sass will compile `a { b: (3.0px); }` to `a { b: 3px; }`. *However*, Ruby Sass has special logic for "static" property values that look like they don't include any SassScript. As a mitigation for Ruby Sass's incredible lack of speed, it avoids fully parsing or interpreting static values and instead passes them on as-is to the CSS output. So if you don't include parentheses, Ruby Sass won't change the input at all and you'll get `3.0px`.

Dart Sass, on the other hand, is much faster than Ruby Sass and as a consequence has no special handling for static property values. It fully parses every value and converts it to CSS using its normal logic, which means that its numbers *always* have trailing zeroes trimmed.

So it's not a bug, but it's not *exactly* a feature either. I guess I'd call it a "non-semantic implementation difference".

xzyfer:
I'm in favour of updating the LibSass output of the spec and moving on.

maximus8891:
Makes sense, especially that there are other specs that have conversions from `opacity: 1.0;` to `opacity: 1.0;`.

Issue #1079, opened 2015-04-09
Currupted error message on a broken code


saper:
Just tried this:

Given the following `hi` file:

```
#{hdr(2,5)} { color: #08c; }
```

    $ sassc hi

gives:

````
Error: unterminated argument to hdr(...)
        on line 1 of [NULL]
>> ¥ö
   ----^
````

Where the funny characters are 0xA5 0xF6 hex (offset 000048):

```
00000000  45 72 72 6f 72 3a 20 75  6e 74 65 72 6d 69 6e 61  |Error: untermina|
00000010  74 65 64 20 61 72 67 75  6d 65 6e 74 20 74 6f 20  |ted argument to |
00000020  68 64 72 28 2e 2e 2e 29  0a 20 20 20 20 20 20 20  |hdr(...).       |
00000030  20 6f 6e 20 6c 69 6e 65  20 31 20 6f 66 20 5b 4e  | on line 1 of [N|
00000040  55 4c 4c 5d 0a 3e 3e 20  a5 f6 0a 20 20 20 2d 2d  |ULL].>> ...   --|
00000050  2d 2d 5e 0a                                       |--^.|
00000054
```

**Comment by xzyfer** [comment]
This is also causing issues for https://github.com/sass-compatibility/sass-compatibility.github.io/pull/36#issuecomment-94145993

**Comment by mgreter** [comment]
This might be fixed with https://github.com/sass/libsass/pull/1180 ... can someone test?

**Comment by mgreter** [comment]
Closing this since I get this result with latest master:
```
Error: unterminated argument to hdr(...)
        on line 1 of sass.scss
>> hdr(2, 5){
   ----^
```

**Comment by maximus8891** [comment]
The latest master doesn't throw an error anymore. Returns:
```
hdr(2, 5) {
  color: #08c;
}
```

PR #2783, opened 2018-12-03
Disallow optional pseudo_prefix

maximus8891:
Resolves: #1079 

Although the issue is closed, the following input does not currently throw an error.
```
#{hdr(2,5)} { color: #08c; }
```
Spec https://github.com/sass/sass-spec/pull/1324


xzyfer:
Thanks?

Issue #2778, opened 2018-12-01
Invalid numbers comparison


alexander-shipilov:
Invalid numbers comparison if one of them is a result of math operations

## input.scss

```scss
$n: (1 + (11 / 1000)) * 1000;

@debug unquote("$n = #{ $n }");
@debug unquote("$n == 1011  :  ")                        $n == 1011;
@debug unquote("$n == (1 + (11 / 1000)) * 1000  :  ")    $n == (1 + (11 / 1000)) * 1000;
```

## Actual results

[todo]: # (update version info!)

[libsass 3.X.y][1]
```cmd
$ node-sass test.scss
test.scss:3 DEBUG: $n = 1011
test.scss:4 DEBUG: $n == 1011  :   false
test.scss:5 DEBUG: $n == (1 + (11 / 1000)) * 1000  :   true
```

## Expected result

ruby sass 3.X.y
```cmd
sass test.scss
test.scss:3 DEBUG: $n = 1011
test.scss:4 DEBUG: $n == 1011  :   true
test.scss:5 DEBUG: $n == (1 + (11 / 1000)) * 1000  :   true
```
version info:
```cmd
$ node-sass --version
node-sass       4.10.0  (Wrapper)       [JavaScript]
libsass         3.5.4   (Sass Compiler) [C/C++]
```
http://libsass.ocbnet.ch/srcmap/#JG46ICgxICsgKDExIC8gMTAwMCkpICogMTAwMDsKCi5yZXN1bHQgewogIGNvbnRlbnQ6ICIjeyRufSAjeyAkbiA9PSAxMDExIH0gI3sgJG4gPT0gKDEgKyAoMTEgLyAxMDAwKSkgKiAxMDAwIH0iLAp9Cg==

mgreter:
Can be boiled down to:

```scss
test {
  eq: ((1 + (11 / 1000)) * 1000) == 1011;
}
```

maximus8891:
Maybe this approach?
Would it be too much overhead?
https://floating-point-gui.de/errors/comparison/

```java
	public static boolean nearlyEqual(float a, float b, float epsilon) {
		final float absA = Math.abs(a);
		final float absB = Math.abs(b);
		final float diff = Math.abs(a - b);

		if (a == b) { // shortcut, handles infinities
			return true;
		} else if (a == 0 || b == 0 || diff < Float.MIN_NORMAL) {
			// a or b is zero or both are extremely close to it
			// relative error is less meaningful here
[...]

PR #2792, opened 2018-12-14
Fix number comparison for floats/doubles by calculations

mgreter:
Fixes https://github.com/sass/libsass/issues/2778

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -38,7 +38,7 @@ namespace Sass {
 
   // ToDo: should this really be hardcoded
   // Note: most methods follow precision option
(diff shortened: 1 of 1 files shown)


xzyfer:
@mgreter do you want to update this to match https://github.com/sass/libsass/issues/2778#issuecomment-448966061?

mgreter:
> 
> 
> @mgreter do you want to update this to match [#2778 (comment)](https://github.com/sass/libsass/issues/2778#issuecomment-448966061)?

We don't keep and pass the setting for precision around everywhere, so this is not easily possible.

PR #2797, opened 2018-12-20
`const Expression_Ptr` -> `Expression_Ptr`

glebm:
The `const` there was useless, as it didn't mean the constness of Expression but merely that `val` pointer could not be reassigned within the function (due to how pointer typedefs and consts interact).

Assignee: @mgreter

If we actually want the Expression to be const there, we'll need to add const conversion functions to Color.

### DIFF
--- a/src/values.cpp
+++ b/src/values.cpp
@@ -7,7 +7,7 @@
 namespace Sass {
 
   // convert value from C++ side to C-API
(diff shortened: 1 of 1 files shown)


glebm:
wrong fix, actually have to make it const for C ABI compat

PR #2798, opened 2018-12-20
 Remove _Ptr(_Const) typedefs

### LINKED ISSUE #2785: Proposal: Remove _Ptr(_Const) typedefs
The libsass codebase uses typedefs for pointers and const pointers, such as `Expression_Ptr`.

However, pointers should be handled with care but it's easy to miss a pointer if it doesn't look like one.

@xzyfer @mgreter What do you think about removing the `_Ptr` and `_Ptr_Const` typedefs?

This can be done with the following command (GNU sed, for macOS and FreeBSD use `sed -i ''`):

```bash
find src \( -name '*.cpp' -o -name '*.hpp' \) -exec sed -i \
'/typedef.*_Ptr[_;]/d;'\
's/\([A-Za-z_]*\)_Ptr_Const/const \1*/g;'\
's/\([A-Za-z_]*\)_Ptr/\1*/g;'\
's/\([a-z]\)##\*/\1*/g' '{}' \+
```

glebm:
Removes these typedefs because pointers should be handled with care. Having them visually stand out in the code helps us avoid common pitfalls, such as accidentally using `Ptr` instead of `Obj`.

This change has been automatically generated using the following command:

```bash
find src \( -name '*.cpp' -o -name '*.hpp' \) -exec sed -i \
'/typedef.*_Ptr[_;]/d;'\
's/\([A-Za-z_]*\)_Ptr_Const/const \1*/g;'\
's/\([A-Za-z_]*\)_Ptr/\1*/g;'\
's/\([a-z]\)##\*/\1*/g' '{}' \+
```

Fixes #2785


glebm:
Will reopen once `ast_node_to_sass_value` is fixed.

PR #2800, opened 2018-12-20
Avoid modifying Expression in ast_node_to_sass_value

glebm:
Previously, the argument to `ast_node_to_sass_value` was a `const Expression_Ptr`.

The `const` there was useless, as it didn't mean the constness of Expression but merely that `val` pointer could not be reassigned within the function (due to how pointer `typedef` and `const` interacts).

Assuming the intended argument type was `const Expression_Ptr *`, I've changed it to `Expression_Ptr_Const`.

This required splitting `to{RGBA,HSLA}(bool)` into separate functions, so that the one that always copies could be marked as `const`.

@mgreter

### DIFF
--- a/src/ast2c.cpp
+++ b/src/ast2c.cpp
@@ -20,7 +20,10 @@ namespace Sass {
   { return sass_make_color(c->r(), c->g(), c->b(), c->a()); }
 
   union Sass_Value* AST2C::operator()(Color_HSLA_Ptr c)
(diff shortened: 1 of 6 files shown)


xzyfer:
Seems reasonable to me.

glebm:
@mgreter ping

glebm:
@xzyfer Let's merge this, it's the only blocker for getting rid of the typedefs.

PR #2801, opened 2018-12-20
[WIP] Enable LTO: 500 KiB smaller binary, 15% faster build

glebm:
Metrics:

* `time make -j1 sassc`: 65s -> 56s
* `libsass.a` size: 6.7 MiB ⭢ 24 MiB
* `libsass.so` size: 3.2 MiB ⭢ 2.7 MiB
* sassc binary size: 2.7 MiB ⭢ 2.2 MiB
* Compiling 10 copies of bootstrap 4: 0.45s ⭢ 0.40s

LTO requires a "recent" (2010+) version of `ar` and `ranlib` (to have LTO support by default). Alternatively, one can use `gcc-ar` / `llvm-ar`. In make and automake we try to detect the compiler and set `AR` and `RANLIB`.

LTO can be disabled by setting `NO_LTO=1`.

I still haven't figured out some builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: cpp
 
 # ASan needs ptrace support which currently requires `sudo: required`.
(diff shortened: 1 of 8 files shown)


glebm:
I think LTO can work on Travis if we install a newer version of binutils (https://packages.ubuntu.com/xenial/binutils), at least for GCC

glebm:
Closing, as this is likely to cause build issues for some older distributions.

PR #2802, opened 2018-12-20
travis: Disable coverage for static OSX build

glebm:
There is no need to run this coverage build as we don't have any OSX-specific code.

This build is very slow, and disabling it reduces our overall travis build time by ~10 minutes.


xzyfer:
Would we still want to have a static build for OSX?

xzyfer:
Thanks!

PR #2806, opened 2018-12-25
Avoid clang warning: expression with side effects will be evaluated despite being used as an operand to 'typeid'

glebm:
This is not a useful warning here but simply extracting `*schema->at(0)` to a variable avoids it.

Warning example: https://travis-ci.org/sass/libsass/jobs/471245025
Refs #1523


glebm:
@xzyfer ping

xzyfer:
Thanks

Issue #2519, opened 2017-12-30
cannot buils libsass on Solaris 11


drook:
libsass cannot be built on Solaris 11 (same story on less recent ones and probably on SmartOS/OmniOS) because Solaris has a SEC macro somewhere inside the system headers:

````build log
In file included from /usr/include/sys/select.h:28:0,
                 from /usr/include/sys/types.h:637,
                 from /usr/include/unistd.h:13,
                 from src/file.cpp:10:
src/units.hpp:38:5: error: expected identifier before numeric constant
     SEC = TIME,
     ^
src/units.hpp:38:5: error: expected ‘}’ before numeric constant
src/units.hpp:38:5: error: expected unqualified-id before numeric constant
In file included from src/ast.hpp:53:0,
                 from src/listize.hpp:7,
                 from src/functions.hpp:4,
                 from src/sass_functions.hpp:6,
                 from src/file.cpp:23:
src/units.hpp:69:1: error: expected declaration before ‘}’ token
 }
 ^
gmake: *** [src/file.o] Error 1
````

This can easily be fixed by changing all of the SEC entries in units.hpp/cpp to SECONDS. Please fix this (since the fix is easy and obvious), - I know Solaris is almost dead by now, and you don't officially support it, but it is still used in many places, and this bug affects the automatic building of node.js node-sass module, which otherwise would build it fine (yeah, I know node-sass has libsass bundled, so anyway the fix would work only on newer versions, but there's no other option anyway). Other than that - libsass seems to be fully working with this fix, at least in what comes when node-sass is using it's bindings.

drook:
````diff
--- units.cpp.orig      2017-12-30 10:12:57.611866142 +0300
+++ units.cpp   2017-12-30 10:13:30.501632897 +0300
@@ -91,7 +91,7 @@
     else if (s == "rad")  return UnitType::RAD;
     else if (s == "turn") return UnitType::TURN;
     // time units
-    else if (s == "s")    return UnitType::SEC;
+    else if (s == "s")    return UnitType::SECONDS;
     else if (s == "ms")   return UnitType::MSEC;
     // frequency units
     else if (s == "Hz")   return UnitType::HERTZ;
@@ -120,7 +120,7 @@
       case UnitType::RAD:     return "rad";
       case UnitType::TURN:    return "turn";
       // time units
-      case UnitType::SEC:     return "s";
+      case UnitType::SECONDS:     return "s";
       case UnitType::MSEC:    return "ms";
       // frequency units
[...]

PR #2817, opened 2019-01-14
Ensure sass.hpp is included before system headers

alyssais:
Without this, system headers on Solaris will export the SEC macro, which clashes with the SEC constant defined in src/units.hpp.

See https://github.com/sass/libsass/issues/2519.

---

I wonder if there might be a better way to address this going forward... This has already been fixed once in the past, but has crept back. Perhaps we could use a compiler flag, rather than using sass.hpp to undefine the macro.

Thoughts?

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -1,7 +1,10 @@
 #ifndef SASS_AST_H
 #define SASS_AST_H
 
+// sass.hpp must go before all system headers to get the
+// __EXTENSIONS__ fix on Solaris.
 #include "sass.hpp"
+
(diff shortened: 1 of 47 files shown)


glebm:
Looks good to me

alyssais:
> Honestly I think even a simple comment would go a long way to stopping this regressing again. Memories are short.

Where would it go? Would it have to be repeated once per file?

xzyfer:
Yep, that's fine

alyssais:
Thank you! While I'm here, any news on a release?

Issue #2831, opened 2019-02-16
function-exists() and mixin-exists() return false for local definitions


nex3:
Both `function-exists()` and `mixin-exists()` should return `true` for functions or mixins (respectively) defined in local scopes, as so:

```scss
.scope {
  @function local-function() {@return null}
  @debug function-exists(local-function);

  @mixin local-mixin() {}
  @debug mixin-exists(local-mixin);
}
```

Both of these `@debug`s should print `true`, but they currently print `false`.

PR #2845, opened 2019-03-17
Fix mixin-exists and function-exists in local scopes

xzyfer:
Spec https://github.com/sass/sass-spec/pull/1362
Fixes https://github.com/sass/libsass/issues/2831


xzyfer:
sass-spec failures are unrelated to this patch.

PR #2849, opened 2019-03-25
Newlines: Parse \f and normalize in comments

### LINKED ISSUE #2843: Treat FF, CR, and CR LF as LF
[The CSS spec](https://drafts.csswg.org/css-syntax-3/#input-preprocessing) says that "U+000D CARRIAGE RETURN (CR) code points, U+000C FORM FEED (FF) code points, or pairs of U+000D CARRIAGE RETURN (CR) followed by U+000A LINE FEED (LF)" should be treated as U+000A LINE FEED characters, but LibSass seems to treat them like any other character in some circumstances.

In particular, all of these characters (or character sequences) are passed through as-is if they appear in comments. If they're used to terminate a line in the indented syntax, CR and CR LF work as expected but FF fails.

glebm:
Also:

1. Moves the `normalize*` functions and `rtrim` to a separate object file,
   so that they can be unit tested quickly and easily.
2. Changes `string_to_output` to also replace `\r\n`, fixing Windows tests.
3. Improves other `normalize_*` functions and adds tests.

Fixes #2843 

sass-spec PRs:

1. Bug fixes to make it run: sass/sass-spec#1365
2. Enabling the newly passing tests: sass/sass-spec#1366

### DIFF
--- a/Makefile.conf
+++ b/Makefile.conf
@@ -29,6 +29,7 @@ SOURCES = \
 	bind.cpp \
 	file.cpp \
 	util.cpp \
+	util_string.cpp \
 	json.cpp \
 	units.cpp \
 	values.cpp \
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -3,6 +3,7 @@
 #include "context.hpp"
 #include "output.hpp"
 #include "emitter.hpp"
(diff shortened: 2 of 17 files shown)


xzyfer:
Thank you fixing the Windows failures

PR #2852, opened 2019-03-28
If OpenBSD, print message showing how to use `clang` instead of `gcc`

ae84:
Kinda hard for ordinary users to figure this out by themselves.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,10 @@ endif
 CXXFLAGS += -std=c++11
 LDFLAGS  += -std=c++11
 
+ifeq ($(OS),OpenBSD)
+    @echo '*** HINT ***: OpenBSD uses `clang` instead of `gcc`, to fix this please visit: https://github.com/sass/libsass/blob/master/docs/build.md#compiling-with-clang-instead-of-gcc'
+endif
+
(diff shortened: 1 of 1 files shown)


glebm:
The default `CXX` is set to `c++` in this Makefile, which on OpenBSD is clang.

Are you compiling from master? 

Before https://github.com/sass/libsass/pull/2708, it did default to `g++`, but this workaround should be no longer necessary for the current version of this Makefile.

ae84:
@glebm I was using RubyGems to install `sass-rails` for my Rails app.

I was just hoping my proposed change to the Makefile could help make things a bit more user-friendly, as I don't think the average user knows that OpenBSD uses clang instead of gcc, let alone that he or she needs to set new env variables in order to fix the problem.

```
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
```

No big deal though. It's not a problem for me anymore, was just thinking of those who might be following in my footsteps as OpenBSD is becoming increasingly more popular.

Thanks!

glebm:
> I was using RubyGems to install sass-rails for my Rails app.

That version doesn't have the fix yet. This PR is unnecessary, as the issue is already fixed on master in a better way (the user won't have to do anything at all).

Issue #2854, opened 2019-04-02
Copy spec/libsass/Sáss-UŢF8 test here


glebm:
See discussion in https://github.com/sass/sass-spec/pull/1365.

TODO:
1. [x] Replicate that test here.
2. [x] Re-HRXify the test in the sass-spec repo when done.

mgreter:
Can this be closed?

PR #2855, opened 2019-04-04
Copy Unicode PWD tests from sass-spec

glebm:
As sass-spec is migrating to HRX, we need to copy this test here so we
can run it directly via `sassc` instead of the sass-spec runner.

Refs #2854
Refs https://github.com/sass/sass-spec/pull/1365

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -80,9 +80,9 @@ test_script:
         echo "spec runner not found (compile error?)"
         exit 1
       }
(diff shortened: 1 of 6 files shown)


asottile:
fwiw this is causing a little bit of an issue with libsass-python: https://github.com/sass/libsass-python/issues/296

I'm trying to figure out a workaround 🤞

PR #2862, opened 2019-04-08
Implement `capture_stderr` C-API option

mgreter:
As an embeddable library we should try to play nice with
our parents resources. It is therefore desirable to have
a way to capture messages on stderr. This makes it easier
and safer to use libsass in eg. threaded environments.

Core errors (mainly memory allocation problems) and
dev/debug data is still printed to stderr directly.

Introduces two new functions to the C-API:
- `sass_option_set_suppress_stderr`
- `sass_context_get_stderr_string`

### DIFF
--- a/docs/api-context-internal.md
+++ b/docs/api-context-internal.md
@@ -51,6 +51,10 @@ struct Sass_Options : Sass_Output_Options {
   // Treat source_string as sass (as opposed to scss)
   bool is_indented_syntax_src;
 
+  // If this options is set, nothing will be printed to stderr anymore
+  // The aggregated output on stderr can be fetched via stderr_string
+  bool suppress_stderr;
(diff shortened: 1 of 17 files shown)


xmo-odoo:
Alternate option providing more flexibility for the caller: a warnings callback, which would default to printing the message to stderr?

mgreter:
Thx @xmo-odoo, that is certainly something I wouldn't be opposed to. Ongoing LibSass 4.0 branch (#3135) already fixes the issue by never directly writing to stderr/stdout if not explicitly requested to. But that branch also complicated error reporting a little by introducing (all optional) ANSI colors, Unicode and line wrapping in the output. Going to close this here as work in this area will certainly not land in any pre 4.0 release.

Issue #2863, opened 2019-04-11
str-slice removes quotes


PhilippSoehnlein:
## input.scss
```scss
$foo: "'abcd'"; // '"abcd"' also doesn't work

@error str-slice( $foo, 1, 2 ); // not real code, @error is just used to get raw value of $foo
```

## Actual results

[todo]: # (update version info!)

[libsass 3.5.4][1]
> ab on line 3 at column 1

## Expected result

> 'a on line 3 at column 1

ruby sass 3.5.6
> 'a on line 3 at column 1

version info:
```cmd
$ node-sass --version
node-sass	4.10.0	(Wrapper)	[JavaScript]
libsass  	3.5.4	(Sass Compiler)	[C/C++]
```

[1]: http://libsass.ocbnet.ch/srcmap/#JGZvbzogIidhYmNkJyI7IC8vICciYWJjZCInIGFsc28gZG9lc24ndCB3b3JrCgpAZXJyb3Igc3RyLXNsaWNlKCAkZm9vLCAxLCAyICk7Cg==

## Further context
We have a string-utility function in our code base, which slices bigger strings in smaller pieces using `str-slice`. It works fine, except when the piece that should be sliced out of a bigger string coincidentally starts with quotes and ends with the same quotes, then `str-slice` returns the sliced string without the quotes, which makes our function return incorrect stuff.

PR #2864, opened 2019-04-11
Remove unquoting for str_slice, str-insert and str-index

mgreter:
Fixes https://github.com/sass/libsass/issues/2863

We need to come up with some spec tests for the three string functions that check the behavior
if "quoted" strings are passes. So far the spec tests did not react/fail because to these changes.

### DIFF
--- a/src/fn_strings.cpp
+++ b/src/fn_strings.cpp
@@ -102,10 +102,8 @@ namespace Sass {
       try {
         String_Constant* s = ARG("$string", String_Constant);
         str = s->value();
-        str = unquote(str);
         String_Constant* i = ARG("$insert", String_Constant);
         std::string ins = i->value();
-        ins = unquote(ins);
         double index = ARGVAL("$index");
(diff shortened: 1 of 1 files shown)


mgreter:
Rebased to include CI spec runner fixes

src/fn_strings.cpp
```diff
@@ -102,10 +102,10 @@ namespace Sass {
       try {
         String_Constant* s = ARG("$string", String_Constant);
         str = s->value();
-        str = unquote(str);
+        // str = unquote(str);
```
glebm: Consider removing these instead of commenting them out.

mgreter: I don't know yet if we can or can't remove those ... WIP ...

glebm: Ah, sorry, missed the WIP tag

mgreter:
My local tests seems to strongly indicate that we can remove the explicit unquoting
for those functions, but I'm still missing the confirmation from CI probe-todo run.

PR #2865, opened 2019-04-12
Remove passing version flag to spec runner

mgreter:
See https://github.com/sass/sass-spec/pull/1375


mgreter:
One failing job from travis seems to be an issue with travis mac instance ...

Issue #2823, opened 2019-02-11
test directory appears unused


QuLogic:
The `test` directory contains files that are never compiled, nor are the referenced in any Makefile. They seem to be broken as well, using include paths that are no longer correct. The `test` directory also shadows the `test` Makefile target, causing `make test` to sometimes report 'nothing to do' instead of running tests.

Can these files be removed now?

mgreter:
Would be fine with me, made PR to remove those tests ...

mgreter:
Obsolete tests have been removed

PR #2866, opened 2019-04-12
Remove obsolete unit tests

mgreter:
Addresses https://github.com/sass/libsass/issues/2823


test/Makefile
```diff
@@ -1,24 +0,0 @@
-CXX ?= c++
```
glebm: The tests mentioned in this Makefile and the Makefile itself are not obsolete, they were all both added recently by me. They also run in CI.

The other test files in this directory can be deleted.

mgreter: Thanks for the heads up, adjusted

PR #2871, opened 2019-04-12
Fix spec testing on mac with shared build

mgreter:
Since spec test runner now executes the tests in the
directory of the spec test, the default library path
resolution does no longer work with relative paths.


mgreter:
No clue why it does not work when `DYLD_LIBRARY_PATH` is defined before
calling `make`; we have to pass the env-setting to `make` in order to pass!?
Anyway, most important that it passes and CI gets green again!

PR #2873, opened 2019-04-13
Disallow non-string argument to `quote`

### LINKED ISSUE #2811: quote() function should reject non-string arguments
The `quote()` function in LibSass accepts arguments of any type. It should only accept string arguments, and produce errors for arguments of any other type.

https://github.com/sass/sass/issues/2597#issuecomment-451480529

glebm:
Tests: https://github.com/sass/sass-spec/pull/1380
Fixes #2811

### DIFF
--- a/src/fn_strings.cpp
+++ b/src/fn_strings.cpp
@@ -65,19 +65,13 @@ namespace Sass {
     Signature quote_sig = "quote($string)";
     BUILT_IN(sass_quote)
     {
-      AST_Node_Obj arg = env["$string"];
-      // only set quote mark to true if already a string
-      if (String_Quoted* qstr = Cast<String_Quoted>(arg)) {
-        qstr->quote_mark('*');
-        return qstr;
-      }
-      // all other nodes must be converted to a string node
-      std::string str(quote(arg->to_string(ctx.c_options), '"'));
-      String_Quoted* result = SASS_MEMORY_NEW(String_Quoted, pstate, str);
+      const String_Constant* s = ARG("$string", String_Constant);
+      String_Quoted *result = SASS_MEMORY_NEW(
+          String_Quoted, pstate, s->value(),
+          /*q=*/'\0', /*keep_utf8_escapes=*/false, /*skip_unquoting=*/true);
       result->quote_mark('*');
       return result;
     }
-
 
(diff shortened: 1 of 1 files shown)


src/fn_strings.cpp
```diff
@@ -65,20 +65,12 @@ namespace Sass {
...
-      // only set quote mark to true if already a string
-      if (String_Quoted* qstr = Cast<String_Quoted>(arg)) {
-        qstr->quote_mark('*');
-        return qstr;
-      }
-      // all other nodes must be converted to a string node
-      std::string str(quote(arg->to_string(ctx.c_options), '"'));
-      String_Quoted* result = SASS_MEMORY_NEW(String_Quoted, pstate, str);
+      const String_Constant* s = ARG("$string", String_Constant);
+      String_Quoted* result = SASS_MEMORY_NEW(String_Quoted, pstate, s->value(), '"');
```
mgreter: AFAICS you are setting the quotemark to `"` on creation, but it is reset one line below to `*`.
IMO you can just use the following and remove the lines below:
`return SASS_MEMORY_NEW(String_Quoted, pstate, s->value(), '*');`

glebm: Do you know what the `*` quote mark means? Perhaps it should be `"`?

glebm: Turns out the `String_Quoted` constructor does a bunch of non-trivial things and passing it a quote mark doesn't actually set it.

Updated, PTAL.

mgreter: A * should mean that on output we choose the best quotemarks automatically.

glebm:
Ah, should we change it to `*` then or always double-quote?

mgreter:
Hehe, depends on what the expected output is, but basically it should be `*` IMO, but old
ruby sass sometimes preserved the original quotemark and sometimes didn't ...

glebm:
I checked the output of Dart Sass and it also chooses a different quote depending on the input, so I've changed it to `*`.

glebm:
Adding more tests as well

glebm:
Added tests for quote detection to https://github.com/sass/sass-spec/pull/1380 and rebased. This should be good to merge.

PR #2875, opened 2019-04-13
Run C++ unit tests on AppVeyor

glebm:
Previously, these only ran on Travis.


xzyfer:
@glebm big thanks for all your work. I've given you commit access to unblock you.

glebm:
@xzyfer Thanks!

glebm:
@mgreter I don't have any msbuild / Windows experience. I've hand-coded these vcxproj files based on MSDN tutorials, please take a look.

glebm:
@xzyfer I don't seem have commit access! No big deal though, as I'd like all my PRs reviewed anyway, but it'd be nice to be able to set labels on issues etc.

xzyfer:
Sorry I misclicked on mobile. You should have write access now.

Your changes are sensible so I'd prefer not to block your windows of productivity. Please to request reviews when you think it makes sense. I personally follow changes via Github emails so I'll pop up if I have string feelings about something.

Additionally this should give you the power to kill AppVeyor and Travis CI builds which can been a huge time saver.

xzyfer:
I know nothing about vs studio but agree with the intent of the change :)

test/test_shared_ptr.vcxproj
```diff
@@ -0,0 +1,8 @@
+<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
nschonni: I think "ToolsVersion" is lower because we're still stuck on 2013. ? Maybe 12.0

glebm: I took it from here but I think you're right: https://github.com/sass/libsass/blob/9740de67d01d1f9da0278dca1ffad984304f9124/win/libsass.vcxproj#2

This [MSDN article](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-toolset-toolsversion?view=vs-2019) says:

> Visual Studio 2015 uses ToolsVersion 14.0

PR #2888, opened 2019-05-19
[build] enable concurrent compilation

ahorek:
this significantly reduce build time

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ ifndef COVERAGE
   CFLAGS   += -O2
   CXXFLAGS += -O2
   LDFLAGS  += -O2
+  MAKEFLAGS += "-j 0"
 else
   CFLAGS   += -O1 -fno-omit-frame-pointer
   CXXFLAGS += -O1 -fno-omit-frame-pointer


mgreter:
IMO this is not really portable, what about mingw on windows?

ahorek:
yes, I was worried about the same thing. mingw on windows doesn't really support concurrent compilation, but it won't fail, the option is just ignored.

mgreter:
Actually mingw does support concurrent compilation, but `nproc` is not installed and will therefore fail.

ahorek:
`-j` option is supported, but it has no effect on windows, it's still single threaded (gcc version 8.2.1 mingw)

I've changed that to `-j 0` (autodetect), so now the code doesn't rely on nproc

mgreter:
I use gmake quite regularly and used dmake before (it uses `-PXX` insteal of `jXX`).
It definitely does compile  in parallel, not sure which make flavor you mean that ignores it. 
Otherwise I'm fine if `0` means autodetect, but could you please add a more descriptive
title/description to the commit? Always useful to have meaningful git logs! Thanks!

ahorek:
done, I've added a benchmark aswell. On my AMD FX-8300 8C it's 4,6x faster.

xzyfer:
Thanks

richardwhiuk:
Surely this is the wrong fix - shouldn't the caller control the fanout by passing `make -j 0` if they wan this behaviour.

Currently this causes unbounded fan out, which is deeply wrong, and it ignores a caller passed in setting - e.g. `make --jobs 1` will be overriden by this change  - which is happening when the Rust build attempts to control fanout to the number of CPUs - see https://github.com/compass-rs/sass-rs/issues/53

xzyfer:
Sounds reasonable. Open to reverting this change.

ahorek:
ok, sorry, I tested again and you're right. -j 0 is indeed unlimited. The original nprocs limit was removed because it's hard to get the value safely on all platforms.  I reverted the change in https://github.com/sass/libsass/pull/2977

btw optflags https://github.com/sass/libsass/pull/2977/files#diff-b67911656ef5d18c4ae36cb6741b7965L19 also can't be overridden.

PR #2907, opened 2019-06-13
Prepare library to be used with wasm32-wasi

yurydelendik:
There is ongoing efforts to support [WASI](https://wasi.dev/) for different runtimes/hosts [such as node.js](https://github.com/nodejs/node/pull/27850). Here is the changes I made to make compilation work for clang with "wasm32-wasi" target and the wasi-sdk. The "wasi_polyfill.*" code will be removed in final PR.

There is also minor change in the sassc: https://github.com/yurydelendik/sassc/tree/wasi 

Please provide any feedback and TODO items to make the patch landable. Thanks.


xzyfer:
Thank you for the effort. I was talking to some wasi folks in Berlin
recently about this. Happy to see it progressing.

src/file.cpp
```diff
@@ -57,10 +67,10 @@ namespace Sass {
     // always with trailing slash
     std::string get_cwd()
     {
-      const size_t wd_len = 4096;
       #ifdef _WASM
-        std::string cwd = "/";
+        std::string cwd = get_cwd_from_env();
```
mgreter: Could you elaborate why `getcwd` does not work for wasm here?

src/file.cpp
```diff
@@ -49,6 +50,15 @@ inline static std::string wstring_to_string(const std::wstring &wstr)
 # endif
 #endif
 
+#ifdef _WASM
+inline static std::string get_cwd_from_env()
+{
+  char* value = getenv("PWD");
+  if (!value) return "(unreachable)";
```
mgreter: This seems to obfuscate the error. Wouldn't it be better to
throw an error or return at least something usable, e.g. "./"?

Makefile
```diff
@@ -198,15 +217,18 @@ debug-shared: shared
 lib:
 	$(MKDIR) lib
 
-lib/libsass.a: $(COBJECTS) $(OBJECTS) | lib
-	$(AR) rcvs $@ $(COBJECTS) $(OBJECTS)
+lib/libsass.a: $(COBJECTS) $(OBJECTS) $(WASMOBJECTS) | lib
+	$(AR) rcvs $@ $(COBJECTS) $(OBJECTS) $(WASMOBJECTS)
```
mgreter: Not sure if this adds overhead in case we compile regular `libsass.a` (without wasm)?

src/plugins.cpp
```diff
@@ -62,6 +64,7 @@ namespace Sass {
     typedef Sass_Function_List (*__plugin_load_fns__)(void);
     typedef Sass_Importer_List (*__plugin_load_imps__)(void);
 
+#ifndef _WASM
```
mgreter: Would make sense to disable all plugin related stuff on wasm.
E.g. all plugin related C-API calls and so on. If so I would prefer
to have it behind a separate define, e.g. `ENABLE_PLUGINS`.

src/file.cpp
yurydelendik: I took `(unreachable)` from [getcwd() man](http://man7.org/linux/man-pages/man3/getcwd.3.html). The libsass doesn't not like "./" as CWD path, so I opted to something "visible".

Shall I use "/" ?

src/file.cpp
yurydelendik: wasi's libc does not implement getcwd yet

src/file.cpp
mgreter: I see that all other "cwd" code-paths just throw:
`Exception::OperationError("cwd gone missing")`
Maybe that would be safer since AFAIK libsass
would not know how to handle this anyway.

yurydelendik: In most of use cases cwd will not be present for the wasm target. Making it throw will be a barrier in usage of the library when user fails to provide an environment variable since libsass initializes the `CWD`. In most of the wasm cases absolute files paths can be provided, making the CWD non-used.

mgreter: I see, in that case I don't know what is actually the best solution.
It might make sense to default it to "/" then, since if I understood
you right, wasm doesn't really have a concept of cwd ...

mgreter: So if I understood it right, wasm doesn't have a concept of cwd (yet?).
Therefore users can define it by defining the PWD environment variable,
and in case this env variable is absent it defaults to the root directory!?

Makefile
yurydelendik: The ifdef at line 191 makes "wasm/libcxxabi_stubs.o" only available for the WASM. We need to place the libcxxabi_stubs.o in the wasm's lib/libsass.a since libc++abi is missing exception handling stuff (you can see it as a hack -- it is not possible to compile libsass without `-fno-exceptions`, which is right way to fix that for wasm atm)

src/file.cpp
yurydelendik: That's correct.

Though I think the chances that PWD will be specified is minimal, and not fully work due to the logic in the sassc connected with the CWD (non)usage with output file name.

mgreter: Let's see how this plays out. I guess this might be an area which also needs
clarification on the wasm side. Maybe libsass is a good example for this, as
see use cases either on the web and on the command line. I guess wasm
needs to deal with this at some point but might not be there yet :)

Makefile
mgreter: I see, I hope the expansion `$(WASMOBJECTS)` when `WASMOBJECTS` is undefined in OK then in the other case.

wasm/cli.js
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node --experimental-wasm-bigint
```
nschonni: This file seems like it should be in a different repo

yurydelendik: Absolutely. Some of the files (e.g. wasm/cli.js, package.json, README.md) just give an example how sassc.wasm can be run. I'll remove them before flipping this PR status to "Ready for review"

nschonni: Sounds good, wasn't sure if it was part of the scaffolding or not

yurydelendik:
> There is also the autotools build. Any plans to
add wasm target also there? 

I have near 0 experience with autotools. I had no plans to touch that portion. Let me know what needs to be done there.

The example files were removed from this repo and placed at https://github.com/yurydelendik/sassc/tree/wasi/wasm

src/file.cpp
```diff
@@ -49,6 +50,15 @@ inline static std::string wstring_to_string(const std::wstring &wstr)
 # endif
 #endif
 
+#ifdef _WASM
+inline static std::string get_cwd_from_env()
+{
+  char* value = getenv("PWD");
```
devsnek: it's common to have `.` as a preopen directory, you might want to check for it first.

yurydelendik: libsass logic changes/normalize paths -- it might be an issue to preserve .

xzyfer:
I suggest rebasing on master because a large changeset was just merged

src/file.cpp
devsnek: its ok for it to be normalized to a full path. think it would be reasonable to try using `realpath(".")` as the first check.

yurydelendik: Added option to avoid removal "./" for the `_WASM` in the `make_canonical_path()` -- "./" shall work now.

xzyfer:
I'm on vacation until next week. I'll take a look when I'm back.

yurydelendik:
Is there anything needs to be done for acceptance of this PR?

Issue #2419, opened 2017-06-16
Drop support for extending compound selectors


xzyfer:
Issue https://github.com/sass/sass/issues/1599
PR https://github.com/sass/sass/pull/2319

Issue #2418, opened 2017-06-16
Add deprecation messages for compound extends


xzyfer:
Issue sass/sass#1599
PR https://github.com/sass/sass/pull/2300
See sass/sass-spec#1134

mgreter:
Should be addressed by sass/libsass#2908

mgreter:
Somehow these specs have gone missing?

nex3:
The warning tests were removed when Ruby-Sass-only tests were removed, since at that point only Ruby Sass was emitting those warnings (Dart Sass instead just rejects compound extensions).

xzyfer:
Seems like we'll need to revive those specs since there is still an open
issue to deprecate this behaviour in LibSass.

mgreter:
The warning has been added to 3.6.2 and wording was fixed with 3.6.3.

PR #2908, opened 2019-06-14
Refactoring / back-porting selectors and extend

mgreter:
Needs sass/sass-spec#1396

Continued from sass/libsass#2904.

I know this is a big commit, and it took quite some effort to get this ready
for review. Positive thing is that it removes more lines than it adds 😄 

Feel free to poke at the changes. I know there are still a lot of loose ends.
Some stuff is just there to really make it work. I tried hard to clean this up
to at least have a better state than before. But maybe don't be too picky.
There are over 7000 lines of code. Easy to nitpick, but I would like to get
this merged at some point, so please focus on the big picture. As always
we can adjust and polish spots in the future.

### Main features

- Ports extend logic back from dart-sass
- Gets rid of old extend and node classes
- Ports superselector logic back from dart-sass
- Overhauls internal selector storage structure
- Refactors media rules to match dart-sass closer
- Simplifies compare operations for selectors a bit
- Adds initial implementation for min/max specificity
- Improves selector schema parsing and evaluation
- Implements custom insertion ordered hash map
- Cleans header includes for files touched
- Tons of other small fixes and improvements

### New selector structure (equal to dart sass)

- A `SelectorList` is an array of `ComplexSelector`
- A `ComplexSelector` is an array of `SelectorComponent`
  - Between each `SelectorComponent` is a `space`
- A `SelectorComponent` is a base class for either
  - A `SelectorCombinator` is either `>`, `+` or `~`
  - A `CompoundSelector` in an array of `Simple_Selector`
    - Need to follow some rules (e.g. only one `ID_Selector`)
    - Are joined without any space in between.

I removed `Parent_Selector`, but there is still the `Parent_Refernce`. The reference
is only used in interpolations. Actual selector now get rid of all the fake parent. The
logic is that a parent selector can only occur at the begining of a `CompoundSelector`.
Therefore this class has a flag `hasRealParent` that indicates that we parsed a real parent!
reference on the input. Implicit connections (previously fake parents) are now indicated
by the `chroots` flag on the `ComplexSelector`. If a selector is chrooted, it should not
connect implicitly to its parent (`hasRealParent` might still play a role here).

### Known spots for improvement

- The min/max specificity for selectors has only be partially implements yet.
- Had to add an `originalSelectorStack` in order to support parent resolving
  like dart sass does. Dart sass stores both items in an object, which we lack.
- Parser for interpolated media queries is now overly complex. We have two
  classes only to store this interpolation. Find or create/adjust a parser that
  parses exactly the schema we need for this interpolation. For now it works.
- It should be questioned in general if schemas/interpolation should be stored
  by the target node (e.g. `SelectorList` vs `Ruleset`). It IMO makes more sense
  to store the `Selector_Schema` for a `Ruleset` on hat `Ruleset`. Once it is
  evaled the selector object can be set. This should solve the disambiguate from
  storing a `Selector_Schema` as a base `Selector`. 

### Known spots to develop further

- Added new `TO_CSS` output target. Currently it takes care to throw an
  error if something css incompatible is being printed (e.g. invalid units).
  ToDo: we might should move more error into this code path (TBD)
- Added new `CssMediaRule` class. On extend we directly work with
  media-queries and need them to be evaluated and parsed beforehand.
  Dart-sass seems to have certain ast-nodes only for static css syntax. This
  is not only useful when parsing pure css, once nodes are evaluated, their
  result should match that of any static css. If we ever want to support
  a pure css parser, we probably need to adopt this route anyway!


mgreter:
Had to throw out some facy template functions for dart sass "compatibility".
Restored the more generic ones, they also work for our use cases.

xzyfer:
By porting over dart sass' extend behaviour does this mean loose extending of compound selectors? I believe this feature was deprecated and maybe even removes from dart sass' entirely. This is important to know given its a breaking change and we haven't had a deprecation period yet.

xzyfer:
See #2418 and #2419

mgreter:
Thanks, good point. It seems that spec tests for this have gone missing?
- https://github.com/sass/sass-spec/tree/3.4/spec/extend-tests/092_test_long_extendee
- https://github.com/sass/sass-spec/tree/3.4/spec/libsass-closed-issues/issue_1654/nested
Should be simple enough to get the old behavior back though. Will try to come up with a fix.

src/ast.hpp
```diff
@@ -110,6 +86,20 @@ namespace Sass {
     void update_pstate(const ParserState& pstate);
     Offset off() { return pstate(); }
     Position pos() { return pstate(); }
+
+    // Some obects are not meant to be compared
+    // ToDo: maybe fallback to pointer comparison?
+    virtual bool operator== (const AST_Node& rhs) const {
```
glebm: How about leaving it unimplemented instead, so that the error is compile time instead of runtime?

```c++
virtual bool operator== (const AST_Node& rhs) const = 0;
```

src/ast.hpp
```diff
@@ -228,37 +218,116 @@ namespace Sass {
...
     void clear()            { return elements_.clear(); }
     T last() const          { return elements_.back(); }
     T first() const         { return elements_.front(); }
+
+    bool operator== (const Vectorized<T,U>& rhs) const {
+      // Abort early if sizes do not match
+      if (length() != rhs.length()) return false;
+      // Otherwise test each ast node in order
+      return std::equal(begin(), end(), rhs.begin(),
+        ObjEqualityFn<CssMediaQuery_Obj>);
```
glebm: Why `CssMediaQuery_Obj`? Shouldn't it be `T`?

src/ast.hpp
```diff
@@ -218,7 +208,7 @@ namespace Sass {
   // "Template Method" design pattern to allow subclasses to adjust their flags
   // when certain objects are pushed.
   /////////////////////////////////////////////////////////////////////////////
-  template <typename T>
+  template <typename T, typename U>
```
glebm: This `U` for chaining seems like unnecessary complexity for little/no gain. Let's not do this, the `Vectorized` class is already quite convoluted.

src/ast.hpp
```diff
@@ -228,37 +218,116 @@ namespace Sass {
   public:
     Vectorized(size_t s = 0) : hash_(0)
     { elements_.reserve(s); }
+    Vectorized(const std::vector<T> vec) : hash_(0)
```
glebm: Since you're copying it here, there is no need for `const`, and you can move the vector in to avoid double copy:

```c++
Vectorized(std::vector<T> vec) : hash_(0), elements_(std::move(vec)) {}
```

src/ast_helpers.hpp
```diff
@@ -0,0 +1,348 @@
...
+  struct PtrHash {
+    template <class T>
+    size_t operator() (const T* ptr) const {
+      return PtrHashFn(ptr);
+    }
+  };
+
+  template <class T>
+  bool PtrEqualityFn(const T* lhs, const T* rhs) {
+    if (lhs == nullptr) return rhs == nullptr;
```
glebm: These 2 checks are redundant, please remove.

src/ast_helpers.hpp
```diff
@@ -0,0 +1,348 @@
...
+  void listEraseItemIf(T& vec, UnaryPredicate* predicate)
+  {
+    vec.erase(std::remove_if(vec.begin(), vec.end(), predicate), vec.end());
+  }
+
+  // Check that every item in `lhs` is also in `rhs`
+  // Note: this works by comparing the raw pointers
+  template <typename T>
+  bool listIsSubsetOrEqual(const T& lhs, const T& rhs) {
+    for (auto const& item : lhs) {
```
glebm: `const auto& item`

src/ast_helpers.hpp
```diff
@@ -0,0 +1,348 @@
...
+
+  // ##########################################################################
+  // Special case insensitive string matcher. We can optimize
+  // the more general compare case quite a bit by requiring
+  // consumers to obey some rules (lowercase and no space).
+  // - `literal` must only contain lower case ascii characters
+  // there is one edge case where this could give false positives
+  // test could contain a (non-ascii) char exactly 32 below literal
+  // ##########################################################################
+  inline bool equalsLiteral(const std::string& literal, const std::string test) {
```
glebm: The second argument should be passed by reference.
The first argument should probably be a `const char *`, based on the usage.

Consider a more descriptive name, such as `equalsCaseInsensitive`.

src/ast_helpers.hpp
```diff
@@ -0,0 +1,348 @@
...
+    return equalsLiteral("nth-child", test)
+      || equalsLiteral("nth-last-child", test);
+  }
+  // isSelectorPseudoBinominal
+
+  // ###########################################################################
+  // Returns [name] without a vendor prefix.
+  // If [name] has no vendor prefix, it's returned as-is.
+  // ###########################################################################
+  inline std::string unvendor(const std::string& name)
```
glebm: Consider moving these to `util_string.cpp` and adding unit tests.

glebm:
> Easy to nitpick, but I would like to get this merged at some point, so please focus on the big picture. As always we can adjust and polish spots in the future.

Please, let's properly clean up before submitting: it is easier to clean up at review time than later.

src/ast.hpp
mgreter: Because then I need to implement it on every class that extends from this.

src/ast.hpp
mgreter: Correct, thanks, will fix that

src/ast.hpp
mgreter: I propose to do this later. The whole Vectorized and Hashed needs overhauling.
And I don't think it adds that much of complexity ...

src/ast.hpp
mgreter: Is a move safe here? IMO we must make copies in this constructor?

src/dart_helpers.hpp
```diff
@@ -0,0 +1,211 @@
...
+  // ##########################################################################
+  // Flatten `vector<vector<T>>` to `vector<T>`
+  // ##########################################################################
+  template <class T>
+  std::vector<T> flatten(std::vector<std::vector<T>>& all)
+  {
+    std::vector<T> flat;
+    for (std::vector<T>& sub : all) {
+      std::move(std::begin(sub), std::end(sub),
+        std::inserter(flat, std::end(flat)));
```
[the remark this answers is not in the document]
mgreter: Any reason why this is better or just personal preference?

mgreter:
> I'm not a big fan of `ObjPtrEquality*` and `PtrObjEquality*`. They are very similar and confusing.
> As I mentioned before, I don't think `(Obj)?Ptr(Equality|Hash)(Fn)?` functions and structs should exist: they are the default comparison / hash functions, and having them is confusing because the reader thinks they are custom / non-default ones. It is also confusing because we now have very similarly named `PtrObj` functions / structs.

The problem is we don't always want the same behavior on every hash. Sometimes I want the hash to have unique pointers and sometimes I want the hash to have unique objects. To do this I need to pass the functor structs to the hash declaration. Also we mostly use SharedImpl<T> as keys, so for those the default implementation is even less obvious. Therefore I needed these explicit functions and functors.

On the naming I could agree a bit more. But I believe it is not too hard to get it right. `ObjPtr` means that you have an object (SharedImpl) and want to hash by the shared pointer (so two SharedImpl pointing to the same object are considered equal). `PtrObj` means you have a raw pointer, but want to compare the actual object behind it (so two SharedImpl with different objects can sill be equal if the two objects are themselves equal). `Obj` and `Ptr` should be obvious them, they either compare the object or the pointer.

src/ast.hpp
glebm: Yes, a move is perfectly safe here, because the `vec` vector is passed by value.

src/ast.hpp
glebm: > I propose to do this later.

Why do it here, if it's not even being used (as far as I can tell by searching through the diffs)?

src/ast.hpp
glebm: That seems to indicate that this class shouldn't have this method at all.
OK with doing this later though.

src/ast_sel_super.cpp
```diff
@@ -0,0 +1,537 @@
...
+  // Returns whether [simple1] is a superselector of [simple2].
+  // That is, whether [simple1] matches every element that
+  // [simple2] matches, as well as possibly additional elements.
+  // ##########################################################################
+  bool simpleIsSuperselector(
+    const SimpleSelectorObj& simple1,
+    const SimpleSelectorObj& simple2)
+  {
+    // If they are equal they are superselectors
+    if (ObjEqualityFn(simple1, simple2)) {
```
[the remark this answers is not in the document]
mgreter: But it lacks the `nullptr` check, so possible to crash ...

glebm:
> Also we mostly use SharedImpl as keys, so for those the default implementation is even less obvious.

It is idiomatic for smart pointers (like `SharedImpl`) to behave just like regular pointers. This means `==` compares the pointers (as we currently do), not the pointed values. This is what `std::shared_ptr` and `std::unique_ptr` do as well. This is why I don't consider the default equality/hash confusing at all.

> On the naming I could agree a bit more.

I have an idea about this:

* `Pointer(Equality|Hash)`
* `PointedValue(Equality|Hash)`

There is no need for `Obj`-specific versions, as these functions are templated and work on anything that compares with `nullptr` and implements `->` (including our smart pointer type).

src/dart_helpers.hpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return flat;
+  }
+
+  // ##########################################################################
+  // Expands each element of this Iterable into zero or more elements.
+  // Calls a function on every element and ads all results to flat array
+  // ##########################################################################
+  template <class T>
+  std::vector<T> expand(std::vector<std::vector<T>> vec) {
```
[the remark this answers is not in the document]
mgreter: I Superfluous from the revert I had to made due to issues with my more generalized versions I had in MSVC.

src/dart_helpers.hpp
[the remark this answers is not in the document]
glebm: Performance + more idiomatic:

https://stackoverflow.com/questions/26307577/whats-the-benefit-of-stdback-inserter-over-stdinserter

mgreter:
I need to compare objects in hashes here: https://github.com/sass/libsass/pull/2908/files#diff-61f3048af5edbbaf974fb08271245c6eR32 ... also see the first one, that compares the raw pointers.

src/ast_sel_super.cpp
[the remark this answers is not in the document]
glebm: Ah. SharedImpl should implement comparison operators like all smart pointers do. I'll send a PR.

src/dart_helpers.hpp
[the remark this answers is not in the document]
mgreter: OK, looks valid, thanks :)

src/ast_sel_unify.cpp
```diff
@@ -1,280 +1,276 @@
...
+  // ##########################################################################
+  CompoundSelector* CompoundSelector::unifyWith(CompoundSelector* rhs)
+  {
     if (empty()) return rhs;
-    Compound_Selector_Obj unified = SASS_MEMORY_COPY(rhs);
-    for (const Simple_Selector_Obj& sel : elements()) {
-      unified = sel->unify_with(unified);
+    CompoundSelectorObj unified = SASS_MEMORY_COPY(rhs);
+    for (const SimpleSelectorObj& sel : elements()) {
+      CompoundSelectorObj cp = unified;
```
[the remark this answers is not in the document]
mgreter: Thanks, will remove

src/ast.hpp
mgreter: I know, but before it was not really better, it actually just ignored those AFAICT. So I consider it at least an improvement.

src/ast.hpp
mgreter: https://github.com/sass/libsass/pull/2908/files/5ed7e8429c1138de625f3fa40a340fcabedaca2a#diff-dd77ff8326ef81051bd3b8cc5cc33669

src/ast.hpp
mgreter: Got it, made the prototype to accept a reference then doing the copy on the initializer.

src/ast_helpers.hpp
mgreter: Thanks, will fix

src/ast_helpers.hpp
mgreter: Thanks, will remove

src/ast_helpers.hpp
mgreter: I would like to keep the name less general, since one should know that it is in fact not a real `equalsCaseInsensitive` method. I copied and adopted it from code we have in the parser. It does the job well enough, but only for certain input. I will change the prototype as you suggested though, I believe that it does make it more clear that this function is a little special.

src/ast_sel_unify.cpp
```diff
@@ -1,280 +1,276 @@
...
+      else {
+        return {};
+      }
+    }
+
+    std::vector<std::vector<SelectorComponentObj>> complexesWithoutBases;
+    for (size_t i = 0; i < complexes.size(); i += 1) {
+      std::vector<SelectorComponentObj> sel = complexes.at(i);
+      sel.pop_back(); // remove last item (base) from the list
+      complexesWithoutBases.push_back(sel);
```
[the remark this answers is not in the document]
mgreter: Thanks, will update

src/ast_sel_super.cpp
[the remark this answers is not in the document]
glebm: https://github.com/sass/libsass/pull/2909

src/ast.hpp
```diff
@@ -218,7 +218,7 @@ namespace Sass {
   public:
     Vectorized(size_t s = 0) : hash_(0)
     { elements_.reserve(s); }
-    Vectorized(const std::vector<T> vec) :
+    Vectorized(const std::vector<T>& vec) :
```
glebm: Passing by value is better if you're going to copy anyway, because that allows for a copy-free constructor if the argument at call site is an rvalue. For example:

```c++
Vectorized a({1, 2, 3}); // No copies made, only 2 moves.
```

src/ast_sel_super.cpp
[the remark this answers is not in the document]
glebm: Nevermind, didn't understand what you meant but now I do.

src/dart_helpers.hpp
```diff
@@ -0,0 +1,211 @@
...
+#include <iterator>
+#include <functional>
+
+namespace Sass {
+
+  // ##########################################################################
+  // Flatten `vector<vector<T>>` to `vector<T>`
+  // ##########################################################################
+  template <class T>
+  std::vector<T> flatten(std::vector<std::vector<T>>& all)
```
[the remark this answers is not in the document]
mgreter: I've adjusted a few functions in helper. I hope everything is addressed now.
I guess there are quite a few edges for performance tuning in the extender path ...

src/ast.hpp
mgreter: OK, thanks for the heads up. Possibly that we have
dozens of other places where this was not known.

src/extender.cpp
```diff
@@ -0,0 +1,1088 @@
...
+          pseudos.push_back(pseudo->withSelector(
+            expanded[i]->wrapInList()
+          ));
+        }
+        return pseudos;
+      }
+    }
+
+    SelectorListObj list = SASS_MEMORY_NEW(SelectorList, "[phony]");
+    return { pseudo->withSelector(list->concat(complexes)) };
```
glebm: Can we avoid `concat` here like this?

```c++
SelectorListObj list = SASS_MEMORY_NEW(SelectorList, "[phony]", complexes));
return { pseudo->withSelector(list); }
```

src/ast.hpp
```diff
@@ -218,8 +218,8 @@ namespace Sass {
   public:
     Vectorized(size_t s = 0) : hash_(0)
     { elements_.reserve(s); }
-    Vectorized(const std::vector<T>& vec) :
-      elements_(vec),
+    Vectorized(const std::vector<T> vec) :
```
glebm: Copy-elision won't work with `const`.
There is little reason to mark arguments passed by value as `const` anyway.

mgreter: OK, adjusted a third time

src/ast_sel_super.cpp
```diff
@@ -0,0 +1,537 @@
...
+  // dart: stderr.writeln("result " + compound.toString());
+  // ##########################################################################
+
+  // ##########################################################################
+  // Returns whether [list1] is a superselector of [list2].
+  // That is, whether [list1] matches every element that
+  // [list2] matches, as well as possibly additional elements.
+  // ##########################################################################
+  bool listIsSuperslector(
+    std::vector<ComplexSelectorObj> list1,
```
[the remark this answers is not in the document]
mgreter: Done

src/extender.cpp
mgreter: That constructor does not exist. Since I'm tired of debating it I removed the Vectorized method chaining experiment.

src/ast.hpp
mgreter: Since I'm tired of debating it, I removed the Vector method chaining experiment.

src/ast_helpers.hpp
mgreter: Done

mgreter:
Polished up the extender code once more ...

mgreter:
Something tripped up MSVC debug build, need to go back to last success and go from there ...

src/emitter.cpp
```diff
@@ -112,7 +111,7 @@ namespace Sass {
     wbuf.buffer = text + wbuf.buffer;
   }
 
-  char Emitter::last_char()
+  const char& Emitter::last_char()
```
glebm: nit: `char` is tiny, it's better to return by value.

src/ast_selectors.cpp
```diff
@@ -363,1113 +312,719 @@ namespace Sass {
...
 
-    // first resolve_parent_refs the tail (which may return an expanded list)
-    Selector_List_Obj tails = tail ? tail->resolve_parent_refs(pstack, traces, implicit_parent) : 0;
+    std::string ourType(this->type());
+    std::string theirType(other->type());
+    std::string ourModifier(this->modifier());
+    std::string theirModifier(other->modifier());
 
-    if (head && head->length() > 0) {
+    std::transform(ourType.begin(), ourType.end(), ourType.begin(), ::tolower);
```
glebm: We shouldn't use `tolower`, because it's C-locale dependent. I'll send a PR shortly adding our own method.

src/ast.cpp
```diff
@@ -80,19 +71,21 @@ namespace Sass {
     return i.get_buffer();
   }
 
-  const std::string AST_Node::to_string() const
+  const std::string AST_Node::to_css(Sass_Inspect_Options opt) const
```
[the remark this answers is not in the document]
mgreter: Done

src/ast.cpp
```diff
@@ -137,6 +130,14 @@ namespace Sass {
     is_root_(ptr->is_root_)
   { }
 
+  bool Block::isInvisible() const
+  {
+    for (auto item : elements()) {
```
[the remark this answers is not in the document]
mgreter: Done

src/ast.hpp
```diff
@@ -228,37 +218,115 @@ namespace Sass {
...
+      hash_(0)
+    {}
     virtual ~Vectorized() = 0;
     size_t length() const   { return elements_.size(); }
     bool empty() const      { return elements_.empty(); }
     void clear()            { return elements_.clear(); }
     T last() const          { return elements_.back(); }
     T first() const         { return elements_.front(); }
+    T& last()               { return elements_.back(); }
+    T& first()              { return elements_.front(); }
```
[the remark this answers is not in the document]
mgreter: Done

src/ast.hpp
```diff
@@ -228,37 +218,115 @@ namespace Sass {
...
+        // A std::find checks for pointer equality
+        if (ObjEqualityFn(el, rhs)) {
+          return true;
+        }
+      }
+      return false;
+    }
+
+    // This might be better implemented as `operator=`?
+    std::vector<T>& elements(std::vector<T>& e) {
```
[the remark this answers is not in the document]
mgreter: Done

src/ast.hpp
```diff
@@ -228,37 +218,115 @@ namespace Sass {
...
     }
-    Vectorized& unshift(T element)
+
+    // Insert one item on the back
+    // ToDo: rename this to push
+    void append(T element)
     {
-      elements_.insert(elements_.begin(), element);
-      return *this;
+      if (element) {
```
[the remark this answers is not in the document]
mgreter: Remove the if

src/ast.hpp
```diff
@@ -806,9 +882,101 @@ namespace Sass {
...
+    ADD_PROPERTY(std::string, modifier);
+
+    // The media type, for example "screen" or "print".
+    // This may be `null`. If so, [features] will not be empty.
+    ADD_PROPERTY(std::string, type);
+
+    // Feature queries, including parentheses.
+    ADD_PROPERTY(std::vector<std::string>, features);
+
+  public:
```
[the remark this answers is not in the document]
mgreter: Done

src/emitter.cpp
mgreter: Done

src/ast.hpp
```diff
@@ -228,37 +218,115 @@ namespace Sass {
...
+
+    // Insert all items from compatible vector
+    void concat(const std::vector<T>& v)
+    {
+      if (!v.empty()) reset_hash();
+      elements().insert(end(), v.begin(), v.end());
+    }
+
+    // Syntatic sugar for pointers
+    void concat(Vectorized<T>* v)
```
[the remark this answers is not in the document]
mgreter: DOne

src/ast_selectors.cpp
mgreter: Can we fix this after this PR has landed?

xzyfer:
Given the magnitude of this chat I'm going to tag a release on the current master

xzyfer:
Done https://github.com/sass/libsass/releases/tag/3.6.1

src/ast_selectors.cpp
glebm: Absolutely

mgreter:
🚀 Final rebase to master. Thanks for the review @glebm.

mgreter:
Please don't merge yet, want to see if the clang warning was the culprit of earlier MSVC issues.

.travis.yml
```diff
@@ -1,70 +0,0 @@
-language: cpp
```
glebm: Did you mean to delete this?

appveyor.yml
```diff
@@ -5,19 +5,12 @@ environment:
   ruby_version: 24-x64
   TargetPath: sassc/bin/sassc.exe
   matrix:
-  - Compiler: msvc
```
glebm: Why have these been deleted?

.travis.yml
mgreter: See comment above, trying something in Appveyor CI

appveyor.yml
mgreter: See comment above, trying something in Appveyor CI

src/ast.hpp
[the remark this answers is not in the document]
glebm: I'm still seeing `const T last()`. Merge/rebase accident?

glebm:
BTW, if AppVeyor gets stuck, it's usually because the program crashes (in debug build, that triggers a debug dialog, which confuses AppVeyor).

The best way I found to resolve these is to run in actual Visual Studio, as that shows the error. I wonder if there is a way to print these errors to the log in AppVeyor instead of hanging.

mgreter:
OK, seems the CI "issue" with MSVC2013 in debug mode still exists.

glebm:
It's usually a real bug as well, not just a CI issue (MSVC standard library in debug mode has lots of useful checks).

mgreter:
I know it either is showing a popup for some assertion, but I can't reproduce with my VS community edition, and I don't really want to install a 2013 version ...

mgreter:
I mean it is not a real problem since we can avoid triggering it. It would still be intersting
to know why this happens, but I'm just not that curious to go down that rabbit hole.

glebm:
You can access the hanging build via Remote Desktop!
https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

mgreter:
Hmm, ok, that sounds interesting ... I give it a whirl ...

mgreter:
OK, got some results, but I don't see why this occurs (VS2013 bug?)

image

glebm:
Trying to reproduce locally in VS2013. What error message do you get and on which test?

glebm:
Reproduced right away:

Capture

glebm:
Reproduces with the following input:

```scss
.a x {a: b}
.a.b + y {@extend x}
```

(this is from sass-spec/spec/extend-tests/174_test_combinator_unification_plus_space, one of the many crashing tests)

glebm:
This happens because here:

```c++
    ExtByExtMap::iterator existingExtensions = extensionsByExtender.find(target);
    const std::vector<Extension>& extensions = existingExtensions->second;
```

Find returns the `end` iterator, and the code attempts to dereference it.

src/extender.cpp
```diff
@@ -350,10 +350,9 @@ namespace Sass {
     newExtensionsByTarget.insert(std::make_pair(target, newExtensions));
     if (hasExistingExtensions) {
       ExtByExtMap::iterator existingExtensions = extensionsByExtender.find(target);
-      const std::vector<Extension>& extensions = existingExtensions->second;
-      if (!extensions.empty()) {
+      if (!existingExtensions->second.empty()) {
```
glebm: The bug is here.

`existingExtensions` can be the `end` iterator, in which case it cannot be dereferenced (undefined behaviour).

mgreter:
That find should not fail since it is in an `if(hasExistingExtensions)`, but can try to add an additional check ..

src/extender.cpp
glebm: Adding a check for `existingExtensions != extensionsByExtender.end()` makes the test suite run without crashes for me.

mgreter: That find should not fail since it is in an if(hasExistingExtensions), but can try to add an additional check ..

glebm:
> That find should not fail since it is in an if(hasExistingExtensions), but can try to add an additional check ..

It doesn't seem to be in any if like that. The check is on the line *after* `existingExtensions->second` (:boom:).

glebm:
Adding a check removes the crashes for me.

mgreter:
Hmm, maybe the key `target` has changed from the check on line 302 ... let's see if CI agrees.
I was able to to move the failing case from MSVC to clang on MacOSX 🙈

glebm:
> Hmm, maybe the key target has changed from the check on line 302 ... let's see if CI agrees.

There is just an assignment  to the bool there, but it's not actually checked later.

The only use of it is:

```c++
      if (hasRule || hasExistingExtensions) {
        newExtensions.insert(complex, state);
      }
```

And then the only early return is:

```c++
    if (newExtensions.empty()) {
      return;
    }
```

So you could get there if `hasExistingExtensions` is false and `hasRule` is true.

mgreter:
Ah, I added that if in `CI challenge 2`, but still have issues now on travis mac build.
Will retrack once again and try to fix this issue first.

glebm:
I wrote up some profiling instructions a while ago: https://github.com/sass/libsass/blob/master/docs/dev-profiling.md

Hopefully this can help with debugging the test that times out on macOS.

mgreter:
It's not timing out, it's actually failing flat out: https://travis-ci.org/sass/libsass/jobs/546511787

glebm:
From that link:

> Execution timed out after 90s

src/extender.cpp
```diff
@@ -420,7 +420,7 @@ namespace Sass {
 
     // for (const Extension extension : oldExtensions) {
     for (size_t i = 0; i < oldExtensions.size(); i += 1) {
```
glebm: Perhaps this should just be:

```c++
for (const Extension& extension : oldExtension) {  
```

or `Extension extension : ...` if it needs to be copied (doesn't look like it needs to though)

mgreter: That's exactly what I tried to achieve two lines below ...

glebm:
Latest error:

Capture

mgreter:
Maybe if you have time feel free to debug. I'm going to bed now. I think the latest commit is ok, but changing the other function prototypes keep triggering MSVC debug build. Will squash continue tomorrow.

glebm:
Figured out where `extension.target` starts pointing to invalid memory:

```
extensionsByExtender[simple].push_back(withExtender);
```

Here once enough elements are pushed, `extensionsByExtender[simple]` reallocates.

Since this invalidates the pointers (`extension.extender` and `extension.target`), I imagine somewhere in the code these are created from pointers to vector elements.

This is unsafe because pointer addresses may change when the vector resizes its buffer.

mgreter:
I feel that it might have something to do with `for (auto x : ` loops, since IMO those use iterators that could get invalid during the whole loop (as opposite to regular `for(size_t i` loops.

glebm:
> I feel that it might have something to do with for (auto x : loops, since IMO those use iterators that could get invalid during the whole loop (as opposite to regular for(size_t i loops.

Definitely not iterators. This is caused by pointers (SharedImpl) pointing to elements in a vector storage. When the backing vector's buffer is resized, these SharedImpls' become dangling pointers. Evidenced by stepping through with a debugger to the crash (more obvious with my patch above).

mgreter:
IMO the for range loops do iterate from begin to end, so if in the meantime, the underlying container is relocated, it makes kinda sense that the iteration in progress is about to fail. I could be wrong but my tests (I did now install MSVC 2013) seem to indicate that this is the case. Not saying the other thing you mention is also an issue.

mgreter:
OK, got it to pass both tests. Will squash now to see what I got to clean up ...

mgreter:
So I finally seem to have found the culprit, took way longer than it should have 😓 

It basically boils down to:
```diff
- for (Extension extension : oldExtensions) {
- for (size_t i = 0; i < oldExtensions.size(); i += 1) {
+ for (size_t i = 0, iL = oldExtensions.size(); i < iL; i += 1) {
```

As expected we seem to append to the underlying vector somewhere. As I wrote earlier, this does invalidate the for range based loop, but also made clang on mac read the newly added items, which is not how the original dart sass behavior is (probably because it does in fact work on a copy). That's why once MSVC timed out and on another occasion clang on mac os did. Will finish up now ...

mgreter:
Just to gain back my sanity, I made a test:
```c++
    for (size_t i = 0, iL = oldExtensions.size(); i < iL; i += 1) {
      if (iL != oldExtensions.size()) {
        std::cerr << "Gotcha\n";
      }
```
And who would have guessed:
```
#   Failed test 'Warnings: t\sass-spec\spec\extend-tests\215_test_multiple_source_redundancy_elimination\input.scss'
#   at t/99_sass_specs.t line 658.
# +---+----------+----------+
# | Ln|Got       |Expected  |
# +---+----------+----------+
# *  1|'Gotcha'  |''        *
# +---+----------+----------+
```

src/ast.hpp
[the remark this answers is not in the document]
mgreter: Did not correctly apply, will change so

mgreter:
That should be the 🚀 final version. Just waiting for 💚 !

mgreter:
While CI is doing its job, I just wanted to mention the magnitude of this commit. It closes and cuts-off one of the most "alien" code bases we dragged around for so many years, namely [node.cpp](https://github.com/sass/libsass/blob/0bab610a893c5bf63f693bd6c9549e2368c1804f/src/node.cpp) and [extens.cpp](https://github.com/sass/libsass/blob/0bab610a893c5bf63f693bd6c9549e2368c1804f/src/extend.cpp). To bad the [bounty](https://www.reddit.com/r/Frontend/comments/2ibko5/extend_now_works_in_libsass_2000_bounty/) is no longer up for grab. Nearly 8000 lines of code mean about 20% of the whole libsass code base. And thanks to the base work by dart-sass and @nex3 this was possible. The time investment was something around 200h as I initially estimated. 🔥 🎆 🍻

mgreter:
Here we go 🛫

glebm:
Great work! :tada:

xzyfer:
Very impressive effort

PR #2911, opened 2019-06-16
Add and use our own ASCII case-conversion routines

glebm:
`std::tolower` and `std::toupper` are locale-dependent and should not be used anywhere we currently use them.

Adds and switches to our own implementations for these and other locale-dependent functions.

### DIFF
--- a/src/util_string.hpp
+++ b/src/util_string.hpp
@@ -29,9 +29,44 @@ namespace Sass {
     char opening_bracket_for(char closing_bracket);
     char closing_bracket_for(char opening_bracket);
 
-  }
-  // namespace Util
-}  
-// namespace Sass
+    // Locale-independent ASCII character routines.
 
+    inline bool ascii_isalpha(unsigned char c) {
+      return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
+    }
+
+    inline bool ascii_isdigit(unsigned char c) {
+      return (c >= '0' && c <= '9');
+    }
+
+    inline bool ascii_isalnum(unsigned char c) {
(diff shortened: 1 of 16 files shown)


xzyfer:
sass2scss is an external library we vender in. The upstream is mgreter/sass2scss

glebm:
Ah, thanks. Removed the sass2scss commit and opened https://github.com/mgreter/sass2scss/pull/39 instead.

xzyfer:
Note that we don't vendor master atm. There was a regression and we had to
revert to an older commit a while back. Pulling in the lastest upstream may
not be trivial.

src/util_string.hpp
```diff
@@ -6,14 +6,65 @@
...
+std::string rtrim(std::string str);
 
 std::string normalize_newlines(const std::string& str);
 std::string normalize_underscores(const std::string& str);
 std::string normalize_decimals(const std::string& str);
 char opening_bracket_for(char closing_bracket);
 char closing_bracket_for(char opening_bracket);
 
+// ASCII routines based on Abseil:
+// https://github.com/abseil/abseil-cpp/blob/8f11724067248acc330b4d1f12f0c76d03f2cfb1/absl/strings/ascii.h
```
mgreter: Can Apache License 2.0 stuff be included in MIT projects. IMO Apache License has more restrictions?

glebm: I'm not sure. Just to be on the safe side, I've re-implemented these functions myself.

My implementation is not as fast as the Abseil one (no lookup tables) but I don't think these are a bottleneck for us, and we can implement lookup tables ourselves later on.

mgreter: Thanks, better safe than sorry

PR #2929, opened 2019-06-28
fix(feature-exists): Parameter name to "feature"

### LINKED ISSUE #2928: feature-exists()'s parameter name should be $feature
LibSass lists `feature-exists()`'s parameter name as `$name`, but it should be `$feature` to match Dart and Ruby Sass.

nschonni:
To match Ruby/Dart Sass
Closes #2928

### DIFF
--- a/src/fn_miscs.cpp
+++ b/src/fn_miscs.cpp
@@ -76,10 +76,10 @@ namespace Sass {
       }
     }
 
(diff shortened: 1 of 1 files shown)


nschonni:
I just eyeball'd this, so maybe i'm not correct

PR #2930, opened 2019-06-28
chore: Ignore MS IDE caches

nschonni:
Ignore VS and VS Code users files


nschonni:
Just removes some user files for people developing on Windows

PR #2943, opened 2019-07-08
Explicitly pass include path for spec directory

mgreter:
Trying to fix CI


nschonni:
FYI, your branch may be a bit behind, I added another test target in https://github.com/sass/libsass/commit/34490a5dc805166d1172b3bc22ee12663a6ac5ae

PR #2945, opened 2019-07-17
fix: Error on non-int str-insert index

### LINKED ISSUE #2921: str-insert() should require an integer index
`str-insert("", "", 0.5)` should throw an error because `$index` isn't an integer. Instead in LibSass it returns `""`.

nschonni:
See https://github.com/sass/sass-spec/pull/1439
Closes #2921


nschonni:
@mgreter I tried swapping the double to an int, but that didn't work out 😝 
Hope this makes sense

nschonni:
@mgreter should I port over you "assertInt" instead?

mgreter:
Don't mind, the "invertInt" is part of a WIP branch ...

PR #2947, opened 2019-07-17
fix: str-slice can't start at 0

nschonni:
See https://github.com/sass/sass-spec/pull/1441
Closes #2920

PR #2948, opened 2019-07-17
fix: Update "call" parameter to $function

nschonni:
See https://github.com/sass/sass-spec/pull/1442
Closes 2941


nschonni:
@mgreter sorry, I just realized I included the commit from #2947, but I rebased it out

PR #2950, opened 2019-07-18
fix: throw on weitht passed to CSS3 invert()

### LINKED ISSUE #2899: invert() shouldn't allow both a number and a weight
The `invert()` function allows a number to be passed in order to support the plain-CSS `invert()` filter. However, if this is done, it should forbid the weight parameter from being passed as well.

nschonni:
See https://github.com/sass/sass-spec/pull/1444
Closes  #2899

### DIFF
--- a/src/fn_colors.cpp
+++ b/src/fn_colors.cpp
@@ -361,12 +361,16 @@ namespace Sass {
     {
       // CSS3 filter function overload: pass literal through directly
       Number* amount = Cast<Number>(env["$color"]);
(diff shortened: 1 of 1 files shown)


nschonni:
@mgreter there is an edge case that I wasn't sure how to handle. If they manually pass in 100% as the weight it will still pass.
I'm not sure if there is a function that I didn't see for checking whether a parameter is set vs. using the default in the signiture.

PR #2967, opened 2019-08-08
Bad code fragment

MrBenGriffin:
Need to allocate memory for data context..


mgreter:
https://github.com/sass/libsass/pull/2968

mgreter:
I've merged my doc changes, please see if anything else should be added. Thanks!

MrBenGriffin:
That’s brilliant.

On a different note, an earlier version of libsass would send some error and warning messages to “cerr” - I’m hoping that this doesn’t happen any more?  Is there an option to be able to add a custom logging handler? I think we are currently parsing the return string..

mgreter:
We still print to stderr, see https://github.com/sass/libsass/pull/2862

PR #2970, opened 2019-08-08
Use deprecated rather than roll your own.

MrBenGriffin:
This is one of the places (outside of debug and error_handling) where cerr is directly used. As error_handling has a deprecated function, it seems sensible to use it.


MrBenGriffin:
This change is rolled into #2971.  
The errors are failing - as using the proper deprecated method is more informative.
I'm not really sure that having character identical error message tests is useful.

PR #2971, opened 2019-08-08
std::clog instead of std::cerr for non-critical.

MrBenGriffin:
std::clog is a std::ostream that represents the standard logging stream, and therefore is more suitable for most of the uses where std::cerr has been used in libsass.

While clog is not tied to another stream, it is synchronized with stderr by default, and it is more efficient, being buffered.

### DIFF
--- a/src/ast_sel_cmp.cpp
+++ b/src/ast_sel_cmp.cpp
@@ -204,7 +204,7 @@ namespace Sass {
 
   bool CompoundSelector::operator== (const CompoundSelector& rhs) const
   {
-    // std::cerr << "comp vs comp\n";
+    // std::clog << "comp vs comp" << std::endl;
     if (&rhs == this) return true;
     if (rhs.length() != length()) return false;
     std::unordered_set<const SimpleSelector*, PtrObjHash, PtrObjEquality> lhs_set;
--- a/src/ast_sel_super.cpp
+++ b/src/ast_sel_super.cpp
@@ -9,7 +9,7 @@ namespace Sass {
 
   // ##########################################################################
   // To compare/debug against libsass you can use debugger.hpp:
-  // c++: std::cerr << "result " << debug_vec(compound) << "\n";
+  // c++: std::clog << "result " << debug_vec(compound) << std::endl;
   // dart: stderr.writeln("result " + compound.toString());
   // ##########################################################################
 
@@ -450,6 +450,7 @@ namespace Sass {
       }
     }
 
+    // Unreachable.
     return false;
 
   }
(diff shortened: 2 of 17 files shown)


MrBenGriffin:
The CI test failures are demonstrating why this is a good change.
I guess I should fix the errors tests also?

-"DEPRECATION WARNING: Passing a string to call() is deprecated and will be illegal"

+"DEPRECATION WARNING on line 2, column 43 of /sass/spec/libsass-issues/issue_1305/input.scss:
473+Passing a string to call() is deprecated and will be illegal in Sass 4.0."

MrBenGriffin:
I'm not really sure that having character identical error message tests is useful.
I would like advice on this from a maintainer before proceeding further.

xzyfer:
Unfortunately it's the best we have right now. There is some amount of
normalisation that happens behind this scenes but this is the way to go for
now.

MrBenGriffin:
> Unfortunately it's the best we have right now. There is some amount of normalisation that happens behind this scenes but this is the way to go for now.

So, if I revise the failing tests so that they do not fail, would the pull-request be accepted? I don't want to waste time here...

xzyfer:
I'll leave that call to @mgreter since he's heavily refactoring LibSass atm.

nschonni:
Maybe a silly question, but would this affect the C API?

MrBenGriffin:
No question is silly to me, @nschonni.
It would not affect the C API because cerr is a side effect. 
It may offer a better path for possible future implementations of log handlers - but it’s default behaviour is not different from using cerr. Also, with std::endl being used, there’s an implicit flush on the stream anyhow. Part of the pull request involved converting many ‘\n’ to  endl.  The only major effect is that many deprecation errors are now far more detailed than current.

nschonni:
Ah, that would be why the build is failing. You'd need to update the specs in https://github.com/sass/sass-spec as well if the message format is changing

MrBenGriffin:
@nschonni, I found the tests at https://github.com/tmastny/sassr-test - I'm guessing they are folded into https://github.com/sass/sass-spec somehow.

I cannot see any means of being able to manage the errors as they are.  An alternative would be to modify the 'deprecated' method in the error_handling classes. 

I'm sure that there are good reasons to separate all the tests on an issue by issue basis - but I cannot, for the life of me, see any dependency mapping to the root coverage that every single one of those messages are derived from.

I'm not going to be able to continue this work. Sorry.

mgreter:
Going to close this. Upcoming LibSass refactor will include something in this direction :)

PR #2977, opened 2019-08-19
revert concurrent compilation

ahorek:
reverts https://github.com/sass/libsass/pull/2888

it looks like -j 0 spawns unlimited number of processes that causes problems on memory constrained environments


Keats:
It would be nice to have a new release for that change as it can prevent some users from actually building libsass

PR #2984, opened 2019-08-28
math.h is deprecated

ahorek:
there should be no reason to include this header because `cmath` is already included

possible fix for https://github.com/sass/sassc-ruby/issues/149

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -13,7 +13,6 @@
 #include <vector>
 #include <string>
 #include <assert.h>
-#include <math.h>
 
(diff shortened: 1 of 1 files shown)


glebm:
The error message in the sassc-ruby issue actually looks like a missing include of `<cmath>`, strange.

glebm:
Ah, the version of libsass in sassc-ruby is actually missing the `<cmath>` include:

https://github.com/sass/libsass/blob/4d229af5500be1023883c38c4a675f0ed919839d/src/util.hpp#L6-L15

PR #2986, opened 2019-08-30
Assorted typo fixes.

XhmikosR:
There one case I wasn't sure about, so I didn't change it: `binominal`

```
C:\Users\xmr\Desktop\libsass>git grep binominal
src/ast_helpers.hpp:  // Pseudo-element selectors that has binominals
src/parser.cpp:  // Helper to clean binominal string
src/parser.cpp:          std::string parsed(lexed); // always compacting binominals (as dart-sass)
src/parser.hpp:    Expression* parse_binominal();
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ sudo: required
 # this will still catch all coding errors!
 # - AUTOTOOLS=yes COVERAGE=no BUILD=static
 
-# currenty there are various issues when
+# currently there are various issues when
 # built with coverage, clang and autotools
 # - AUTOTOOLS=yes COVERAGE=yes BUILD=shared
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ cache:
   - C:\Ruby%ruby_version%\lib\ruby\gems
   - C:\mingw64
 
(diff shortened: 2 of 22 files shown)


script/tap-driver
```diff
@@ -370,7 +370,7 @@ function setup_result_obj(line)
   sub("^(not )?ok[ \t]*", "", line)
 
   # If the result has an explicit number, get it and strip it; otherwise,
-  # automatically assing the next progresive number to it.
+  # automatically assing the next progressive number to it.
```
nschonni: This might need to be fixed somewhere else first, since it's a GPL'd script that is vendor'd in because of Automake.

XhmikosR: I'll just drop it from this patch :P

XhmikosR:
I dropped that specific change, I didn't notice it's a generated file.

XhmikosR:
How about `binominal`? Isn't that a typo also?

nschonni:
Not sure, but it does look like it

XhmikosR:
I'll leave [these 4 `binominal` instances](https://github.com/sass/libsass/pull/2986#issue-312721643) for someone else to tackle then.

XhmikosR:
Can someone merge this?

PR #2990, opened 2019-09-16
unrecognized psuedos with arguments parse as declaration-values

### LINKED ISSUE #2424: Parse pseudo selector arguments as declaration values
Issue https://github.com/sass/sass/issues/2120
PR https://github.com/sass/sass/pull/2304
Spec sass/sass-spec#1137

Possibly related to #2383

### LINKED ISSUE #2944: libsass throws on non-selector psuedo selector params
## input.scss

This is a hacky syntax that `css-modules` uses to define dependecies across files, in general we'd want sass to ignore it, which is what the other sasses do.

```scss
:import("./foo") {
  color: red;
}
```

## Actual results
libsass 3.5.4

```css
Invalid CSS after ":import(": expected selector, was '"./foo") {' on line 1 at column 1
```

## Expected result

This is what dart and ruby sass compile to
```css
:import("./foo") {
  color: red;
}
```

jquense:
See sass/sass-spec#1462

fixes #2424, fixes #2944  I think.

Not a c++ dev so please forgive me if this is bad 🙏 It seems to work, i've added a few tests and removed some existing TODO's but i guess i need to PR those to the spec repo directly. I'm not sure on the order of operations on that...

I didn't know know how to parse the unknown case, so reused the css custom variable logic since I think they should be the same, or close?

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -766,10 +766,23 @@ namespace Sass {
           if (peek_css< exactly<')'>>()  && Util::equalsLiteral("nth-", name.substr(0, 4))) {
             css_error("Invalid CSS", " after ", ": expected An+B expression, was ");
           }
-          if (SelectorListObj wrapped = parseSelectorList(true)) {
-            if (wrapped && lex_css< exactly<')'> >()) {
-              Pseudo_Selector* pseudo = SASS_MEMORY_NEW(Pseudo_Selector, p, name, element);
-              pseudo->selector(wrapped);
+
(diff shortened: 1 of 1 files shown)


nschonni:
Think you might need to enable some sass-spec tests to make sure this is working.
You can see the format used by the PR description so that libsass and sass-spec can test runs for particular PRs in https://github.com/sass/libsass/pull/2949

jquense:
Yes PR for that coming soon, took me a bit to get the repo setup, but I think I've got it now. We'll send over

PR #2992, opened 2019-09-16
Compressed style: Fix space removal after ancestor

### LINKED ISSUE #2988: Compressed style should not remove meaningful space between selectors
## input.scss

```scss
.foo .bar {
  top: 0;
}
```

## Actual results (output style: compressed)

```css
.foo.bar{top:0}
```

The combinator space between `.foo` and `.bar` is incorrectly removed.

## Expected result (output style: compressed)

```css
.foo .bar{top:0}
```

There is a space between `.foo` and `.bar` as in the source (meaning that the rule is valid in case `.bar` is a descendant of `.foo`).
## Version info

```cmd
$ sassc -v
sassc: 3.6.1-5-g507f0
libsass: 3.6.1-36-geb54
sass2scss: 1.1.1
sass: 3.5
```

glebm:
echo '.foo .bar{top:0}' | sassc/bin/sassc -t compressed

Before: `.foo.bar{top:0}`
After:  `.foo .bar{top:0}`

Fixes #2988

### DIFF
--- a/src/inspect.cpp
+++ b/src/inspect.cpp
@@ -1054,14 +1054,21 @@ namespace Sass {
   }
   void Inspect::operator()(ComplexSelector* sel)
   {
-    bool many = false;
     if (sel->hasPreLineFeed()) {
       append_optional_linefeed();
     }
+    const SelectorComponent* prev = nullptr;
     for (auto& item : sel->elements()) {
-      if (many) append_optional_space();
+      if (prev != nullptr) {
+        if (typeid(*item) == typeid(SelectorCombinator) ||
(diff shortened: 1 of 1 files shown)


glebm:
Note that this change affects all selector combinator, e.g.:

```
echo '.foo>.bar{top:0}' | sassc/bin/sassc -t compressed
```

Before: `.foo>.bar{top:0}`
After:  `.foo > .bar{top:0}`

glebm:
Merging as this is a hotfix to a severe bug.

zengabor:
Thank you so much!

futureweb:
great Gleb ... thx!!! ;-)

PR #3011, opened 2019-10-21
Change "sanctioned" to "approved"

### LINKED ISSUE #3010: Change "Sanctioned" to "Approved" in Readme.md for 3rd-Party Implementation List
# Change `"Sanctioned"` to `"Approved"` in Readme.md for 3rd-Party Implementation List

Have you noticed how the word **sanction** can have different, practically opposite, meanings in different contexts? In some cases, it can mean something like a _penalty_, while in others, it can mean something like an _endorsement_. The meaning can also be shaped by whether the use is in noun-form, like "a sanction", or verb-form, like "to sanction" or "sanctioned activity" (past participle).

This [English StackExchange post](https://english.stackexchange.com/questions/48837/how-did-sanction-come-to-have-two-opposite-meanings) discusses the differences quite well and points out how many usages today by journalists muddy the difference.

This is absolutely a subjective opinion, but to me, when I see

> Excerpt of "sanctioned" implementations:

in the [Readme.md](https://github.com/sass/libsass/blob/4da7c4bd13b8e9e5cd034f358dceda0bbba917d2/Readme.md) , I read, "Excerpt of *bad for some reason/unapproved* implementations".

The addition of the quotes surrounding "sanctioned" also gives me pause as to what the writer is signalling. Is this a quotation? **Is this unusual usage of the word**? Hmm.

But then, you can sort of piece it together and realize the writer is listing implementations that are somehow _approved_.

Again, subjective, but, regarding the [current Readme.md](https://github.com/sass/libsass/blob/4da7c4bd13b8e9e5cd034f358dceda0bbba917d2/Readme.md), I would like to see:

- [x] Change of the word "sanctioned" to "approved" to make it explicitly clear the purpose of the list, given the ambiguous meaning of that word.
- [x] Don't use quotations, regardless of the chosen word. 
- [x] Title-case this heading line altogether.

Please pardon me if this non-productive.

t-mart:
Clarify the meaning of the list in the Readme.md.

Fixes #3010 if the community/leadership accept it.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -20,7 +20,7 @@ ship with a command line interface (CLI) you can use directly. There is also
 [SassC](https://github.com/sass/sassc), which is the official lightweight
 CLI tool built by the same people as LibSass.
 
-### Excerpt of "sanctioned" implementations:
+### Excerpt of Approved Implementations:
 
 - https://github.com/sass/node-sass (Node.js)
 - https://github.com/sass/perl-libsass (Perl)
--- a/Readme.md
+++ b/Readme.md
@@ -20,7 +20,7 @@ ship with a command line interface (CLI) you can use directly. There is also
 [SassC](https://github.com/sass/sassc), which is the official lightweight
 CLI tool built by the same people as LibSass.
 
-### Excerpt of Approved Implementations:
+### Excerpt of Supported Implementations:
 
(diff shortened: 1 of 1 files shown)


nschonni:
Or maybe "suggested", because I think some may not actually be "supported" anymore 😉

PR #3016, opened 2019-10-23
Optimize big stack allocation (move to heap)

mgreter:


src/file.cpp
```diff
@@ -80,18 +80,18 @@ namespace Sass {
     bool file_exists(const std::string& path)
     {
       #ifdef _WIN32
-        wchar_t resolved[32768];
+        std::vector<wchar_t> resolved(32768);
```
glebm: Why is this so big anyway?
Should use `MAX_PATH` instead on Windows, which is only 260

https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation

glebm: Alternative is to enable long paths: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later

glebm:
Stack allocation is free but heap isn't; how is this an optimization?

src/file.cpp
mgreter: > To extend this limit to 32,767 wide characters, call the Unicode version of the function (GetFullPathNameW), and prepend "\?" to the path.

We do use GetFullPathNameW and already use long UNC paths.

src/file.cpp
```diff
@@ -437,18 +437,18 @@ namespace Sass {
...
         // windows unicode filepaths are encoded in utf16
         std::string abspath(join_paths(get_cwd(), path));
         if (!(abspath[0] == '/' && abspath[1] == '/')) {
           abspath = "//?/" + abspath;
         }
         std::wstring wpath(UTF_8::convert_to_utf16(abspath));
         std::replace(wpath.begin(), wpath.end(), '/', '\\');
-        DWORD rv = GetFullPathNameW(wpath.c_str(), 32767, resolved, NULL);
+        DWORD rv = GetFullPathNameW(wpath.c_str(), 32767, &resolved[0], NULL);
```
glebm:
```suggestion
        DWORD rv = GetFullPathNameW(wpath.c_str(), max_chars, resolved.data(), NULL);
```

src/file.cpp
glebm: Perhaps we want to initialize with zeroes explicitly instead of expecting the allocator to do so?

```suggestion
        const std::size_t max_chars = 32767;
        std::vector<wchar_t> resolved(max_chars + 1, 0);
```

src/file.cpp
```diff
@@ -80,18 +80,18 @@ namespace Sass {
...
```
glebm:
```suggestion
        DWORD rv = GetFullPathNameW(wpath.c_str(), max_chars, resolved.data(), NULL);
```

src/file.cpp
```diff
@@ -437,18 +437,18 @@ namespace Sass {
...
         if (rv > 32767) throw Exception::OperationError("Path is too long");
         if (rv == 0) throw Exception::OperationError("Path could not be resolved");
-        HANDLE hFile = CreateFileW(resolved, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
+        HANDLE hFile = CreateFileW(&resolved[0], GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
```
glebm:
```suggestion
        HANDLE hFile = CreateFileW(resolved.data(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
```

src/file.cpp
```diff
@@ -80,18 +80,18 @@ namespace Sass {
...
           abspath = "//?/" + abspath;
         }
         std::wstring wpath(UTF_8::convert_to_utf16(abspath));
         std::replace(wpath.begin(), wpath.end(), '/', '\\');
-        DWORD rv = GetFullPathNameW(wpath.c_str(), 32767, resolved, NULL);
+        DWORD rv = GetFullPathNameW(wpath.c_str(), 32767, &resolved[0], NULL);
         if (rv > 32767) throw Exception::OperationError("Path is too long");
         if (rv == 0) throw Exception::OperationError("Path could not be resolved");
-        DWORD dwAttrib = GetFileAttributesW(resolved);
+        DWORD dwAttrib = GetFileAttributesW(&resolved[0]);
```
glebm:
```suggestion
        DWORD dwAttrib = GetFileAttributesW(resolved.data());
```

src/file.cpp
```diff
@@ -437,18 +437,18 @@ namespace Sass {
       #ifdef _WIN32
         BYTE* pBuffer;
         DWORD dwBytes;
-        wchar_t resolved[32768];
+        std::vector<wchar_t> resolved(32768);
```
glebm:
```suggestion
        const std::size_t max_chars = 32767;
        std::vector<wchar_t> resolved(max_chars + 1, 0);
```

src/file.cpp
mgreter: This shouldn't be necessary, since GetFullPathNameW will fill the buffer anyway!

mgreter:
@glebm care to create a PR with your suggested changes!?

saper:
> > Stack allocation is free but heap isn't; how is this an optimization?
> 
> Out of stack issues? https://docs.microsoft.com/en-us/visualstudio/code-quality/c6262?view=vs-2019

There is nothing wrong per se in having 32KB on stack. We could add an option to have `/analyze:stacksize40000` or something.

Did this cause real-life crashes of libsass due to running out of stack? Even Alpine Linux lets us having more stack ... (but this code does not apply to Linux).

src/file.cpp
glebm: This suggestion is to avoid repeating `32767` and to use `.data()` instead of `&[0]`

mgreter:
The memory payload is actually 64k, since the allocated array is of `wchar`.
In the end this is just a precaution to satisfy MSVC static code analysis.
But I think it does make sense to allocate those 64k on the heap!?

PR #3038, opened 2019-11-12
Add custom memory allocator (opt-in only so far)

mgreter:
For CI purpose first try with opt-in enabled.

Performance improvements 20% for complex cases.
Potential for up to 30% improvement later ...

Also updates the documentation a bit.


src/file.cpp
```diff
@@ -68,7 +68,7 @@ namespace Sass {
         wchar_t wd[wd_len];
         wchar_t* pwd = _wgetcwd(wd, wd_len);
         if (pwd == NULL) throw Exception::OperationError("cwd gone missing");
-        sass::string cwd = wstring_to_string(pwd);
+        sass::string cwd = wstring_to_string(pwd).c_str();
```
glebm: Why call `c_str()` here? Seems entirely redundant

glebm: Also why do we have a `sass::string`?

mgreter: Because it's a specialization of std::string with the custom allocator.
Those can't be directly converted to std::string AFAICT.

glebm: Strange, `basic_string` does define this https://en.cppreference.com/w/cpp/string/basic_string/operator%3D

glebm: What error are you getting? This should just work.

glebm: My guess is it's trying to call `Allocator(std::allocator const&)`, which we don't define. Defining this copy-constructor could then solve the incompatibility.

mgreter: Feel free to poke around once I get this to pass CI.

Makefile.conf
```diff
@@ -14,6 +14,8 @@ SOURCES = \
 	ast_sel_super.cpp \
 	ast_sel_weave.cpp \
 	ast_selectors.cpp \
+	allocator.cpp \
```
glebm: Consider moving these 2 files to `src/memory`

Issue #3084, opened 2020-04-30
wrong :before position


ahorek:
Hi, was this change intentional or is it a bug? Probably it's not a bug, but I just want to be sure.

## input.scss

```scss
.red-icon {
  color: #e50026;
}

a.nested:before {
  &.error {
    @extend .red-icon;
  }
}
```

## Actual results

[libsass 3.6.3][1]
```css
.red-icon, a.nested:before.error {
  color: #e50026; }
```

dart sass also have this behavior

in this case ```<div class="red-icon">text</div>``` isn't red.

## Expected result

[libsass 3.6.1][1]
```css
.red-icon, a.nested.error:before {
  color: #e50026; }
```

[1]: http://libsass.ocbnet.ch/srcmap/#LnJlZC1pY29uIHsKICBjb2xvcjogI2U1MDAyNjsKfQoKYS5uZXN0ZWQ6YmVmb3JlIHsKICAmLmVycm9yIHsKICAgIEBleHRlbmQgLnJlZC1pY29uOwogIH0KfQ==

mgreter:
Can you please open an issue with dart-sass since you established it has the same issue?

mgreter:
Note that the extend is not even needed, minimal reproduction also with:
```
a.nested:before {
    &.red-icon {
        asd: qwe;
    }
}
```

PR #3086, opened 2020-05-01
Ensure correct output order of compound selectors

mgreter:
Fixes https://github.com/sass/libsass/issues/3084, Needs https://github.com/sass/sass-spec/pull/1529

~I believe this will fail quite a few spec tests, since it seems we have
some of these as expected outputs!? Lets check what CI returns.~

Opened PR to address this in sass-spec.

### DIFF
--- a/src/ast_selectors.cpp
+++ b/src/ast_selectors.cpp
@@ -940,6 +940,16 @@ namespace Sass {
 
   }
 
+  bool cmpSimpleSelectors(SimpleSelector* a, SimpleSelector* b)
+  {
+    return (a->getSortOrder() < b->getSortOrder());
+  }
+
+  void CompoundSelector::sortChildren()
+  {
+    std::sort(begin(), end(), cmpSimpleSelectors);
+  }
+
   /* better return sass::vector? only - is empty container anyway? */
   SelectorList* ComplexSelector::resolve_parent_refs(SelectorStack pstack, Backtraces& traces, bool implicit_parent)
   {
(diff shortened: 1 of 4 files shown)


mgreter:
I am going to merge this, as I have some time at hand right now to make a release. Even though knowing it will fail CI, but I cant merge the changes in sass-spec- But I'm certain this is an improvement in the right direction; why not emitting correct css selector when we can. Maybe dart-sass comes up with a different approach than to ensure the order on output.

Note to myself: It might be worthwhile to check if we can integrate id and type selectors directly into the CompoundSelector, since css only allows to have one of each per compound selector. I already did the same for `hasRealParent`. Could safe some allocations and might also simplify selector unifying.

ahorek:
thanks for the release @mgreter !

mgreter:
No problem @ahorek, still quite a few steps to do for a complete release :)

nex3:
This breaks compatibility with Dart Sass and with the specified behavior of Sass. Please roll back for now.

If you want to propose a change to the language itself, you need to follow [the language change process](https://github.com/sass/sass/blob/master/CONTRIBUTING.md).

xzyfer:
Given the discussion in https://github.com/sass/dart-sass/issues/996 we'll need to revert this PR. As @nex3 this is change to the Sass language and makes us non-compliant. Embedders like Node Sass are now blocked on updating LibSass at the risk of introduce incompatible Sass code into the ecosystem.

PR #3101, opened 2020-05-21
Add warning for invalid selectors (parsed or generated)

mgreter:
Needs https://github.com/sass/sass-spec/pull/1535

### DIFF
--- a/src/ast_selectors.cpp
+++ b/src/ast_selectors.cpp
@@ -433,6 +433,16 @@ namespace Sass {
     return false;
   }
 
+  bool ComplexSelector::isInvalidCss() const
+  {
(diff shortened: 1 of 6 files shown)


ahorek:
those warnings might be too verbose.

for instance
```a:hover.button { }```
actually doesn't break anything

I would propose to warn about pseudo-elements, but keep unordered pseudo-classes without warnings. Thoughts?

ahorek:
I have to agree with @nex3 feedback https://github.com/sass/sass-spec/pull/1535 . Thank you very much for your effort, but this change shouldn't be merged.

mgreter:
Unfortunately this approach was denied by upstream sass: https://github.com/sass/sass-spec/pull/1535#issuecomment-635668097

PR #3129, opened 2020-09-23
Prevent compiler warning about unnecessary copy

acocuzzo:
This compiler warning comes up a lot, and is due to the unnecessary copy introduced by accidentally using const value instead of const ref here.


xzyfer (approved):
Thanks

PR #3131, opened 2020-10-19
fix error of working with the pointer.

ihsinme:
I think this is a simple mistake. at this point, the pointer becomes a number. which can be dangerous under certain circumstances.
but here I would pay attention to the broken functionality.


mgreter:
Thanks for your contribution.

PR #2918, opened 2019-06-22
[WIP] Early preview of dart-sass parser backport

mgreter:
Needs https://github.com/sass/sass-spec/pull/1425

Please don't review, only leaving this here since I don't know when I'll have time to finish this up.
My vacations are over soon, so I don't know when I'll get around to complete this.
It's already semi usable, passing 80% of current spec tests.
Tried to preserve current libsass behavior, but
probably needs changes in eval stage too!
All the debug glory is still in the code!

Edit: I started to bring back some stuff to master in order to make this change a little bit smaller.
I also tagged all [open issues that should be fixed by this backport/refactoring][1] (ATM 76/120).

[1]: https://github.com/sass/libsass/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3A%22Dart+Backport+Done%22+

PR #3135, opened 2020-11-06
[DNM] LibSass 4.0 Alpha Big Bang

mgreter:
Continued from https://github.com/sass/libsass/pull/2918

Please note that this PR is a little bit bigger than your regular PR :)

During the last year development of LibSass seems to have stagnated, which is certainly partially true!
Although corona has hit us all, I still did enjoy the beautiful summer here in my hometown and took
some time off from GitHub and opensource. None the less I already had invested quite some time into
a dart-sass parser back-port, after the back-port for the extend code already landed in master (not without
any headaches). During the year I've back-ported, refactored or rewritten nearly 90% of the code-base.

I always had a version that was working, but the code behind it was never 100% clean. It still isn't in a lot
of parts, but in many others the code is now reviewed and properly done. The new version will also require
at least a very decent C++11 compiler, as I took great care to use move semantics wherever possible!

Don't bother to try to review this PR, you can just review the whole code base instead ;)
There are still edges not fully reworked or that still have some debugging traits attached!
Corresponding sass-spec branch at https://github.com/mgreter/sass-spec/tree/refactor/libsass-4-alpha

Open points for LibSass 4.0 ALPHA release:
- [x] Define, implement and finalize C-API
- [ ] Fully finalize the C-API (85%)
- [ ] Update documentation pages (90%)
- [ ] Cleanup exception handling (40%)
- [x] Replace addFinalStackTrace (100%)
- [x] Change default precision to 10.
- [ ] Finalize terminal/logger (90%)
- [ ] Check source mappings (80%)
- [ ] Check output styles (35%)

Additional points to be done:
- [ ] Get approval for sass-spec changes (or use fork repo)
- [ ] Get approval to merge into sass/libsass master
- [x] Render error css to output file

Optional stuff:
- [x] Add training target for MSVC
- [ ] Move warn return to C-API invocation.
- [ ] Extend C-API for SassNumber units.
- [ ] Extend C-API for SassFunctions.
- [ ] Custom importer for `@use` etc.
- [ ] Compile to wasm32-wasi target.
- [ ] Support external callbacks for log/warn.

I will update those list if I find any other points.

For the release notes I might also want to mention the main improvements:
- Native css/sass imports are now supported
- Added HWB color modes (only for already available functions)
  - Functions `whiteness`/`blackness` are only exposed via module system
- CRTP changed to more clear visitor pattern (tested no performance impact)
- Detangled expression, statement, value and selector base classes
- Fixed quoted/literal strings, binary operations and interpolations
- Initial module use and forward support

BTW I will keep squashing and force pushing this branch!


nschonni:
Since most of the code is changing, does it make sense to setup a formatter like in the (broken) https://github.com/sass/libsass/pull/2324

mgreter:
@nschonni Still want to wait with that, I used default formatter of MSVC up to now.
IMO it's something we can easily do at any point and nothing gained or lost by not doing it.
There are still bigger fish to fry, and so long I'm mostly the only one working with the code.

asottile:
(not going to subscribe so if you need to reply, please @-mention me -- thanks!)

I took a look at this [on stream today](https://twitch.tv/anthonywritescode) and ran into a few things and I think I'm stuck (maybe an api is missing?).  here's the notes I have for this:

- `sass_value_get_tag` should take `const struct SassValue*` instead
- I think I need `sass_make_import_error` for custom importers, I couldn't figure out how to port the existing code
- how do I replace `sass_make_data_context` -- I assume with `SassCompiler*`
  but I don't see any apis for it.  same for `sass_make_file_context`

I also created two PRs against your branch to fix a few small problems I ran into:
- https://github.com/mgreter/libsass/pull/6
- https://github.com/mgreter/libsass/pull/7

excited for the changes here!  some of the map / list push / set stuff is really really nice!

asottile:
neat! I got a lot further today, here's my next round of feedback!

I couldn't get sourcemaps working at all, I'm guessing they're not completed yet or I'm holding it wrong -- probably the latter let's be realistic 😄 

I also couldn't get custom function working 🤔 

I've got two new PRs for today: https://github.com/mgreter/libsass/pull/8 https://github.com/mgreter/libsass/pull/9

and I have a bunch of oddities I can't explain :S

- compiling `a{}` used to lead to a result of an empty string, but now produces no output at all (null?) -- I didn't dig into this much, I'm not quite sure where this comes from, my guess is there's a truthy check where there should be a null check somewhere 🤷 

compressed mode seems to be overly aggressive, breaking some selectors:

```console
$ pysassc testpkg/testpkg/static/scss/a.scss -t compressed
pa{color:red}pb{color:blue}

$ cat testpkg/testpkg/static/scss/a.scss
p {
  a {
    color: red;
  }
  b {
    color: blue;
  }
}
```

the default output mode seems to have changed indentation slightly:

```console
$ pysassc test/a.scss  # NEW
body {
  background-color: green; }

body a {
  color: blue; }

$ /tmp/venv/bin/pysassc test/a.scss  # OLD
body {
  background-color: green; }
  body a {
    color: blue; }

```

and lastly, it seems the quotes have changed from single to double:

```console
$ pysassc test/d.scss  # NEW
@charset "UTF-8";
body {
  background-color: green; }

body a {
  font: "나눔고딕", sans-serif; }


$ /tmp/venv/bin/pysassc test/d.scss  # OLD
@charset "UTF-8";
body {
  background-color: green; }
  body a {
    font: '나눔고딕', sans-serif; }
```

mgreter:
SourceMaps and custom functions/importers are indeed last on my list!
Although SourceMaps should be producible, but I haven't checked the results.

Was busy the last to weeks implementing use and forward modules.
Had experimentally done it before so I knew it needs a major overhaul of scope handlings.
This should now be done and I only seem to have lost about 5% of performance currently.
It seems to be pretty complete already, but no module extend barriers have been added yet.
Anyway should see a sea of passing spec-tests :)

mgreter:
For the other points you mentioned:

```
pa{color:red}pb{color:blue}
```

Is this really invalid, seems to be perfectly readable to me as a human :)
Nah, need to check if this really breaks, then yes, this needs to be fixed.

For the indentation I try to match dart-sass, so need to check against that.
I consider these white-space changes not very important or do you disagree?
The same goes for the quotes, as long as the are valid I think it should be ok.

kerryj89:
I came from here https://github.com/sass/sass/issues/2737 and saw the huge leaps in performance you've made!  Really impressive stuff.

My project has one entrypoint file for sass and uses the latest gulp-sass (dart-sass js) to build it. This takes about about 10s to build for every change which is about the same time as node-sass surprisingly (I would have thought node-sass using libsass would be much faster but I think the dart-sass team has made some great gains in performance in the last few years to make the JS version comparable to the libsass binary).

Anyway, what surprised me was that dart-sass cli only took 1.734s to build my project.  Dart-sass team are working on "Embedded Dart Sass" to leverage this efficiency because this level can't be reached by relying on the JS version.

With all that said, if your version is out soon I may change back to node-sass just for the sake of these massive speed gains.  Thanks for maintaining and backporting even though this project is listed as deprecated, I always like having options.  I would take speed over feature-set until my needs for the project are no longer met, and having survived without most of the new features I think that can last a few more years.

mgreter:
Sorry for the long silence, but as you know, life and stuff 😃 The C-API part of the todo list is closely to be finished, one of the biggest open points for a LibSass 4.0 alpha release. I guess most other points can be prolonged into the alpha phase, as they are often optional and highly OS dependent (e.g. terminal Color or Unicode recognition). Implementers can often simply disable a feature via options if it turns out to be unreliable or bugged. On the other hand there are so many changes that we would like to get it right in the first place as much as possible.

johnfairh:
Great to see all the progress on this.  I've been playing at wrapping a Swift API around this branch; the core of the thing seems really good, here are a few minor problems I found:
* Output path is not inferred from input path -- because [it's initialized non-empty](https://github.com/mgreter/libsass/blob/5bc2b86d6b6d38e84447e23753d38903f062b7d8/src/compiler.cpp#L55)
* `error_get_string()` always comes back `NULL` -- I think because the `what` variable is nobbled by [this loop](https://github.com/mgreter/libsass/blob/5bc2b86d6b6d38e84447e23753d38903f062b7d8/src/capi_compiler.cpp#L28) before it's copied
* `set_logger_style()` doesn't always work -- the requested style is [ignored unless AUTO](https://github.com/mgreter/libsass/blob/5bc2b86d6b6d38e84447e23753d38903f062b7d8/src/logger.cpp#L22)
* `Units::unit(u)` does not parse the '(x*y)^-1' style that can be produced by `Units::unit()`
    (hmm I can't remember why I cared about this -- so FYI I suppose!)
* `@use` doesn't seem to consult custom importers (no problems with `@import`)

Happy to PR at least the easy ones...

Some less specific things:
* The `CWD` caching may be scoped too widely.  I was surprised to find that when I changed my process's current directory between separate Sass compile jobs that LibSass retained the previous directory.  Easily addressed with `sass_chdir()` but the [comment on the prototype](https://github.com/mgreter/libsass/blob/5bc2b86d6b6d38e84447e23753d38903f062b7d8/include/sass/base.h#L70) is a bit discouraging -- what design point do you have in mind here?
* The [signal handling](https://github.com/mgreter/libsass/blob/5bc2b86d6b6d38e84447e23753d38903f062b7d8/include/sass/base.h#L70) is more opinionated than I expect in a library -- perhaps this is just for your bringup though?
* It's probably worth exposing a bunch of the unit APIs under `sass_number_` as per Dart -- it is tough to work with the pretty-printed string.  I realise this is all new WIP code for LibSass.
* Did you consider unifying the `Sass::Function` and `SassFunction` types at all?  Dart Sass allows the return value of a custom function to be [a new custom function](https://pub.dev/documentation/sass/latest/sass/SassFunction-class.html) which I don't _think_ is possible here.

Again I'm really glad to see your efforts here.

mgreter:
Thx @johnfairh for the feedback, very useful and encouraging as it seems you already got quite far with the current state.

- Output path is not inferred from input path -> seems to be a bug then
- `error_get_string()` always comes back `NULL`  -> will check that, thx
- `set_logger_style()` doesn't always work -> else case seems missing, thx
- `Units::unit(u)` does not parse the '(x*y)^-1' -> need to check how far I want to extend this for a initial 4.0 relesae
- `@use` doesn't seem to consult custom importers -> currently deliberately left out. Not sure if want to up the API a bit to support these new cases different. IMO this could wait until after an initial 4.0 release until people adapt this new feature more often.
- The `CWD` caching may be scoped too widely. -> IMHO changing the working directory via sysfs calls is dangerous, as the CWD is a process wide entity and is therefore shared between all threads and linked libraries of your app. So if this would be the only way for an implementor to tell LibSass where to look for the initial file load, we could introduce issues if an app uses different libraries or code that depends on `CWD`. Therefore I guess it's safest to just get it once and then use a local variable for the CWD (which makes it necessary to have the sass_chdir function). Not sure why I added the warning on the API though.
- signal handling -> not sure what you mean exactly as the link seems incorrect!?
- exposing a bunch of the unit APIs -> could be a future improvement, IMO for an initial release this has very low priority. I will add it to the Optional ToDo list in the initial post in this thread.
- unifying the Sass::Function and SassFunction types -> I did consider it but is also something that can be added after an initial release. There are certainly more of these nice to have details, but I think it's better now to focus on bringing what is currently there to a releasable state. But will also add it to the Optional ToDo list.

Thanks again and please feel free to poke around more in the code ;)
Maybe I can also squeeze in some more hours over Easter!

[1]

mgreter:
Hopefully fixed `error_get_string()`and `set_logger_style()`. Also checked "Output path is not inferred from input path", but after looking at it again, I'm not sure we actually can infer the output path, although we might have done so in the past, so this might be valid after all, but needs a bit more clarification. E.g. if CLI only specifies input path we assume the output is just printed to the console and this kinda implies that we can't e.g. reliably create sourcemap references (I guess this is were I was aiming at with it).

johnfairh:
Thanks; fixes look good.  Your general approach of prioritising releasing something sounds right to me!

I understand your point about CWD as global shared state.  Think I agree current behaviour is fine with docs.

[This](https://github.com/mgreter/libsass/blob/a34cd4ba04cc250c966254e7eadc1986e37e9d33/src/compiler.cpp#L723-L728) is the signal code I meant to link: I don't generally expect a library to grab all these and decide what the right behaviour is for the host program.

Output path inference.  Old LibSass [infers the output path](https://github.com/sass/libsass/blob/288610acfaa797b44c1a58a18b2cbd2e03937498/src/context.cpp#L36-L43) if the user supplies an input file; as you hint the only usage is to fill in the optional 'file' SourceMap field -- so really maybe any behaviour here (1 - infer a path; 2 - assume stdout; 3 - don't fill in the field) is acceptable as long as it's documented; the SourceMap can always be fixed up by the client later.  Marginal preference for keeping the old behaviour I suppose.

johnfairh:
Looking good.  Source maps definitely getting better; fyi I saw some non-deterministic behaviour with a spurious 'AAAA' segment sometimes appearing -- seems to have gone away by adding an initial value of `false` for `Emitter::force_next_mapping` [here](https://github.com/mgreter/libsass/blob/53396bd0f49250400f480b348a406c9e69c19918/src/emitter.cpp#L20) (in the debugger I spotted its value was 152 which seemed fishy)

mgreter:
Thx @johnfairh, the initialization of `force_next_mapping` was certainly missing, good catch. Interestingly `AAAA` mappings should never be emitted anymore, since I added a check for that specific case. As you may have seen I struggle a little with how to handle source-maps, since UAs are a bit limited and strange how they interpret source-maps. I hope with the two new options (`useOptionalOpeners` and `useOptionalClosers`) implementors can choose what they like best and LibSass doesn't need to bother to make this decision for them all.

mgreter:
Lately I've been working more on the bare metal of the C-API again (and updated the docs substantially). In that sprint I've moved more logic from sassc into LibSass, mainly the parsing of the command-line arguments. This was done with two main goals in mind. First I never liked and understood the getopt implementation we had included in sassc. E.g. we've introduced bugs in the past because nobody could understand how to add new short-options (the API is very cryptic in that regard). Re-implementing this in C was out of question, since it is so much easier in C++ (IMHO). Current implementation also lacks the ability to recognize shortened options which are already unambiguous (with GNU getopts you can shorten long-names, as long as they are unambiguous).

So I've added a very similar implementation in C++, and since sassc needs to use it, it is also exposes on the C-API. This may feel a bit counter intuitive to only move it from sassc implementer to the actual library. But the interface does make sense to have, since in its root, it simply passes an array of strings to LibSass to configure the compiler. This might be actually interesting for other implementations too.

Finally the [sassc.c][1] code that is left is so little that I'm really tempted to kill the sassc repo completely and just move sassc into LibSass directly. There a quite a lot of pro arguments and a few contra arguments. The main reason to keep it as it is, is probably the clear separation (which is also a major pain in regard to finding the right sassc for an old libsass version). Or does anybody see any other reasons why not to move sassc into LibSass?

[1]: https://github.com/mgreter/sassc/blob/refactor/libsass-4-alpha/sassc.c

mgreter:
Minor update as I'm currently trying to rebase sass-spec to latest master. There have been some major additions in dart-sass in regard to "slash" lists and "math.div" which are a bit more complicated to support, mainly because of the deprecation warnings, as LibSass doesn't have a way (yet) to convert Expressions to strings. So this might add another month or two of development to support correctly.

mgreter:
Continues in https://github.com/mgreter/libsass-ng/tree/develop
And https://github.com/mgreter/libsass-ng/actions/runs/9558047752
Still a lot of work to do: https://github.com/sponsors/mgreter 😄

Issue #2880, opened 2019-04-18
url() should be able to contain an exclamation mark


nex3:
The [CSS spec](https://drafts.csswg.org/css-syntax-3/#url-token-diagram) says that a `url()` can contain `!`, so Sass should allow that too. Note that we don't allow *everything* that CSS does, because we want to be able to parse `url($foo)` as a URL taking a variable, but `!` is unlikely to be used in a SassScript expression.

See https://github.com/sass/dart-sass/issues/646

glebm:
libsass should already allow it (https://github.com/sass/libsass/blob/8d220b74770ad16aaf7819b5e18d85746aa63d7d/src/constants.cpp#L161), need to check if the original example works in libsass.

Note that the test case added in https://github.com/sass/sass-spec/commit/eb787d9f071bb872738375e80111b1dbec0f96c8 is invalid because `!` by itself is not a valid URL (`!` is only allowed in certain parts of the URL, not everywhere).

glebm:
Confirmed. The spec should be fixed as well with valid URLs (we shouldn't make any guarantees about invalid ones).

nex3:
According to [CSS Syntax Level 3](https://www.w3.org/TR/css-syntax-3/#url-token-diagram), a `<url-token>` without quotes can contain any unescaped character other than `"`, `'`', `(`, `)`, `\`, whitespace, or non-printable characters. Whether it constitutes a valid WhatWG URL or not is not syntactically relevant.

mgreter:
Should be fixed with LibSass 3.6.5

PR #3142, opened 2020-12-17
fix: url() contains exclamation marks cause parsing error (#2880)

Mutefish0:
it should work:
```scss
.a {
  background-image: url(https://example.com/abc!!def.png);
}
```

Refs [sass/libsass#2880](https://github.com/sass/libsass/issues/2880)


mgreter:
Thx!

PR #3150, opened 2021-05-01
Replace Travis CI with GitHub Actions

mgreter:


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,218 @@
...
+      run: echo "/d/a/libsass/libsass/lib" >> $GITHUB_PATH
+    - name: Compile libsass ${{ matrix.config.build }} build for ${{ matrix.config.platform }}
+      run: make ${{ matrix.config.build }} BUILD=${{ matrix.config.build }}
+    - name: Copy library over to pass call test
+      if: matrix.config.build == 'shared'
+      run: copy /a/libsass/libsass/lib/libsass.dll sassc/bin/
+    - name: Compile sassc ${{ matrix.config.build }} build for ${{ matrix.config.platform }}
+      run: make sassc BUILD=${{ matrix.config.build }}
+    - name: Execute spec test runner
+      run: ruby sass-spec/sass-spec.rb --probe-todo --impl libsass -c sassc/bin/sassc.exe -s sass-spec/spec
```
stof, line 214: why using the old ruby runner rather than the new typescript one ?

mgreter: Was enough work to convert the existing code, so this didn't change.
Of course, if somebody wants to convert this, PRs are always welcome!

PR #3152, opened 2021-05-10
Use pseudo-random rand() instead of cryptographically secure random numbers

### LINKED ISSUE #3151: random_device: rdrand failed - Crashes host application
When using an application that uses libsass, the application crashes (segfaults) with `random_device: rdrand failed`.

# Reproduction

1. Start an application that uses libsass, e.g. simply run `hugo`.

## Actual results

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  random_device: rdrand failed
Cancelled (Segfault)
```

## Expected result

Hugo works

## Version

This happens on the latest version of libsass1 in Ubuntu, currently libsass version 3.6.3. However, the problem also exists on current libsass master, because the relevant code has not changed.

## Cause

1. Some AMD CPUs seems to return a non-random number, but still claim success. See e.g. [reports on Twitter](https://twitter.com/FiloSottile/status/1125840275346198529).
2. `std:random_device` throws an exception.
3. libsass is unable to cope, throws the exception up into the caller, which cannot possibly handle this error and fails.

The underlying root cause is that libsass is using cryptographically secure random numbers. Why? I see no need in a CSS parser for that. Pseudo-random should be sufficient.

## Fix

Simply use pseudo-random numbers, from `rand()`.

I've tested that, and that fixes the bug. PR will follow.

benbucksch:
Some AMD CPUs seems to return a non-random number, but still claim success. This makes `random_device` throw an exception, which is passed to the host application, which then crashes.

To work around this, simply use pseudo-random numbers.

Fixes #3151 and probably sass/node-sass#3047

This is a simplistic fix, but is tested to work and fix the problem on a CPU where it was broken before. I can surely imagine more elaborate fixes, e.g. to use `random_device` first, catch the exception, and use `rand()` only as fallback.

However, I do wonder why libsass is using cryptographically secure random numbers in the first place, and what for. I don't see why CSS would need that. I would think that pseudo-random is sufficient. So, maybe this is even the right fix, as-is.

I am contributing the fix that I found. If you would like to improve this, please go ahead. My involvement stops here. My objectively was merely to continue to use the host application, which suddenly failed on the new computer. I hope to help others who run into the same problem.

### DIFF
--- a/src/fn_numbers.cpp
+++ b/src/fn_numbers.cpp
@@ -41,8 +41,7 @@ namespace Sass {
     #else
       uint64_t GetSeed()
       {
-        std::random_device rd;
-        return rd();
+        return rand();
       }
     #endif
 


mgreter:
As mentioned in the issue ticket, this approach will generate the same output over and over again for the following sass:
```
test { nr: random(); }
```

On my PC it always produces:
```css
test { nr: 0.6805727266; }
```

See https://github.com/sass/libsass/pull/3153 for a more robust implementation.
BTW. the failing CI test are an issue with `sass-spec` and its ruby runner which two new tests have triggered ...

benbucksch:
> See #3153 for a more robust implementation.

That makes sense. Thank you!

PR #3153, opened 2021-05-10
Fix seed generator if std::random_device fails (#3151)

mgreter:
Fall back to regular "old-school" seed initialization by using pid, clock and time if std::random_device fails.


saper:
Shall we backport this to 3.5.something to include in node-sass?

PR #3160, opened 2021-07-17
Remove references to libsass.org

xzyfer:
This domain has expired.

### DIFF
--- a/res/resource.rc
+++ b/res/resource.rc
@@ -22,7 +22,7 @@ BEGIN
       VALUE "FileDescription", "A C/C++ implementation of a Sass compiler"
       VALUE "FileVersion", "1.0.0.0"
       VALUE "InternalName", "libsass"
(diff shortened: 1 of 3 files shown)


res/resource.rc
```diff
@@ -22,7 +22,7 @@ BEGIN
       VALUE "FileDescription", "A C/C++ implementation of a Sass compiler"
       VALUE "FileVersion", "1.0.0.0"
       VALUE "InternalName", "libsass"
-      VALUE "LegalCopyright", "\251 2017 libsass.org"
+      VALUE "LegalCopyright", "\251 2017 sass-lang.org"
```
nschonni, line 25: Not sure about the "LegalCopyright" being held by either domain, but not a blocker to the cleanup

nschonni (approved):

res/resource.rc
xzyfer: Yeah I dunno what these are about, I just did a replace.

PR #3163, opened 2021-08-31
power ^ added

pratik-a:


xzyfer:
Is there documentation for this operator in the Sass docs?

stof:
Sass has the feature available as `math.pow()` in the `sass:math` module. But there is no such operator.

xzyfer:
We don't support the sass module system. We are an implementation of the sass language specification. It's not our place to add new language features. If you want a new feature you should suggest it on the sass/sass repo.

